PR #91, opened 2019-02-11
Complete Model Abstraction

jmaack24:
These changes allow PowerSimulations to create power network models with StructJuMP.StructuredModel as well as the standard JuMP.Model.  PowerSimulations should now be able to create a model that is a subtype of JuMP.AbstractModel provided the user has defined the appropriate functions for that model.

These changes were made and tested with Julia 1.0.3 and JuMP v0.19-beta2.


jmaack24:
I did have to make some additions to StructJuMP to get this working.  So if anybody wants to run these changes with StructJuMP, I can provide the code.

jd-lara:
We should make sure this works with Julia 1.1 I dropped support for anything earlier in the other PR

jmaack24:
I've been able to build both JuMP and StructJuMP models using Juila 1.1.  The tests fail in the same places as they do on the master branch.

jd-lara:
@jmaack24 I made changes to support the abstraction in the refactor branch. We should avoid having containers with AbstractTypes as mentioned here. https://docs.julialang.org/en/v1/manual/performance-tips/#Avoid-containers-with-abstract-type-parameters-1

jmaack24:
I read through that conversation and their suggestions are what I was advocating yesterday and how I implemented the changes here.  Basically, you have to determine the variable type at run time rather than at "compile" time.

jd-lara:
I added these two functions to do that in the `add_variable` stage. Right now the StructJuMP one is commented out since StructJuMP depends on an unregistered package. But it should work well. 

```
function _container_spec(m::JuMP.Model, ax1, ax2)
    return JuMP.Containers.DenseAxisArray{JuMP.VariableRef}(undef, ax1, ax2)
end

#=
function _container_spec(m::StructJuMP.StructuredModel, ax1, ax2)
    return JuMP.Containers.DenseAxisArray{StructJuMP.StructuredVariableRef}(undef, ax1, ax2)
end
=#
```

jmaack24:
You can do away with the explicit StructJuMP dependency here by using the function JuMP.variable_type on the model. Something like

function _container_spec(m::M, ax1, ax2) where M <: JuMP.AbstractModel
    return JuMP.Containers.DenseAxisArray{JuMP.variable_type(m)}(undef, ax1, ax2)
end

Then we wouldn't need to implement a new version of this function for any other models that come along since this function (according to the doc string) is expected to be implemented by every JuMP.AbstractModel.

jd-lara:
True, this is a nice solution. I wasn't aware of the `variable_type` function.

jd-lara:
cc. @sourabhdalvi to answer your question.

PR #108, opened 2019-03-18
Refactor device models

kdheepak:


kdheepak:
I agree with what you are saying. I was hoping separating it with comments would help. There is value in having all the functions associated with one type together, since it is easy to understand how dispatch will occur. See [array.jl in base](https://github.com/JuliaLang/julia/blob/4dc15938bbb1b5f9fda9def3a85e80e3357a8193/base/array.jl), which is over 2000 lines but still easy to find all functions associated with the `Array` type. If a user wants to write a few specializations for a abstract type that we have defined they'll have to hunt through many folders. I guess when we generate documentation we could help with this, but looking through the code is a little tedious at the moment. If you don't have any other suggestions for making this better, I will close this for now and we can revisit later.

jd-lara:
I am mostly ok separating the files for the devices but not in the common folder.

jd-lara:
@kdheepak are we merging these? some of the changes I am making for simulations will break if we try to merge later.

kdheepak:
If the latest changes look good to you we can merge. I made the changes for everything else in the folder except common.

PR #120, opened 2019-04-02
Pm translator fix

claytonpbarrows:


jd-lara:
@claytonpbarrows  The objective function of the test for DCPlossless needs to be corrected.

PR #132, opened 2019-04-17
Jd/psy ts update

jd-lara:
Keeps progress of integrating PowerSystems.jl#dev into PowerSimulations.


jd-lara:
The code base in this PR seems to have broken the Package Manager. Substituted by #134

PR #145, opened 2019-05-14
V0.1.0

claytonpbarrows:
PR to clean dependencies in Examples repo


jd-lara:
You need to manually re-add the targets in the Project.toml file
see https://github.com/JuliaLang/Pkg.jl/issues/1184

Issue #105, opened 2019-03-16
Enable warm starting


jd-lara:

jd-lara:
This requires creating a code to store/access them when variables are created.

PR #161, opened 2019-05-29
Pm interface refactor

### LINKED ISSUE #148: remove dict creation in pass_to_pm()
See 
https://github.com/lanl-ansi/PowerModels.jl/pull/510
and 
https://github.com/lanl-ansi/PowerModels.jl/pull/505

In the near future, it will be possible to make a function that makes sys accessible like a dict without creating it removing the need to duplicate the data. Bypass `core_ref!(pm)`

jd-lara:
This PR updates the interface of PowerSimulations with PowerModels to address the requirements of some users to have access to the flow variables.  

In Summary. 

- Creates a PowerSimulations reduced version of `core_ref!` 
- Substitutes `build_generic_model` with a model instantiation 
- Changes the creation of the line flow variables in order to reference them in the PowerSimulations model and the PowerModels model. 
- implements #105 partially for some variables
- Updates the testing  

closes #148

### DIFF
--- a/src/devices/device_models/AC_branches.jl
+++ b/src/devices/device_models/AC_branches.jl
@@ -38,76 +38,226 @@ function flow_variables(ps_m::CanonicalModel,
 end
 
 
-function flow_variables(ps_m::CanonicalModel,
-                        system_formulation::Type{S},
-                        devices::PSY.FlattenedVectorsIterator{B},
-                        time_steps::UnitRange{Int64}) where {B <: PSY.ACBranch,
-                                                             S <: PM.AbstractActivePowerFormulation}
 
-    pm_object = ps_m.pm_model
-    var_name_from = Symbol("Fbr_fr_$(B)")
-    var_name_to = Symbol("Fbr_to_$(B)") 
-    ps_m.variables[var_name_to] = _container_spec(ps_m.JuMPmodel, 
-                                                  (d.name for d in devices), 
-                                                  time_steps)    
-
-    ps_m.variables[var_name_from] = _container_spec(ps_m.JuMPmodel, 
-                                                    (d.name for d in devices), 
-                                                    time_steps)    
-
-    pm_index = PM.ref(pm_object, 1, :arcs)
-
-    for t in time_steps 
-        pm_array = _container_spec(ps_m.JuMPmodel, pm_index)
-        for (ix,d) in enumerate(devices)
-            ix, d.connectionpoints
-            bus_from = d.connectionpoints.from.number
-            bus_to = d.connectionpoints.to.number
-            pm_array[(ix, bus_from, bus_to)] = ps_m.variables[var_name_from][d.name,t] = JuMP.@variable(ps_m.JuMPmodel,
-                                                           base_name="$(var_name_from)($(bus_from),$(bus_to))_{$(d.name),$(t)}",
-                                                           start = 0.0)
-                                                           
-            pm_array[(ix, bus_to, bus_from)] = ps_m.variables[var_name_to][d.name,t] = JuMP.@variable(ps_m.JuMPmodel,
-                                                            base_name="$(var_name_to)($(bus_to),$(bus_from))_{$(d.name),$(t)}",
-                                                            start = 0.0)
-        end
-            PM.var(pm_object, t, 1)[:p] = pm_array
-    end                                                        
-
-    return
-
-end
 
 function flow_variables(ps_m::CanonicalModel,
                         system_formulation::Type{S},
                         devices::PSY.FlattenedVectorsIterator{B},
                         time_steps::UnitRange{Int64}) where {B <: PSY.ACBranch,
                                                              S <: PM.AbstractPowerFormulation}
+    #Get PowerModels dicts
+    pm_object = ps_m.pm_model
+    ref = pm_object.ref
+    var = pm_object.var
+    arc_ix = pm_object.ext[:arc_ix]
 
-    add_variable(ps_m,
-                 devices,
-                 time_steps,
-                 Symbol("Fbr_to_P_$(B)"),
-                 false)
-    add_variable(ps_m,
-                 devices,
-                 time_steps,
-                 Symbol("Fbr_fr_P_$(B)"),
-                 false)
+    var_name_from_p = Symbol("Pbr_fr_$(B)")
+    var_name_to_p = Symbol("Pbr_to_$(B)")
+    var_name_from_q = Symbol("Qbr_fr_$(B)")
+    var_name_to_q = Symbol("Qbr_to_$(B)")
 
-    add_variable(ps_m,
-                 devices,
-                 time_steps,
-                 Symbol("Fbr_to_Q_$(B)"),
-                 false)
-    add_variable(ps_m,
-                 devices,
-                 time_steps,
-                 Symbol("Fbr_fr_Q_$(B)"),
-                 false)
+    ps_m.variables[var_name_to_p] = PSI._container_spec(ps_m.JuMPmodel,
+                                                       (d.name for d in devices),
+                                                       time_steps)
+
+    ps_m.variables[var_name_from_p] = PSI._container_spec(ps_m.JuMPmodel,
+                                                          (d.name for d in devices),
+                                                          time_steps)
+
+
+    ps_m.variables[var_name_to_q] = PSI._container_spec(ps_m.JuMPmodel,
+                                                       (d.name for d in devices),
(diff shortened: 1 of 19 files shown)


ccoffrin:
I will have a chance to review this in the next day or so.

src/devices/device_models/AC_branches.jl
```diff
@@ -26,89 +26,252 @@ function flow_variables(ps_m::CanonicalModel,
...
 
+function flow_variables(ps_m::CanonicalModel,
+                        system_formulation::Type{S},
+                        devices::PSY.FlattenedVectorsIterator{B},
+                        time_steps::UnitRange{Int64}) where {B <: PSY.ACBranch,
+                                                             S <: PM.AbstractPowerFormulation}
+    #Get PowerModels dicts
+    pm_object = ps_m.pm_model
+    ref = pm_object.ref
+    var = pm_object.var
```
ccoffrin: Instead of accessing the internal ref and var objects.  I recomend to use the ref and var functions (e.g. `var(pm_object, t, :p)`).  These allow you to ignore more advanced PowerModels features, like conductors.

src/devices/device_models/AC_branches.jl
```diff
@@ -26,89 +26,252 @@ function flow_variables(ps_m::CanonicalModel,
...
+            push!(ref[:nw][t][:arcs_to], arcs_to)
+            #careful here the order of the push has to match the order of the variable creation
+            push!(ref[:nw][t][:arcs], arcs_fr, arcs_to)
+
+            #Active Power Variables
+            ps_m.variables[var_name_from_p][d.name,t] = JuMP.@variable(ps_m.JuMPmodel,
+                                                                            base_name="$(bus_fr),$(bus_to)_{$(d.name),$(t)}",
+                                                                            upper_bound = d.rate,
+                                                                            lower_bound = -d.rate,
+                                                                            start = 0.0)
```
ccoffrin: Instead of creating these variables yourself, why not let PowerModels create them and extract them into your data structure?

src/devices/device_models/common/pm_translator.jl
```diff
@@ -143,13 +143,19 @@ function get_branches_to_pm(sys::PSY.System)
...
         "zone"     => 1,
         "bus_i"    => bus.number,
-        "bus_type" => 2,
+        if bus.bustype == "PV"
+            "bus_type" => 1
+        elseif bus.bustype == "PQ"
+            "bus_type" => 2
+        elseif bus.bustype == "SF"
+            "bus_type" => 3
+        end,
```
ccoffrin: Missing type 4?

src/network/network_models/powermodels_interface.jl
```diff
@@ -62,17 +144,14 @@ end
...
+    return PM.build_generic_model(data, model_constructor, post_nip_expr!; multinetwork=multinetwork, kwargs...)
 end
 
 ""
-function post_nip_expr(pm::PM.GenericPowerModel)
+function post_nip_expr!(pm::PM.GenericPowerModel)
     for (n, network) in PM.nws(pm)
         @assert !PM.ismulticonductor(pm, nw=n)
         PM.variable_voltage(pm, nw=n)
-        PM.variable_branch_flow(pm, nw=n)#, bounded=false)
```
ccoffrin: I would not recomend removing this.

src/network/network_models/powermodels_interface.jl
```diff
@@ -255,20 +334,26 @@ function powermodels_network!(ps_m::CanonicalModel,
...
+    ref = ps_m.pm_model.ref
     _remove_undef!(ps_m.expressions[:nodal_balance_active])
     _remove_undef!(ps_m.expressions[:nodal_balance_reactive])
 
-    for t in time_steps, bus in buses
-        pm_data["nw"]["$(t)"]["bus"]["$(bus.number)"]["pni"] = ps_m.expressions[:nodal_balance_active][bus.number,t]
-        pm_data["nw"]["$(t)"]["bus"]["$(bus.number)"]["qni"] = ps_m.expressions[:nodal_balance_reactive][bus.number,t]
+    for t in time_steps
+        var[:nw][t][:cnd][1][:p] = JuMP.Containers.DenseAxisArray(var[:nw][t][:cnd][1][:p],ref[:nw][t][:arcs])
+        var[:nw][t][:cnd][1][:q] = JuMP.Containers.DenseAxisArray(var[:nw][t][:cnd][1][:q],ref[:nw][t][:arcs])
```
ccoffrin: It is not clear to me why this mapping is happening here.

jd-lara:
The objective of this PR is to create a way to make the branch flow constraints in PowerSimulations such that we can impose other constraints like flow schedules in interties or make branch models for controllable branches like tap changer transformers or angle dispatch with phase shifters. The process to integrate the constraints and variables from PowerModels into PowerSimulations follows this procedure in the current version of v0.2.0. 

1. We load the data from the System struct into a PowerModels dict, and we loop over the branches once here. 
2. We then call the `GenericPowerModel` with the formulation spec and returns an anonymous function we use in the `build_generic_model` function call. 
3. `build_generic_model` runs `core_ref!` and goes over the data again to makes the ref fields. After the `build_generic_model` call the `post_nip_expr method` that goes over the data again to make the variables. 
4. *If we reference the variables in PowerSimulations, we would go over the branches and time_steps once more and loop over them again impose further constraints if we need it.  

The idea is to take the previous process and split it in two to integrate it better as follows: 

1. Remove the anonymous function creation and instantiate the PowerModels object with the buses only in the `canonical_model` build,  add the place holder for the variables and constraints similar to what the `GenericPowerModels` function call does. Here part a part of `core_ref!` got added such that there is a place holder for the archs when they get added during the variable creation. Maybe this is not the best place to do it, but I need the keys in the `ref` dict before going into the variable creation. 
2. When we get to creating the device models, the code loops over the branches only once. In that loop the data is added to the `ref` dict directly, generate the arcs for the constraint creation, create the variables and reference them both in the PowerSimulations `canonical_model` and in the PowerModels var dict that was pre-generated in step 1. The idea is that by doing this, we can avoid step 4. 
3. Once this is done, call the remaining routines of `core_ref!` (a simplified version is what I added in `psi_ref!` that is needed for the integration and run the `post_nip_expr` but now the flow variables have been already created, and the arcs added accordingly. 

@ccoffrin is correct that this exposes a lot of the PowerModels complexity in the `add_flow_variable` calls and it is not super clean. The idea is to remove the repeated loops over the branches data and variables.

src/devices/device_models/common/pm_translator.jl
jd-lara: We don't have islanded buses in PowerSystems.Bus currently.

src/devices/device_models/AC_branches.jl
jd-lara: solved via b9d9919

ccoffrin:
A few more questions based on your response.

Re flow schedules on interties:  Why not do the following, let power models build the flow variables, then you impose extra constraints on them?

Re tap changer transformers or angle dispatch with phase shifters: I would recommend we just add this type of component as a feature to PowerModels, which you can then use through the data model.  Based on the types of problems you have discussed in the past, I highly doubt PowerSimulations wants to get into the complexities of modeling tap changers; It gets particularly complex once you consider variants of complex power beyond AC Polar.  Phase shifters in the DC model are fairly easy to add some some models (e.g. DC), but following the previous devision of functionality, this would be best implemented in PowerModels.

jd-lara:
> Re flow schedules on interties: Why not do the following, let power models build the flow variables, then you impose extra constraints on them?

I wanted to avoid looping over the branches twice. This created some uncleanliness in the integration.  
 
> Re tap changer transformers or angle dispatch with phase shifters: I would recommend we just add this type of component as a feature to PowerModels, which you can then use through the data model. Based on the types of problems you have discussed in the past, I highly doubt PowerSimulations wants to get into the complexities of modeling tap changers; It gets particularly complex once you consider variants of complex power beyond AC Polar. Phase shifters in the DC model are fairly easy to add some some models (e.g. DC), but following the previous devision of functionality, this would be best implemented in PowerModels.

It's true it will be possible to contribute those models to PowerModels instead of making them ourselves. 

Overall I think @ccoffrin comments make sense, I can make a referencing using the accessor functions 
between the `canonical_model` and the `pm_model` to add more constraints to the flows when necessary. 

@claytonpbarrows if you are ok, I will close this PR and then try a new variant.

claytonpbarrows:
sounds good to me

PR #172, opened 2019-07-10
Jd/sim update

mkratochvil:
Made minor changes to device_commitment.jl, and made duration_constraints.jl more readable.


src/devices/device_models/common/duration_constraints.jl
```diff
@@ -13,11 +13,14 @@ function device_duration_retrospective(ps_m::CanonicalModel,
     time_steps = model_time_steps(ps_m)
     name_up = Symbol(cons_name,:_up)
     name_down = Symbol(cons_name,:_down)
-
-    ps_m.constraints[name_up] = JuMPConstraintArray(undef, set_names, time_steps)
-    ps_m.constraints[name_down] = JuMPConstraintArray(undef, set_names, time_steps)
-
-
+    var1 = var(ps_m, var_names[1])
```
jd-lara: Let's give the variables names that are a bit more representative, in my original example I didn't do it but in a second look at it probably is best to given them some sensical name (not a single letter). Also, if you can add a Docstring with the function constraint represented in common that would be great.

src/devices/device_models/common/range_constraint.jl
```diff
@@ -69,25 +79,30 @@ function reserve_device_semicontinuousrange(ps_m::CanonicalModel,
     time_steps = model_time_steps(ps_m)
     ub_name = Symbol(cons_name,:_ub)
     lb_name = Symbol(cons_name,:_lb)
+    
+    var1 = var(ps_m, var_name)
+    varb = var(ps_m, binvar_name)
```
jd-lara: use `varbin`

jd-lara:
@mkratochvil We applied some changes that mean that some conflicts need to be fixed before merging.

jd-lara:
@mkratochvil can you fix this conflict and then we can merge

mkratochvil:
Finished fixing the conflict. Docstrings non-duration constraints look the way I want, and all constructor tests (except for UC) are passing so I think things are ready on my end.

jd-lara:
Thanks, I will merge now

PR #177, opened 2019-07-15
resolving variable name retrevial issue

claytonpbarrows:


src/devices/device_models/common/add_variable.jl
```diff
@@ -12,8 +12,10 @@ function add_variable(ps_m::CanonicalModel,
    ps_m.variables[var_name] = _container_spec(ps_m.JuMPmodel, (PSY.get_name(d) for d in devices), time_steps)
```
claytonpbarrows: @jd-lara could we do this in the below for loop? It seems like we are looping through the devices twice...

jd-lara: The `()` is just a generator it actually doesn't allocate any memory. The issue here is that we need the axis to allocate the array before looping.

src/devices/device_models/AC_branches.jl
```diff
@@ -44,9 +44,10 @@ function flow_variables(ps_m::CanonicalModel,
         cp = PSY.get_arch(d)
         bus_fr = cp.from.number
         bus_to = cp.to.number
+        name = PSY.get_name(d)
         for t in time_steps
-            ps_m.variables[var_name][PSY.get_name(d),t] = JuMP.@variable(ps_m.JuMPmodel,
-                                                                base_name="$(bus_fr),$(bus_to)_{$(PSY.get_name(d)),$(t)}",
+            ps_m.variables[var_name][name,t] = JuMP.@variable(ps_m.JuMPmodel,
+                                                                base_name="$(bus_fr),$(bus_to)_{$(name),$(t)}",
                                                                 upper_bound = PSY.get_rate(d),
```
jd-lara: Same here. `rate = get_rate(d)`

src/devices/device_models/common/add_variable.jl
daniel-thom: It does eventually have to be evaluated (iterate through the loop), so if there is a way to do it in the following for loop, that would be better.

daniel-thom: Is this code even performance-sensitive?

jd-lara: I can look into it but it is not an obvious solution. The only way is to evaluate it inside of the macro and specify it in a way that the container is created inside of the macro.

jd-lara: > Is this code even performance-sensitive?

The allocation of the Array isn't the creation of the @constraint it is. The only way out of this double call is to make it inside on a long macro call. I haven't looked into it since the update to JuMP v0.19.

jd-lara:
@claytonpbarrows you might need to rebase this branch, I made some changes on whitespace to have space after a comma.

PR #182, opened 2019-07-22
Unit tests for demand-response functions

bwbush:
Removed erroneous underbar, changing `export demand_constraints` to `export demandconstraints`.


bwbush:
This is obsolete.

PR #184, opened 2019-07-22
Duration Constraints Fixes

sourabhdalvi:
- Retrospective Constraint supports both Parameter or Float initial conditions 
- Mixed and look-ahead Minor fix in inequality sign

### DIFF
--- a/src/devices/device_models/common/duration_constraints.jl
+++ b/src/devices/device_models/common/duration_constraints.jl
@@ -20,37 +20,30 @@ function device_duration_retrospective(ps_m::CanonicalModel,
 
 
         for t in time_steps, (ix,name) in enumerate(set_names)
-            if t - duration_data[ix].up >= 1
-                tst = duration_data[ix].up
-            else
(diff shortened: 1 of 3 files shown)


src/devices/device_models/common/duration_constraints.jl
```diff
@@ -126,30 +119,34 @@ function device_duration_look_ahead(ps_m::CanonicalModel,
...
 
 
-    for t in time_steps[2:end], (ix,name) in enumerate(set_names)
+    for t in time_steps, (ix,name) in enumerate(set_names)
         # Minimum Up-time Constraint
         lhs_on = JuMP.GenericAffExpr{Float64, _variable_type(ps_m)}(0);
         for i in t-duration_data[ix].up:t
-            if in(time_steps,i)  
-                JuMP.add_to_expression!(lhs_on,1,ps_m.variables[var_names[1]][name,i])
+            if in(i,time_steps)
```
jd-lara: doing for `i in time_steps` will be easier to read.

sourabhdalvi: Sure, I'll make that change

PR #192, opened 2019-08-01
Test Updates

### LINKED ISSUE #189: fix tests
- [x] thermal gen (data) @jd-lara 
 - [x] renewable gen @jd-lara 
 - [x] load @jd-lara 
 - [x] storage @sourabhdalvi 
 - [x] hydro  @sourabhdalvi 
 - [x] network @claytonpbarrows 
 - [x] operation constructor @claytonpbarrows
 - [x] operation solve @claytonpbarrows

sourabhdalvi:
closes #189

### DIFF
--- a/src/routines/get_results.jl
+++ b/src/routines/get_results.jl
@@ -23,11 +23,10 @@ function _result_dataframe_d(constraint::JuMP.Containers.DenseAxisArray)
     names = Array{Symbol, 1}(undef, length(constraint.axes[1]))
 
     for (ix, name) in enumerate(constraint.axes[1])
-
-        result[ix] = JuMP.dual(constraint[name])/100
-
-        # names[ix] = Symbol(name)
-
+        try result[ix] = JuMP.dual(constraint[name])
+        catch
+            result[ix] = NAN
+        end
     end
 
     return DataFrames.DataFrame(Price = result)
--- a/src/routines/get_results.jl
+++ b/src/routines/get_results.jl
@@ -16,6 +16,24 @@ function _result_dataframe(variable::JuMP.Containers.DenseAxisArray)
 
 end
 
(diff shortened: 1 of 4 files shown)


src/routines/get_results.jl
```diff
@@ -16,6 +16,24 @@ function _result_dataframe(variable::JuMP.Containers.DenseAxisArray)
...
 
+function _result_dataframe_d(constraint::JuMP.Containers.DenseAxisArray)
+
+    result = Array{Float64, length(constraint.axes)}(undef, length(constraint.axes[1]))
+    # TODO: Remove this line once PowerSystems moves to Symbols
+    names = Array{Symbol, 1}(undef, length(constraint.axes[1]))
+
+    for (ix, name) in enumerate(constraint.axes[1])
+
+        result[ix] = JuMP.dual(constraint[name])/100
```
jd-lara: We shouldn't do the hardcoded division here. The interpretation of the dual should depend on the user.

src/routines/get_results.jl
```diff
@@ -16,6 +16,24 @@ function _result_dataframe(variable::JuMP.Containers.DenseAxisArray)
 
 end
 
+function _result_dataframe_d(constraint::JuMP.Containers.DenseAxisArray)
```
jd-lara: We should add to this function a try - catch when JuMP throws an error because it can't find the dual.

src/routines/get_results.jl
sourabhdalvi: Sure, I can remove that.

src/routines/get_results.jl
sourabhdalvi: Do you want it to be like this ??
```julia
try result[ix] = JuMP.dual(constraint[name])
catch
    result[ix] = NAN
end
```

PR #193, opened 2019-08-06
Reading and Writing results to feather files

Lilyhanig:
The OperationModelResults spelling was corrected. The write_model.jl was moved to src/routines and some write functions from get_results.jl were moved to write_model.jl. Functions were created to write automatically in the solve_op_model to feather files (with a key word argument) or write after solving to feather files. A function was added to results_model.jl to read back to memory from the written feather files.


Project.toml
```diff
@@ -5,17 +5,24 @@ version = "0.1.3"
...
 JuMP = "4076af6c-e467-56ae-b986-b466b2749572"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MathOptFormat = "f4570300-c277-12e8-125c-4912f86ce65d"
 MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
+Memento = "f28f55f0-a522-5efc-85c2-fe41dfb9b2d9"
 ParameterJuMP = "774612a8-9878-5177-865a-ca53ae2495f9"
 Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 PowerModels = "c36e90e8-916a-50a6-bd94-075b64ef4655"
 PowerSystems = "bcd98974-b02a-5e2f-9ee0-a103f5c450dd"
+Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
```
jd-lara: Why was Revise added to the project? Some of these packages are used for testing and development and don't make part of the main dependencies. Please remove them.

src/core/core_structs/results_model.jl
```diff
@@ -1,13 +1,51 @@
...
+    return res_model.optimizer_log 
+end
+
+function get_times(res_model::OperationModelResults, key::Symbol)
+    return res_model.times
 end
+
+# passing in name of folder path and the name of the folder
+ 
+function operation_model_read(path::AbstractString, directory::AbstractString)
```
jd-lara: Let's make sure this function reflects that the files being read are from the results. Something like 
```load_model_results```

File: src/routines/write_model.jl [hunk 3]
```diff
@@ -0,0 +1,122 @@
...
+end
+
+function write_memory_results(results::OperationModelResults, save_path::String)
+
+    new_folder = mkdir("$save_path/$(round(Dates.now(),Dates.Minute))")
+    folder_path = new_folder
+    write_variable_results(results.variables, folder_path) 
+    write_optimizer_results(results.optimizer_log, folder_path)
+    write_time_results(results.times, folder_path)
+    return println("Files written to $folder_path.")
```
jd-lara: Don't add a print statement in a return.

src/routines/write_model.jl
```diff
@@ -0,0 +1,122 @@
...
+    file_path = joinpath(save_path,"optimizer_log.feather")
+    Feather.write(file_path, df)
+    # println("feather file written to $file_path")
+    
+    return
+end
+
+# taking the outputted files for the time_Series DataFrame and writing them to a featherfile
+
+function write_time_results(time_stamp::DataFrames.DataFrame, save_path::AbstractString)
```
jd-lara: let's change the name to `write_timestamps`

src/PowerSimulations.jl
```diff
@@ -54,6 +54,7 @@ export construct_device!
 export construct_network!
 ## Op Model Exports
 export solve_op_model!
+export write_memory_results
```
jd-lara: rename to `_write_variables` and don't export it.

src/core/core_structs/results_model.jl
```diff
@@ -1,13 +1,51 @@
...
     variables::Dict{Symbol, DataFrames.DataFrame}
-    total_cost::Dict{Symbol,Float64}
+    total_cost::Dict{Symbol, Any}
     optimizer_log::Dict{Symbol, Any}
+    times::DataFrames.DataFrame
+
 end
 
-function get_variable(res_model::OpertationModelResults, key::Symbol)
+function get_variable(res_model::OperationModelResults, key::Symbol)
```
jd-lara: Add a message if the variable is not found.

src/routines/get_results.jl
```diff
@@ -60,19 +64,7 @@ function get_model_duals(op_m::OperationModel, cons::Vector{Symbol})
 
 end
 
-function write_model_result(op_m::OperationModel, path::String)
```
jd-lara: Is this the new `_write_variable`  code?

src/routines/get_results.jl
```diff
@@ -91,42 +83,18 @@ function get_optimizer_log(op_m::OperationModel)
...
-
-    Feather.write(file_path, df)
+# Function to create a dictionary for the time series of the simulation
 
-    return
+function get_time_stamp(op_model::OperationModel)
 
+    initial_time = PSY.get_forecast_initial_times(op_model.sys)[1]
+    interval = PSY.get_forecasts_resolution(op_model.sys)
+    range = collect(initial_time:interval:initial_time+ Dates.Hour(24))
```
jd-lara: Why is this range hardcoded? the range values have to be read from the OperationModel. 

```range = collect(initial_time:interval:initial_time + horizon*interval)```

File: src/routines/solve_routines.jl [hunk 9]
```diff
@@ -1,37 +1,68 @@
 """ Solves Operational Models"""
-function solve_op_model!(op_model::OperationModel; kwargs...)
 
-    timed_log = Dict{Symbol, Any}()
+function _write_op_model(results::OperationModelResults, save_path::String)
+    
+    if !isdir(save_path) 
 
-    if op_model.canonical.JuMPmodel.moi_backend.state == MOIU.NO_OPTIMIZER
+        @error "Specified path is not valid. Run write_memory_results to save results."
```
jd-lara: use error() instead of @error.

src/routines/solve_routines.jl
```diff
@@ -1,37 +1,68 @@
...
+        println("Files written to $folder_path folder.")
+    
+    end
 
-        else
-            _, timed_log[:timed_solve_time],
-            timed_log[:solve_bytes_alloc],
-            timed_log[:sec_in_gc] = @timed JuMP.optimize!(op_model.canonical.JuMPmodel,
-                                                          kwargs[:optimizer])
+return results
```
jd-lara: There is no need to have a return here. The function should only write to disk the results.

src/routines/write_model.jl
```diff
@@ -0,0 +1,122 @@
...
+    df = DataFrames.DataFrame(time_stamp)
+    file_path = joinpath(save_path,"time_stamp.feather")
+    Feather.write(file_path, df)
+    
+    return
+end
+
+# These functions are writing directly to the feather file and skipping printing to memory.
+
+function write_model_result(ps_m::CanonicalModel, path::String)
```
jd-lara: This method can be eliminated.

src/routines/write_model.jl
```diff
@@ -0,0 +1,122 @@
...
+    new_folder = mkdir("$save_path/$(round(Dates.now(),Dates.Minute))")
+    folder_path = new_folder
+    write_variable_results(results.variables, folder_path) 
+    write_optimizer_results(results.optimizer_log, folder_path)
+    write_time_results(results.times, folder_path)
+    return println("Files written to $folder_path.")
+end
+# taking the outputted files for the variable DataFrame and writing them to a featherfile
+
+function write_variable_results(vars_results::Dict{Symbol, DataFrames.DataFrame}, save_path::AbstractString)
```
jd-lara: Why do we have this method if there is a `write_variable_results` already.

src/utils/printing.jl
```diff
@@ -10,6 +10,10 @@ function Base.show(io::IO, op_model::Simulation)
     println(io, "Simulation Model")
 end
 
-function Base.show(io::IO, res_model::OpertationModelResults)
-    println(io, "Results Model")
-end
+# function Base.show(io::IO, res_model::OperationModelResults)
```
jd-lara: Keep the show method until we have better printing.

src/PowerSimulations.jl
Lilyhanig: This is the function that writes the results in memory to a feather file (if no path was included as a kwarg or if the path was invalid). It's called within the development file I'm using, so it's not actually an internal function. Should I still rename it?

jd-lara: Then let's call it `write_results` it is more intuitive.

src/routines/get_results.jl
Lilyhanig: this code was already in write model, it *I believe* is what is being used to right directly from the multi-day simulations to feather files, so I haven't touched it.

src/routines/write_model.jl
Lilyhanig: I'm confused. I only see the one method for write_variable_results? It's called in each of the write results functions. Could you clarify?

jd-lara: It looks like we have 2 write results functions. 
1. that takes a canonical. 
2. that takes vars results

Lilyhanig: yes. the one that takes the canonical was already written for the case that nothing is outputted and all results are written straight to feather files (like in a sequential scenario). I wrote the vars_results function so that the variable results can be written to a feather file without the whole simulation running again. Does this make sense? I'm not using the write_results with the canonical_model at all, but I also didn't think I should delete it.

PR #194, opened 2019-08-07
Jd/sim update

mkratochvil:
This should be only updating cost_functions.jl, duration_constraints.jl, and timeseries_constraints.jl in common, but I had to update my fork to get this to work. 

The update to TimeZones is not letting me get PowerSimulations to work on my laptop so could someone check that the docstrings look fine? (My changes did not break any of the tests.) Also, @sourabhdalvi, could you check line 313 of duration_constraints.jl? I believe there is a typo.


src/devices/device_models/common/duration_constraints.jl
```diff
@@ -145,19 +303,17 @@ function device_duration_param(ps_m::CanonicalModel,
...
             else
-                JuMP.add_to_expression!(lhs_off, ps_m.variables[var_names[3]][name, i])
+                JuMP.add_to_expression!(lhs_off, varstop[name, i])
             end
         end
         if t <= duration_data[ix].down
-            lhs_off += initial_duration_on[ix].value
-            ps_m.constraints[name_down][name, t] = JuMP.@constraint(ps_m.JuMPmodel,
-             ps_m.variables[var_names[2]][name, t]*duration_data[ix].down - lhs_off <= 0.0)
+            lhs_off += initial_duration_on[ix].value #typo? off not on?
```
sourabhdalvi: Yes, that should be `initial_duration_off[ix].value`. Could you make the change.
Thanks

mkratochvil: The change has been made!

PR #196, opened 2019-08-15
plot recipes

Lilyhanig:
The creation of the structs that the three plot recipes look for are in results_model.jl. The plot recipes are in plot recipes.jl and below is an example of the use cases for the recipes:
"
    using Plots 
    using RecipesBase
    
    gr()
    
    Order_therm = [:Brighton, :Solitude, Symbol("Park City"),  :Alta, :Sundance]
    Order_renew = [:WindBusC, :WindBusB, :WindBusA]
    Order_gen = [:P_ThermalStandard, :P_RenewableDispatch]
    
    stacked_renew = get_stacked_plot_data(res, "P_RenewableDispatch"; sort = Order_renew) #generates data for stacked plot
    plot(stacked_renew, "P_RenewableDispatch")
    
    stacked_therm = get_stacked_plot_data(res, "P_ThermalStandard"; sort = Order_therm) #generates data for stacked plot
    plot(stacked_therm, "P_ThermalStandard")
    
    bar_renew = get_bar_plot_data(res, "P_RenewableDispatch"; sort = Order_renew)
    plot(bar_renew, "P_RenewableDispatch")
    
    bar_therm = get_bar_plot_data(res, "P_ThermalStandard"; sort = Order_therm)
    plot(bar_therm, "P_ThermalStandard")
    
    stacked_gen = get_stacked_generation_data(res; sort = Order_gen) #generates data for stacked plot
    plot(stacked_gen)
    
"
    The sort kwarg allows the user to define the order for plotting, the default is alphabetical.
Each function takes the generic results and outputs a data struct compatible with the desired plot recipe. The first recipe makes a per generator stack plot, the second a per generator bar plot, and the last a per variable stack plot.


results_extraction/PSI Dev.jl
```diff
@@ -0,0 +1,137 @@
+ cd "~"/.julia/dev/PowerSimulations
```
jd-lara: This file shouldn't be included in the PR. This is your personal file for development. 
For future reference, these files shouldn't be kept in the same folder as the package you are developing.

src/core/core_structs/results_model.jl
```diff
@@ -6,6 +6,28 @@ struct OperationModelResults
 
 end
 
+struct StackedArea
```
jd-lara: These structs should be in the same file as the plots recipes.

src/core/core_structs/results_model.jl
```diff
@@ -53,4 +75,75 @@ function load_operation_results(path::AbstractString, directory::AbstractString)
 
     return results
 
+end
+
+function get_stacked_plot_data(res::OperationModelResults, variable::String; kwargs...)
```
jd-lara: These functions as well should be moved to a plots file in utils.

src/utils/plot_recipes.jl
```diff
@@ -0,0 +1,117 @@
+
+
+@recipe function StackedPlot(res::StackedArea, variable::String)
```
jd-lara: These recipes should internally call the get_data functions and not requires to functions to see the results. I see the value in having functions that retrieve results in a particular format to use other plotting tools but these ones should be in a single call.

src/utils/printing.jl
```diff
@@ -9,11 +9,12 @@ end
 function Base.show(io::IO, op_model::Simulation)
     println(io, "Simulation Model")
 end
-
- function Base.show(io::IO, res_model::OperationModelResults)
+ #=
```
jd-lara: Please uncomment these lines, we need to develop proper show methods here but the code shouldn't be commented out when you make the PR.

results_extraction/PSI Dev.jl
Lilyhanig: oops, don't know how this ended up in here!

PR #199, opened 2019-08-22
Fix documenter

jd-lara:


docs/make.jl
```diff
@@ -1,3 +1,5 @@
+import Pkg;
+Pkg.add("Documenter")
```
kdheepak: You don't need these lines if you already have it in `docs/Project.toml`, right?

.travis.yml
```diff
@@ -52,9 +52,9 @@ after_success:
 jobs:
   include:
     - stage: "Documentation"
-      julia: 1.0
+      julia: 1.1
```
kdheepak: You don't have to change it in this PR but I believe if you use `julia: 1` it'll always use the latest stable version. If you want to ensure the documentation is always built on the latest stable version you can do that. Typically, I fix every version for something like documentation so that I can ensure it's always possible to build in the future, i.e. pin dependencies in Project.toml, use a fixed version for Julia etc, even if it's an older version of the software.

docs/Project.toml
```diff
@@ -1,2 +1,5 @@
 [deps]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+DocumenterTools = "35a29f4d-8980-5a13-9543-d66fff28ecb8"
+PowerSimulations = "e690365d-45e2-57bb-ac84-44ba829e73c4"
+PowerSystems = "bcd98974-b02a-5e2f-9ee0-a103f5c450dd"
```
kdheepak: Why is PowerSystems here? Isn't `PowerSystems` a dependency of `PowerSimulations`? Also if you have `Pkg.build("PowerSimulations");` in the `.travis.yml` file, why do you need it here in the `docs/Project.toml` file again? Just trying to understand what is going on here.

.travis.yml
jd-lara: ok. I will make this change.

docs/Project.toml
jd-lara: I am also not completely clear about why this happens. Without this dependency the docs code throws a dependency error. I noticed PowerModels had to include InfrastructureModels as well.

docs/make.jl
jd-lara: I don't know. Couldn't get the package manager to handle it.

PR #200, opened 2019-08-23
Initial Condition Fix

sourabhdalvi:
Fixes Implementation of initial Conditions in
- Duration Constraints
- Commitment Constraints
- Ramping Constraints


src/devices/device_models/thermal_generation.jl
```diff
@@ -197,10 +197,17 @@ function commitment_constraints!(canonical_model::CanonicalModel,
...
     key = Symbol("status_$(T)")
-
-    if !(key in keys(canonical_model.initial_conditions))
+    
+    if !(key in keys(canonical_model.initial_conditions)) 
         @warn("Initial status conditions not provided. This can lead to unwanted results")
         status_init(canonical_model, devices)
+    else
+        status_miss = missing_init_cond(canonical_model.initial_conditions[key], devices)
+        @show status_miss
```
jd-lara: There is an @show lingering here

src/devices/device_models/thermal_generation.jl
```diff
@@ -464,15 +520,37 @@ function time_constraints!(canonical_model::CanonicalModel,
...
+        duration_init_off(canonical_model, devices)
+    else
+        dur_off_miss = missing_init_cond(canonical_model.initial_conditions[key_off], devices)
+        if !isnothing(dur_off_miss)
+            @warn("Initial Conditions for Minimum Down Time constraints not provided. This can lead to unwanted results")
+            duration_init_off(canonical_model, dur_off_miss)
+        end
+    end
+
+    sort!(canonical_model.initial_conditions[key_on],by=x-> PSY.get_name(x.device))
```
jd-lara: can we avoid this sort?

sourabhdalvi: That the only way I found to make sure both the initial conditions arrays have the same ordering. I told Dan this problem and this is what he recommended. If there is a better way to do it, then I'll definitely change this

PR #203, opened 2019-09-03
SOS2 Cost implementation 

### LINKED ISSUE #173: Implement a fallback to either Linear Cost function or SOS2 based PWL Cost function

sourabhdalvi:
The model will check if linear PWL function can be implemented, if not it will use SOS-2 
closes #173

### DIFF
--- a/src/devices/device_models/common/cost_functions.jl
+++ b/src/devices/device_models/common/cost_functions.jl
@@ -116,6 +116,17 @@ function ps_cost(canonical_model::CanonicalModel,
 
 end
 
+@doc raw"""
+    _pwlparamcheck(cost_)
+
+Returns True/False depending on compatibility of the cost data with the linear implementation method
+
+Returns ```flag```
+
+# Arguments
+
+* cost_::PSY.VariableCost{NTuple{2, Float64}} : container for quadratic and linear factors
+"""
 function _pwlparamcheck(cost_)
     flag = true;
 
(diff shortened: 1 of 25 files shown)


src/devices/device_models/common/cost_functions.jl
```diff
@@ -129,15 +129,51 @@ function _pwlparamcheck(cost_)
     return flag
 end
 
+function _gen_cost(canonical_model::CanonicalModel,
+                    variable::JV,
+                    cost_component::Vector{NTuple{2, Float64}}) where {JV<:JuMP.AbstractVariableRef}
+    
+    if !_pwlparamcheck(cost_component) 
+        @warn("Data provide is not suitable for linear implementation of PWL cost, Using SOS-2 implementation") ;
```
jd-lara: The error message should be a bit more descriptive and also has a typo. how about 
```The cost function provided for device $() is not compatible with a linear PWL cost function. An SOS-2 formulation will be added to the model. This will result in additional binary variables added to the model.```

File: src/devices/device_models/common/cost_functions.jl [hunk 2]
```diff
@@ -129,15 +129,51 @@ function _pwlparamcheck(cost_)
...
+end
+
+
+function _pwlgencost_sos(canonical_model::CanonicalModel,
+        variable::JV,
+        cost_component::Vector{NTuple{2, Float64}}) where {JV<:JuMP.AbstractVariableRef}
+
+    gen_cost = JuMP.GenericAffExpr{Float64, _variable_type(canonical_model)}()
+    pwlvars = JuMP.@variable(canonical_model.JuMPmodel, [i = 1:length(cost_component)], 
+                            base_name = "{$(variable)}_{pwl}",
```
jd-lara: ```base_name = "{$(variable)}_{pwl-sos}"```

sourabhdalvi:
I'll make the suggested changes

sourabhdalvi:
> The implementation looks good to me. We should add the docstring to the cost function call to explain all of this logic and also mention that the SOS implementation is directly from MOI and add a reference to it.
> 
> We should also add tests to this in order to make sure that the cost function is correctly created.

I have created test for Ramping ,Duration & Cost function. I dont have enough time this week to get it all done

PR #204, opened 2019-09-05
plot_recipes cleaned up

Lilyhanig:


Project.toml
```diff
@@ -1,12 +1,14 @@
 name = "PowerSimulations"
 uuid = "e690365d-45e2-57bb-ac84-44ba829e73c4"
 authors = ["Jose Daniel Lara", "Dheepak Krishnamurthy", "Clayton Barrows"]
-version = "0.2.0"
+version = "0.1.3"
```
jd-lara: Why did this change? It seems you pulled the wrong Project.toml file. Please do not modify this manually in an attempt to fix it.

jd-lara:
@Lilyhanig this branch has commits already merged into the branch. Also, you added a *.jmd file into the commit history. There are a lot of things here that aren't relevant to the implementation of the plots. 

@kdheepak as the git-master what's the best approach here to merge only the relevant files?

PR #205, opened 2019-09-09
add plot recipes and results with sorting

Lilyhanig:


src/utils/plot_recipes.jl
```diff
@@ -1,108 +1,223 @@
...
-    #create filled polygon
-    sy = vcat(z[:,1],zeros(n-1))
-    sx = [time[1:n-1]; reverse(time[1:n-1])]
+for a results with 6 variables, each with 10 generators:
 
-    for c=1:size(z,2)
+bar_plot(results)
 
-      if c !== 1
+generates a bar plot for each variable, with 10 generators,
```
jd-lara: Why does it say here the number of generators?

PR #211, opened 2019-09-17
Reports

Lilyhanig:
reports functionality for vscode using Weave.


src/utils/make_report.jl
```diff
@@ -0,0 +1,37 @@
...
+report(results, jmd, out_path)
+
+kwargs: doctype = "md2html" to create an HTML
+default is PDF via latex
+
+"""
+
+
+
+function report(res::OperationModelResults, jmd::String, out_path::String; kwargs...)
```
jd-lara: we should have a default action for the jmd and re-write this function as 
```Julia
function report(res::OperationModelResults, out_path::String, jmd::String="deault_string")
```

src/utils/report_design.jmd
```diff
@@ -0,0 +1,35 @@
+---
```
jd-lara: Put this file in a new folder called report designs inside of utils

Issue #217, opened 2019-09-25
Question on idiom used in PowerSimulations


daniel-thom:
PowerSimulations frequently uses code like this:

```
var = 999.0*(PSY.get_activepower(g) > 0)
```

I'm not used to seeing multiplication between floats and bools and was confused the first time I saw it.  I would have written like this:

```
var = PSY.get_activepower(g) > 0 ? 999.0 : 0.0
```

@jd-lara @kdheepak  Is the former idiomatic for Julia developers?

jd-lara:
This is a short cut, ugly one. The value that has to go into JuMP has to be 0.0 or 999.0 depending on the status of the unit

kdheepak:
Use the latter.

daniel-thom:
The action is now to find all places where this exists and change them.

jd-lara:
closed via #218

PR #218, opened 2019-09-27
Pr comments

### LINKED ISSUE #213: Code creates directories with colons which aren't supported in Windows.
This happens in a few places.  Consider src/routines/write_models.jl write_model_results().  The timestamp will have colons.

```
new_folder = mkdir("$save_path/$(round(Dates.now(),Dates.Minute))")
```

Windows has paths like c:\data, so this will crash at runtime.

jd-lara:
fixes #213 and #217


src/devices/device_models/common/model_utils.jl
```diff
@@ -31,3 +31,7 @@ function _remove_underscore(original::Symbol)
 
     return parts[1]
 end
+
+function remove_char(s::String, char::String)
+    return replace(s, Regex("[$char]") => "")
```
daniel-thom: What do you think about replacing ":" with "-" or "_"?

src/routines/make_initial_conditions.jl
```diff
@@ -98,8 +99,9 @@ function duration_init(canonical_model::CanonicalModel,
             ini_conds = canonical_model.initial_conditions[key] = Vector{InitialCondition}(undef, length_devices)
 
             for (ix, g) in enumerate(devices)
-                times = [999.0*(PSY.get_activepower(g) > 0),  # Time on
-                         999.0*(PSY.get_activepower(g) <= 0)] # Time off
+                time_on = (PSY.get_activepower(g) > 0) ? 999.0 : 0.0
```
daniel-thom: Magic number.  Can 999.0 be defined somewhere?

src/devices/device_models/common/model_utils.jl
daniel-thom: Also, I think it would be better to use the normal, non-Regex version of the function.

jd-lara: The purpose is to have a function that can do several character removals and replacements. See 05912f8

src/routines/make_initial_conditions.jl
```diff
@@ -1,3 +1,5 @@
+const missing_initial_conditions_time_count = 999.0
```
daniel-thom: constants should be in the form of MISSING_INITIAL_CONDITIONS_TIME_COUNT

src/devices/device_models/common/model_utils.jl
daniel-thom: A docstring would make that clear. :)  Also, maybe the function should be plural.

PR #226, opened 2019-10-02
fixing time stamp isue

Lilyhanig:
this sets the correct timestamp for the simulation results

### DIFF
--- a/src/routines/get_results.jl
+++ b/src/routines/get_results.jl
@@ -126,3 +126,13 @@ function get_time_stamp(op_model::OperationModel)
 
     return time_stamp
 end
+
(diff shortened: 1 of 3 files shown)


src/routines/get_results.jl
```diff
@@ -126,3 +126,13 @@ function get_time_stamp(op_model::OperationModel)
 
     return time_stamp
 end
+
+function get_time_stamp(op_model::OperationModel, start_time::Dates.DateTime)
```
jd-lara: make it _get_time_stamp please

jd-lara:
Had to merge to do a fix that wouldn't conflict with this

PR #227, opened 2019-10-03
Jd/run stage with rebuld

### LINKED ISSUE #224: Is chronology synchronized properly?
I'm wondering if https://github.com/NREL/PowerSimulations.jl/blob/cdde344d74104d60cbe0aa2f76eb0298a2b80191/src/routines/solve_routines.jl#L91 should actually update with the `interval` of the prior stage, not the current stage?

For example, if stage one steps daily and stage two depends on stage one and steps every five minutes, then one would normally want 24 executions of stage two where the start of each of those executions differs by one hour.  However, in the current code that offset between each execution of stage two is five minutes.  The chronology checks prevent one from having stage two simulation the whole day.

jd-lara:
- Adds functionality to have a stage be re-built during the stage_update process. 
- Changes Chronology to include values. Resolves #224 
- Addresses more comments by @kdheepak


src/core/build_simulations.jl
```diff
@@ -17,7 +17,7 @@ function _prepare_workspace!(ref::SimulationRef, base_name::AbstractString, fold
     ref.raw = raw_ouput
     ref.models = models_json_ouput
     ref.results = results_path
-
+    cd(current_working_directory)
```
daniel-thom: This code block seems fishy.  Shouldn't it be `if !isdir`?  Seems like you could do this without changing directories.

src/core/core_structs/simulations_stages.jl
```diff
@@ -1,24 +1,31 @@
...
 
     function _Stage(key::Int64,
-                   model::OperationModel,
-                   executions::Int64,
-                   chronology_ref::Dict{Int64, Type{<:Chronology}},
-                   cache::Vector{<:AbstractCache})
+                    reference::ModelReference,
+                    op_model::Type{M},
+                    sys::PSY.System,
+                    canonical::CanonicalModel,
```
daniel-thom: An op_model has a canonical_model, right? If so, would it be better to only pass op_model here?

src/core/build_simulations.jl
```diff
@@ -80,17 +80,29 @@ function _build_stages(sim_ref::SimulationRef,
...
+                                    v.model.services,
+                                    v.sys,
+                                    v.optimizer,
+                                    verbose;
+                                    parameters = true,
+                                    kwargs...)
         stage_path = joinpath(sim_ref.models,"stage_$(k)_model")
         mkpath(stage_path)
-        write_op_model(op_mod, joinpath(stage_path, "optimization_model.json"))
+        write_op_model(canonical, joinpath(stage_path, "optimization_model.json"))
```
daniel-thom: If what this function really does is write the canonical model to disk then should it be renamed to "write_canonical_model"?  Or "write_model"?

src/core/build_simulations.jl
```diff
@@ -80,17 +80,29 @@ function _build_stages(sim_ref::SimulationRef,
 
     for (k, v) in stages
```
daniel-thom: If "v" is a Stage then can you call it "stage" instead?  It would be more clear.

src/core/build_simulations.jl
```diff
@@ -80,17 +80,29 @@ function _build_stages(sim_ref::SimulationRef,
...
+                                    kwargs...)
         stage_path = joinpath(sim_ref.models,"stage_$(k)_model")
         mkpath(stage_path)
-        write_op_model(op_mod, joinpath(stage_path, "optimization_model.json"))
+        write_op_model(canonical, joinpath(stage_path, "optimization_model.json"))
         system_to_file && IS.to_json(v.sys, joinpath(stage_path ,"sys_data.json"))
-        _populate_cache!(v.cache, op_mod)
-        mod_stages[k] = _Stage(k, op_mod, v.execution_count, v.chronology_ref, v.cache)
+        _populate_cache!(v.cache, canonical)
+        mod_stages[k] = _Stage(k,
```
daniel-thom: It looks like almost everything you need to construct _Stage is in a Stage. Could you just pass "v" to this constructor instead of all of these members?

jd-lara: Yes, the idea is to make only one. For now, I just needed more fields in _Stage to implement the capability of re-building a stage in each run.

src/core/core_structs/simulations_stages.jl
jd-lara: This has changes since the review

jd-lara:
@daniel-thom I addressed all the comments but also added some code to enable some functionalities that I needed over the weekend

src/PowerSimulations.jl
```diff
@@ -28,16 +28,16 @@ export HVDCDispatch
 export StaticPowerLoad
 export InterruptiblePowerLoad
 export DispatchablePowerLoad
-######## Renewable Formulations ########
+######## Renewable Forms ########
```
daniel-thom: This is not a great abbreviation.

src/core/build_cache.jl
```diff
@@ -1,5 +1,9 @@
 function build_cache!(cache::TimeStatusChange, op_model::OperationModel)
-    parameter = get_value(op_model.canonical, cache.ref)
+    build_cache!(cache, op_model.canonical)
+end
+
+function build_cache!(cache::TimeStatusChange, canonical::CanonicalModel)
```
daniel-thom: Would it be better to put these functions in the file that defines TimeStatusChange (hopefully time_status_change.jl).

src/core/build_simulations.jl
```diff
@@ -77,21 +75,32 @@ function _build_stages(sim_ref::SimulationRef,
 
     system_to_file = get(kwargs, :system_to_file, true)
     mod_stages = Vector{_Stage}(undef, length(stages))
-
-    for (k, v) in stages
+    for (k, stage) in stages
```
daniel-thom: Since you aren't using k it would be better to write:
`for cache in values(stage.cache)`

src/core/build_simulations.jl
```diff
@@ -77,21 +75,32 @@ function _build_stages(sim_ref::SimulationRef,
...
+                                    kwargs...)
         stage_path = joinpath(sim_ref.models,"stage_$(k)_model")
         mkpath(stage_path)
-        write_op_model(op_mod, joinpath(stage_path, "optimization_model.json"))
-        system_to_file && IS.to_json(v.sys, joinpath(stage_path ,"sys_data.json"))
-        _populate_cache!(v.cache, op_mod)
-        mod_stages[k] = _Stage(k, op_mod, v.execution_count, v.chronology_ref, v.cache)
-        sim_ref.date_ref[k] = PSY.get_forecast_initial_times(v.sys)[1]
+        _write_canonical_model(canonical, joinpath(stage_path, "optimization_model.json"))
+        system_to_file && IS.to_json(stage.sys, joinpath(stage_path ,"sys_data.json"))
```
daniel-thom: Is the data (System or time series) changing across stages.  This could be expensive and I'm wondering if you only need to do it once.  Or maybe if only the System is changing then perhaps we could provide a way to serialize it but not the time series data.

src/core/build_simulations.jl
```diff
@@ -77,21 +75,32 @@ function _build_stages(sim_ref::SimulationRef,
...
+        mod_stages[k] = _Stage(k,
```
daniel-thom: "k" is not very clear.  Is this stage_id?  stage_index?

src/core/core_structs/device_model.jl
```diff
@@ -1,6 +1,6 @@
-abstract type AbstractDeviceFormulation end
+abstract type AbstractDeviceForm end
```
daniel-thom: Again, not a great abbreviation. I prefer the original.

src/simulation_models/stage_update.jl
```diff
@@ -185,6 +208,8 @@ end
 #############################Interfacing Functions##########################################
 
 function cache_update!(stage::_Stage)
+
```
daniel-thom: Why did you add two blank lines?  As far as I can tell the Julia convention, like any other language, is it have no blank lines in between the function declaration and the first line.

src/core/build_cache.jl
jd-lara: This would be changed later when we do a file structure refactor.

src/core/build_simulations.jl
jd-lara: I am using the k to generate the paths.

src/core/build_simulations.jl
```diff
@@ -77,21 +75,32 @@ function _build_stages(sim_ref::SimulationRef,
...
+                                    parameters = true,
+                                    kwargs...)
         stage_path = joinpath(sim_ref.models,"stage_$(k)_model")
         mkpath(stage_path)
-        write_op_model(op_mod, joinpath(stage_path, "optimization_model.json"))
-        system_to_file && IS.to_json(v.sys, joinpath(stage_path ,"sys_data.json"))
-        _populate_cache!(v.cache, op_mod)
-        mod_stages[k] = _Stage(k, op_mod, v.execution_count, v.chronology_ref, v.cache)
-        sim_ref.date_ref[k] = PSY.get_forecast_initial_times(v.sys)[1]
+        _write_canonical_model(canonical, joinpath(stage_path, "optimization_model.json"))
```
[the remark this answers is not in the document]
jd-lara: This is writing the JuMP model inside the canonical model. Since we don't have references inside the JuMP object they are basically the same.

src/core/build_simulations.jl
jd-lara: This is done just once to record the state of the system used in the model. This is useful if the data was parsed and modified before running the simulation. It only happens once.

src/core/build_simulations.jl
jd-lara: Changed to key, _Stages to have keys as a field.

src/core/build_simulations.jl
```diff
@@ -61,10 +59,10 @@ function _get_dates(stages::Dict{Int64, Stage})
 
 end
 
-function _populate_cache!(cache_vector::Vector{<:AbstractCache}, op_model::OperationModel)
+function _populate_cache!(stage::_Stage)
 
-    for cache in cache_vector
-        build_cache!(cache, op_model)
+    for (k, cache) in stage.cache
```
daniel-thom: nit: you still aren't using k.

jd-lara: 381b630

src/core/core_structs/simulations_stages.jl
```diff
@@ -1,24 +1,31 @@
 ######## Internal Simulation Object Structs ########
 abstract type AbstractStage end
 
-mutable struct _Stage <: AbstractStage
+mutable struct _Stage{M<:AbstractOperationModel} <: AbstractStage
     key::Int64
```
daniel-thom: Just a comment that you can take or leave:  stage_id or stage_index seems more descriptive for this variable.

src/core/build_simulations.jl
```diff
@@ -77,21 +75,32 @@ function _build_stages(sim_ref::SimulationRef,
...
-    for (k, v) in stages
-        verbose && @info("Building Stage $(k)")
-        op_mod = OperationModel(DefaultOpModel, v.model, v.sys;
-                                optimizer = v.optimizer,
-                                parameters = true,
-                                verbose = verbose,
-                                kwargs...)
-        stage_path = joinpath(sim_ref.models,"stage_$(k)_model")
+    for (key, stage) in stages
+        verbose && @info("Building Stage $(key)")
```
daniel-thom: Not an immediate request for change, but I think you should rely on the logging configuration to control what gets logged rather than a boolean that you have to pass everywhere.

jd-lara: Ok, this can be addressed when we add the logging capabilities. This should be done for v0.2.0.

src/core/build_simulations.jl
```diff
@@ -77,21 +75,32 @@ function _build_stages(sim_ref::SimulationRef,
...
         mkpath(stage_path)
-        write_op_model(op_mod, joinpath(stage_path, "optimization_model.json"))
-        system_to_file && IS.to_json(v.sys, joinpath(stage_path ,"sys_data.json"))
-        _populate_cache!(v.cache, op_mod)
-        mod_stages[k] = _Stage(k, op_mod, v.execution_count, v.chronology_ref, v.cache)
-        sim_ref.date_ref[k] = PSY.get_forecast_initial_times(v.sys)[1]
+        _write_canonical_model(canonical, joinpath(stage_path, "optimization_model.json"))
+        system_to_file && IS.to_json(stage.sys, joinpath(stage_path ,"sys_data.json"))
+        mod_stages[key] = _Stage(key,
+                               stage.model,
```
daniel-thom: nit: Since I noticed you realigned things elsewhere, this is now out of alignment. Ignore if you want to wait until we discuss a new style guide formatter from Dheepak.

jd-lara: yeah, let's do this whenever we implement the auto format. More changes are coming when changing the PowerModels names so we would have to tackle spacing later.

src/core/core_structs/simulations_stages.jl
jd-lara: This is used in the code as stage.key. We would have then stage.stage_id.

test/test_load_constructors.jl
```diff
@@ -4,7 +4,7 @@
     op_model = OperationModel(TestOptModel, PM.DCPlosslessForm, c_sys5)
     @test_logs (:warn, warn_message) construct_device!(op_model, :Load, model);
     model = DeviceModel(PSY.PowerLoad, PSI.DispatchablePowerLoad)
-    warn_message = "The Formulation DispatchablePowerLoad only applies to Controllable Loads, \n Consider Changing the Device Formulation to StaticPowerLoad"
+    warn_message = "The Formulation DispatchablePowerLoad only applies to FormulationControllable Loads, \n Consider Changing the Device Formulation to StaticPowerLoad"
```
daniel-thom: IMHO this is unnecessarily testing an implementation detail. If some ever changes the warning message, like say removing the oddly placed "\n" then they will have to also change the test. It would be better to just test that a warning message occurred or perhaps the first few words of the text.

jd-lara: The objective is to test that the warning is thrown.

daniel-thom: But it is asserting that the text matches exactly.

PR #230, opened 2019-10-08
Write model

### LINKED ISSUE #215: Refactor write_model.jl
Suggest some refactoring in this file to make the functionality easier to read and reuse.

- [x] _write_variable_results() writes a dictionary of dataframes to files.  This could be made more generic. write_data could be a good name.  Different methods could dispatch to different implementations. One method could implement this logic. Another method could take a DataFrame and write that to a single file.  Secondly, we could consider supporting different file types.  This could be either a method parameter or just detecting the file type.  All of this functionality could go into InfrastructureSystems.
- [ ]  _write_optimizer_log() writes 3 integers and a timestamp to a DataFrame.  Is a JSON file more appropriate?  I suggest creating a struct with these fields and then serialize/deserialize with JSON.
- [x] _write_time_stamps() accepts a dataframe, creates a new dataframe from that, and then writes that to a feather file. The second dataframe doesn't appear necessary.  If InfrastructureSystems has a write_data(df::DataFrame, filename) method then this call site can call that instead. This function can be removed.
- [x] If MOI.get(canonical_model.JuMPmodel, MOI.SolveTime()) throws an exception in _export_optimizer_log() then the code writes an error message as a string to a variable that would otherwise get what seems like a timestamp or duration. This mixes types and adds complexity during parsing. A status field might be better.

Lilyhanig:
closes #215


jd-lara:
@Lilyhanig looks like there are some conflicts that need resolution with the master branch .

src/routines/write_model.jl
```diff
@@ -1,70 +1,75 @@
...
 
-    for (k,v) in vars(canonical)
-         file_path = joinpath(save_path,"$(k).feather")
-         Feather.write(file_path, _result_dataframe_vars(v))
+function write_data(vars_results::OperationModel, save_path::AbstractString; kwargs...)
+
+    file_type = get(kwargs, :file_type, Feather)
+    for (k,v) in vars(vars_results.canonical)
+         file_path = joinpath(save_path,"$(k).$file_type")
+         file_type.write(file_path, _result_dataframe_vars(v))
```
daniel-thom: I think it's unlikely that other formats will support the exact same write function as Feather.  And there is no error checking.  I suggest something like this instead:
`if file_type == Feather
    Feather.write(...)
else
    error("unsupported file type:  $file_type")
end

Another alternative is to discern the file type by looking at the file extension.  That is what I did here:  https://github.nrel.gov/Hosting-Capacity-Analysis/disco/blob/master/disco/utils/postprocess.py#L173, but I am sure that some people would not like that approach.

daniel-thom: If another developer read the function signature only then they might assume that they should pass "out.feather" instead of "out".  You could fix that with a docstring.  Or you could strip the extension if they pass one.

src/routines/write_model.jl
```diff
@@ -1,70 +1,75 @@
...
-function _write_variable_results(canonical::CanonicalModel, save_path::AbstractString)
 
-    for (k,v) in vars(canonical)
-         file_path = joinpath(save_path,"$(k).feather")
-         Feather.write(file_path, _result_dataframe_vars(v))
+function write_data(vars_results::OperationModel, save_path::AbstractString; kwargs...)
+
+    file_type = get(kwargs, :file_type, Feather)
+    for (k,v) in vars(vars_results.canonical)
+         file_path = joinpath(save_path,"$(k).$file_type")
```
daniel-thom: This will result in "out.Feather" instead of "out.feather"

src/routines/write_model.jl
Lilyhanig: I'm slightly confused, Wouldn't that if/else only support feather in that case? I see your point, but what about, say, CSV that is supported in this format. Should I do a try and catch statement? @daniel-thom

src/routines/write_model.jl
Lilyhanig: ^actually I tested it out, and it still resulted in a out.feather

src/routines/write_model.jl
daniel-thom: Right, you would have to add an ```elseif CSV``` code block.

src/routines/write_model.jl
```diff
@@ -1,70 +1,75 @@
...
-
 end
 
-function _write_optimizer_log(optimizer_log::Dict{Symbol, Any}, save_path::AbstractString)
-
-    optimizer_log[:termination_status] = Int(optimizer_log[:termination_status])
-    optimizer_log[:primal_status] = Int(optimizer_log[:primal_status])
-    optimizer_log[:dual_status] = Int(optimizer_log[:dual_status])
-    optimizer_log[:solve_time] = optimizer_log[:solve_time]
+function write_data(data::DataFrames.DataFrame, save_path::AbstractString, file_name::String; kwargs...)
```
daniel-thom: Some users might not notice that file_name should not have an extension.  You could add a docstring or strip it if they add one.

src/routines/write_model.jl
```diff
@@ -1,70 +1,75 @@
...
 end
 
-# taking the outputted files for the time_Series DataFrame and writing them to a featherfile
-function _write_time_stamps(time_stamp::DataFrames.DataFrame, save_path::AbstractString)
+function serialize_opt(optimizer_log::OptimizerLog)
 
-    df = DataFrames.DataFrame(time_stamp)
-    file_path = joinpath(save_path,"time_stamp.feather")
-    Feather.write(file_path, df)
+    return JSON.json(Int(optimizer_log))
```
daniel-thom: What does this return?

src/routines/write_model.jl
```diff
@@ -1,70 +1,75 @@
...
+end
 
+function _write_optimizer_log(optimizer_log::Dict, save_path::AbstractString)
+
+    OptimizerLog = OptimizerLog(optimizer_log[:termination_status], optimizer_log[:primal_status],
```
daniel-thom: Variables should be snake_case, not PascalCase.

src/routines/write_model.jl
```diff
@@ -1,70 +1,75 @@
...
+                               optimizer_log[:dual_status], optimizer_log[:solve_time])
+   
+    optimizer_log = DataFrames.DataFrame(serialize_opt(OptimizerLog))
+    file_path = joinpath(save_path,"optimizer_log.feather")
+    Feather.write(file_path, optimizer_log)
+    println("hello")
```
daniel-thom: Stale debug print

src/routines/write_model.jl
Lilyhanig: A dictionary of the Int within the enum, I was honestly slightly confused by making a JSON script, so I apologize if this wasn't proper

src/routines/write_model.jl
```diff
@@ -1,70 +1,75 @@
+struct OptimizerLog
+
+    termination_status::Enum
```
daniel-thom: What is Enum?

Lilyhanig: like BusType, I can change it

src/routines/write_model.jl
daniel-thom: I'm a bit confused on what your end goal is.  It looks like you create a DataFrame from this output and then write it to a Feather file.  I'd suggest that writing it to a JSON file would be better.  If that's what you want then here are two things you could try (JSON2 should figure out the types for you):
```
# produce a JSON-encoded string.
text = JSON2.write(optimizer_log)
# Write to a JSON file.
open("out.json, "w") do io
    JSON2.write(io, optimizer_log)
end
```

src/routines/write_model.jl
```diff
@@ -1,70 +1,75 @@
...
-function _write_time_stamps(time_stamp::DataFrames.DataFrame, save_path::AbstractString)
+function serialize_opt(optimizer_log::OptimizerLog)
 
-    df = DataFrames.DataFrame(time_stamp)
-    file_path = joinpath(save_path,"time_stamp.feather")
-    Feather.write(file_path, df)
+    return JSON.json(Int(optimizer_log))
+end
 
+function _write_optimizer_log(optimizer_log::Dict, save_path::AbstractString)
```
daniel-thom: I didn't follow this throughout the code, but I'm curious if whoever creates optimizer_log as a Dict could create it as a struct. Instead of "_write_optimizer_log" this function could be ```write_data(log::OptimizerLog, save_path)```.  Seems to be the Julian way.

PR #231, opened 2019-10-09
Write models

Lilyhanig:


daniel-thom:
That looks much cleaner! Thanks for making the changes.

PR #234, opened 2019-10-14
Jd/pm update

### LINKED ISSUE #233: :network_flow constraints incorrectly calculated in StandardPTDF network formulation
There is an error in the ptdf_networkflow function in the file ptdf_model.jl.
 
When adding the flow variable definitional constraints (roughly line 25), the expression 
` PTDF[PSY.get_name(b), :]'*canonical_model.expressions[expression].data[:, t]`
Assumes that the bus axes are in the same order.  This is not the case as can be seen by printing out the axes of both objects.

jd-lara:
This PR updates to PM v0.13, re-exports the supported networks. 
Also fixes #233 and #235.


src/devices/device_models/AC_branches.jl
```diff
@@ -117,7 +134,7 @@ function branch_rate_constraint(canonical_model::CanonicalModel,
                                 device_formulation::Type{D},
                                 system_formulation::Type{S}) where {B<:PSY.ACBranch,
                                                                     D<:AbstractBranchFormulation,
-                                                                    S<:PM.AbstractActivePowerFormulation}
+                                                                    S<:PM.AbstractDCPLLModel}
```
ccoffrin: Should this be `AbstractActivePowerModel`

jd-lara: yes, thanks for that.

src/network_models/powermodels_interface.jl
```diff
@@ -304,7 +300,7 @@ function PMvarmap(system_formulation::Type{S}) where {S<:PM.DCPlosslessForm}
     return pm_var_map
 end
 
-function PMvarmap(system_formulation::Type{S}) where {S<:PM.AbstractActivePowerFormulation}
+function PMvarmap(system_formulation::Type{S}) where {S<:PM.AbstractDCPLLModel}
```
ccoffrin: Double check this. `AbstractActivePowerModel` would be the normal update.

PR #243, opened 2019-10-18
adding aggregation functionality to the simulation

Lilyhanig:


src/routines/solve_routines.jl
```diff
@@ -130,8 +136,140 @@ function run_sim_model!(sim::Simulation; verbose::Bool = false, kwargs...)
...
+# Example
+```julia
+sim = Simulation("test", 7, stages, "/Users/lhanig/Downloads/"; 
+verbose = true, system_to_file = false)
+run_sim_model!(sim::Simulation; verbose::Bool = false, kwargs...)
+references = make_references(sim, "2019-10-03T09-18-00-test")
+```
+"""
+
+function make_references(sim::Simulation, date_run::String)
```
jd-lara: Any reason to put this in the solve routines?

src/core/core_structs/aggregation.jl
```diff
@@ -0,0 +1,195 @@
+# This function returns the length of time_stamp for each step
```
jd-lara: Why is this file in the core_structs folder? I don't see any struct here.

src/routines/solve_routines.jl
Lilyhanig: hmmm not particularly, should I just make a new file?

src/core/core_structs/results_model.jl
```diff
@@ -62,11 +65,13 @@ function load_operation_results(path::AbstractString, directory::AbstractString)
         optimizer = Dict{Symbol, Any}(eachcol(Feather.read("$file_path"),true))
 
         file_path = joinpath(folder_path,"time_stamp.feather")
-        time_stamp = Feather.read("$file_path")
+        temp_time_stamp = Feather.read("$file_path")
```
daniel-thom: This should already be a string.

src/routines/solve_routines.jl
```diff
@@ -130,8 +136,140 @@ function run_sim_model!(sim::Simulation; verbose::Bool = false, kwargs...)
             @assert stage.executions == stage.execution_count
             stage.execution_count = 0 # reset stage execution_count
         end
+        
+    end
+    if (:no_dict in keys(kwargs)) == true
```
daniel-thom: I see that you are controlling the return type of the function with a keyword arg.  I'll suggest that it might be cleaner to accept an optional dict as a parameter defaulting to nothing.  If not nothing, fill it in.  But that's more style.

src/routines/solve_routines.jl
jd-lara: What is the `add_reference` function supposed to do?

PR #245, opened 2019-10-21
Jd/sim forecast update

jd-lara:
Not perfect implementation, but it is an update for the new forecasts


src/simulation_models/stage_update.jl
```diff
@@ -4,12 +4,17 @@ function parameter_update!(param_reference::UpdateRef{T},
...
-        for (ix, val) in enumerate(param_array[device,:])
-            value = PSY.get_forecast_value(forecast, ix)
+    devices = PSY.get_components(T, sim.stages[stage_number].sys)
+    initial_forecast_time = sim.ref.date_ref[stage_number]
+
+    for d in devices
+        ts_vector = TS.values(PSY.get_data(PSY.get_forecast(PSY.Deterministic,
+                                            d,
+                                            initial_forecast_time,
+                                            "$(param_reference.access_ref)")))
```
daniel-thom: I find it more natural to write ```string(param_reference.access_ref)```, but I know you do this a lot.

src/simulation_models/stage_update.jl
```diff
@@ -4,12 +4,17 @@ function parameter_update!(param_reference::UpdateRef{T},
...
+        device_name = PSY.get_name(d)
+        for (ix, val) in enumerate(param_array[device_name,:])
+            value = ts_vector[ix]
```
daniel-thom: Where does this get multiplied by the component field (label) value?

jd-lara: I changed the implementation in PowerSimulations. We don't need to do that anymore

PR #251, opened 2019-11-01
Remove instability introduced from solver

sourabhdalvi:
Solvers don't round results which can lead to some instability in simulation as cache value is not equal to variable value in conditional statements like this`var_status =  var_value >= eps() ? 1.0 : 0.0`.
As  is `eps() = 2.220446049250313e-16` and can wrongfully set variable values to 1.0 even when solver returns `1.00e-15`


sourabhdalvi:
I would also suggest we change this function but I'm sure what you're trying to do here? is `var_value` suppose to be 0 or 1 or actual thermal output ?
```julia 
function _calculate_ic_quantity(initial_condition_key::ICKey{DevicePower, PSD},
                               ic::InitialCondition,
                               var_value::Float64,
                               cache::Union{Nothing,AbstractCache}) where PSD <: PSY.ThermalGen


    if isnothing(cache)
        status_change_to_on = value(ic) <= eps() && var_value >= eps()
        status_change_to_off = value(ic) >= eps() && var_value <= eps()
    else
        name = device_name(ic)
        time_cache = cache_value(cache, name)
        status_change_to_on = time_cache[:status] >= eps() && var_value <= eps()
        status_change_to_off = time_cache[:status] <= eps() && var_value >= eps()
    end
```

src/simulation_models/stage_update.jl
```diff
@@ -130,7 +130,7 @@ function _calculate_ic_quantity(initial_condition_key::ICKey{TimeDurationON, PSD
 
     current_counter = time_cache[:count]
     last_status = time_cache[:status]
-    var_status = var_value > eps() ? 1.0 : 0.0
+    var_status = isapprox(var_value, 0.0, atol = 1e-4) ? 1.0 : 0.0
```
sourabhdalvi: And why not use the binary status variable value here? instead of using the actual output variable ?
This is not clear to me, may be I'm missing something here.

jd-lara: Because the binary won't always be available if the FF is from a dispatch to a dispatch

PR #252, opened 2019-11-01
Sim aggregation

Lilyhanig:
- created an aggregation file with aggregation functions
- made changes recommended on previous PR
- changed run_sim_model to always return a dictionary (no key words with regards to the dictionary)
- changed the results struct to have an abstract type for scenarios where other things are included in the results
- changed the optimizer log to be written to a JSON


src/core/core_structs/results_model.jl
```diff
@@ -0,0 +1,85 @@
...
+                      duals::Dict{Symbol, Any})
+    return AggregatedResults(variables, total_cost, optimizer_log, time_stamp, duals)
+end
+function get_variable(res_model::OperationsProblemResults, key::Symbol)
+        try
+            !isnothing(res_model.variables)
+        catch
+            error("No variable with key $(key) has been found.")
+        end
+    return get(res_model.variables, key, nothing)
```
daniel-thom: I think you're missing some updates from Jose.  This is a bug that he fixed today.

src/core/core_structs/results_model.jl
```diff
@@ -0,0 +1,85 @@
...
+    optimizer = JSON.parse(open(joinpath(folder_path, "optimizer_log.json")))
+    time_stamp = Feather.read(joinpath(folder_path,"time_stamp.feather"))
+    shorten_time_stamp!(time_stamp)
+    obj_value = Dict{Symbol, Any}(:OBJECTIVE_FUNCTION => optimizer["obj_value"])
+    results = make_results(variables, obj_value, optimizer, time_stamp)
+    return results
+end
+
+function shorten_time_stamp!(time::DataFrames.DataFrame)
+    time = time[1:(size(time,1)-1),:]
```
daniel-thom: I don't think this will do what you want.  It will create a new DataFrame and return it.  The passed DataFrame should be unmodified.  You aren't catching the return at the call site above.

src/routines/get_results.jl
```diff
@@ -48,7 +48,32 @@ function _result_dataframe_variables(variable::JuMP.Containers.DenseAxisArray)
     end
 
 end
-
+function _result_dataframe_duals(constraint::JuMP.Containers.DenseAxisArray)
+    if length(axes(constraint)) == 1
+        result = Vector{Float64}(undef, length(first(constraint.axes)))
+        for t in constraint.axes[1]
+            try result[t] = JuMP.dual(constraint[t])
+            catch
```
daniel-thom: Suggest catching only the exact exception you're expecting.  Otherwise, someone might end up with a confusing problem to debug.

src/utils/aggregation.jl
```diff
@@ -1,74 +1,147 @@
...
+    end
+    ref = DataFrames.DataFrame()
+    for (ix,time) in enumerate(step_df.Date)
+        try
+            file_path =  step_df[ix, :File_Path]
+            time_file_path = joinpath(dirname(file_path), "time_stamp.feather")
+            time_stamp = DataFrames.DataFrame(Feather.read("$time_file_path"))
+            shorten_time_stamp!(time_stamp)
+            append!(ref,time_stamp)
+        catch
```
daniel-thom: You could explicitly detect the bad date range without relying on the exception.  In this case if someone makes any coding error that causes an exception, it will manifest as this bad date range.

src/utils/aggregation.jl
```diff
@@ -1,74 +1,147 @@
...
+    for n in 1:length(step)
+        step_df = vcat(step_df,date_df[date_df.Step .== step[n], :])
+    end
+    ref = DataFrames.DataFrame()
+    for (ix,time) in enumerate(step_df.Date)
+        try
+            file_path =  step_df[ix, :File_Path]
+            time_file_path = joinpath(dirname(file_path), "time_stamp.feather")
+            time_stamp = DataFrames.DataFrame(Feather.read("$time_file_path"))
+            shorten_time_stamp!(time_stamp)
```
daniel-thom: Same as above

src/utils/aggregation.jl
```diff
@@ -1,74 +1,147 @@
...
+    if size(unique(ref),2) == size(ref,2)
+    extra_time_length = 0
+    else
+    extra_time_length = size(unique(ref),1)./(length(step)+1)
+    end
+    return extra_time_length
+end
 
-for (ix,time) in enumerate(date_df.Date)
+function _reading_references(results::Dict, names::Array, stage::String, step::Array,
```
daniel-thom: names is a function in Base. You should use another word.

src/utils/aggregation.jl
```diff
@@ -1,74 +1,147 @@
...
+    return extra_time_length
 end
 # This is the count_time_overlap for if all results for a stage are desired
 
-function _count_time_overlap_all(stage::String, references::Dict{Any,Any})
-
-variable = collect(keys(references[stage]))
-date_df = references[stage][variable[1]]
-ref = DataFrames.DataFrame()
+function _count_time_overlap(stage::String, references::Dict{Any,Any})
```
daniel-thom: This looks like very similar code to the function above.

src/utils/aggregation.jl
```diff
@@ -153,43 +225,116 @@ step = ["step-1","step-2", "step-3"] # has to match the date range
...
+    optimizer = optimizer = JSON.parse(open(joinpath(file_path, "optimizer_log.json")))
+    obj_value = Dict{Symbol, Any}(:OBJECTIVE_FUNCTION => optimizer[:obj_value])
+
+    if !isepmty(dual)
+        duals = _reading_references(duals, dual, stage, references, extra_time_length)
+        results = make_results(variables, obj_value, optimizer, time_stamp, duals)
+    else
+        results = make_results(variables, obj_value, optimizer, time_stamp)
+    end
+    if (:write in keys(kwargs)) == true
```
daniel-thom:
```suggestion
    if :write in keys(kwargs)
```

src/utils/plot_results.jl
```diff
@@ -138,56 +132,28 @@ function get_bar_gen_data(res::OperationsProblemResults)
...
+```julia
+sort_data!(results)
+```
+***Note:*** only the generators included in 'my_order' will be in the
 results, and consequently, only these will be plotted. This can be a nice
 feature for variables with more than 5 generators.
 
 """
-function sort_data(res::OperationsProblemResults; kwargs...)
+function sort_data!(res::OperationsProblemResults; kwargs...)
```
daniel-thom: Why did you add the '!'?  It does look like res is being indirectly modified, but then it returns a new instance.  Semantically, when I read the function signature I would assume that this will sort res in place and not return anything.

src/utils/aggregation.jl
```diff
@@ -62,27 +62,27 @@ function _count_time_overlap(stage::String, references::Dict{Any,Any})
...
 
-    for name in (names)
+    for name in (variables)
         date_df = references[stage][dual[name]]
         step_df = DataFrames.DataFrame(Date = Dates.DateTime[], Step = String[], File_Path = String[])       
         for n in 1:length(step)
             step_df = vcat(step_df,date_df[date_df.Step .== step[n], :])
         end
-        results[(names[name])] = DataFrames.DataFrame()
+        results[(variables[name])] = DataFrames.DataFrame()
```
daniel-thom: It doesn't seem like variables[name] should work.  Do you want ```results[name]```

Project.toml
```diff
@@ -4,10 +4,14 @@ authors = ["Jose Daniel Lara", "Dheepak Krishnamurthy", "Clayton Barrows"]
 version = "0.2.0"
 
 [deps]
+Cbc = "9961bab8-2fa3-5c5a-9d89-47fab24efd76"
```
jd-lara: Why did these got added to the deps? These are only needed for testing.

src/utils/aggregation.jl
```diff
@@ -153,43 +217,116 @@ step = ["step-1","step-2", "step-3"] # has to match the date range
...
 
+# Example
+```julia
+sim = Simulation("test", 7, stages, "/Users/lhanig/Downloads/"; 
+verbose = true, system_to_file = false)
+run_sim_model!(sim::Simulation; verbose::Bool = false, kwargs...)
+references = make_references(sim, "2019-10-03T09-18-00-test")
+```
+"""
+function make_references(sim::Simulation, date_run::String; kwargs...)
```
jd-lara: @Lilyhanig this code is broken and was commented out in 988842a33535d8743b6c893f837de4ef8f450579 

See details below

src/utils/aggregation.jl
```diff
@@ -153,43 +217,116 @@ step = ["step-1","step-2", "step-3"] # has to match the date range
...
+    sim.ref.date_ref[1] = sim.daterange[1]
+    sim.ref.date_ref[2] = sim.daterange[1]
+    references = Dict()
+    for (ix, stage) in enumerate(sim.stages)
+        variables = Dict()
```
jd-lara: `Dict{Symbol, DataFrame}`

src/utils/aggregation.jl
```diff
@@ -153,43 +217,116 @@ step = ["step-1","step-2", "step-3"] # has to match the date range
...
+        interval = PSY.get_forecasts_interval(stage.sys)
+        variable_names = collect(keys(sim.stages[ix].canonical.variables))
+        if :dual_constraints in keys(kwargs) && !isnothing(get_constraints(stage.canonical))
+            dual_cons = _concat_string(kwargs[:dual_constraint])
+            variable_names = vcat(variable_names, dual_cons)
+        end
+        for name in variable_names
+            variables[variable_names[name]] = DataFrames.DataFrame(Date = Dates.DateTime[],
```
jd-lara: variable_names is an Array and name is a symbol. This produces an error because you can't index an Array with a Symbol. What is the objective here?

PR #253, opened 2019-11-03
Jd/sim aggregation

jd-lara:


jd-lara:
@Lilyhanig I am merging this branch into master as a replacement from #252. Please review to make sure that the conflict fix didn't make any bad changes.

PR #258, opened 2019-11-07
Cb/sim stage updates

claytonpbarrows:
- [x] update stage validation
 - [x] update stage construction
 - [x] update stage update forecast retrieval
 - [x] update `UpdateRef` to store the label accessors
 - [ ] update and enable simulation tests


src/core/core_structs/simulations_stages.jl
```diff
@@ -60,6 +60,9 @@ end
 mutable struct Stage <: AbstractStage
     op_problem::Type{<:AbstractOperationsProblem}
     model::OperationsTemplate
+    initial_time::Dates.DateTime
+    horizon::Int64
```
jd-lara: This information will be duplicated in canonical. I am fine with that, just warning about it.

src/simulation/stage_update.jl
```diff
@@ -6,12 +6,15 @@ function parameter_update!(param_reference::UpdateRef{T},
 
     devices = PSY.get_components(T, sim.stages[stage_number].sys)
     initial_forecast_time = sim.ref.date_ref[stage_number]
```
jd-lara: Can we add getter functions for this? I think it will be better.

claytonpbarrows:
An example is here: https://github.com/NREL-SIIP/Examples/blob/tutorials-refactor/notebook/PowerSimulations_examples/sequential_simulations.ipynb

claytonpbarrows:
This is working. @jd-lara an RTS example is here: https://github.com/NREL-SIIP/Examples/blob/tutorials-refactor/script/PowerSimulations_examples/sequential_simulations.jl

PR #259, opened 2019-11-07
fixed issue with loading the simulation results

Lilyhanig:


jd-lara:
@Lilyhanig I changed the PR base to @claytonpbarrows that is already testing the simulations with the RTS. It will be easier to review this way and make sure that the output of the simulation is what we expect.

Issue #271, opened 2019-11-27
Code base refactor 


jd-lara:
- [ ] File reorganization
- [x] Style guide consistency
- [x] Refactor DeviceModel and uses of construct_device

jd-lara:
Partially addressed in #320 with the addition of style guide checks and folder structure simplification.

Issue #219, opened 2019-09-28
re-implement common using structs


jd-lara:
@daniel-thom I think your proposal to change how the common modeling functions are implemented is valuable. We should incorporate this in v0.2.0 as it will make the code easier to read.

jd-lara:
@sourabhdalvi we will need to consider this in order to address the services modeling.

PR #275, opened 2019-11-27
Jd/services

jd-lara:
This PR is a partial implementation of the changes required to implement "constraint modification" when a feedforward or services are present. It partially addresses #219 and #271.


src/devices_models/devices/common/range_constraint.jl
```diff
@@ -1,64 +1,76 @@
...
+            expression_lb = variable[name, t]
+            con_ub[name, t] = JuMP.@constraint(psi_container.JuMPmodel, expression_ub <= limits.max)
+            con_lb[name, t] = JuMP.@constraint(psi_container.JuMPmodel, expression_lb >= limits.min)
+        end
     end
 
     return
+end
 
+struct DeviceRangeSemiContinous
```
daniel-thom: typo in struct name

PR #278, opened 2019-12-02
Create Julia 1.2 version of test project

daniel-thom:
This adds logging during tests as well as the ability to run individual tests.  It doesn't yet load the full test environment unless you call Pkg.test().  I'll add that as a separate PR.


test/runtests.jl
```diff
@@ -13,6 +14,8 @@ using Cbc
 using OSQP
 using TimeSeries
 using ParameterJuMP
+using Compat: Test
```
jd-lara: Why do we need to use compat here to import the Test package?

daniel-thom: Per the TestSetExtensions docs:  https://github.com/ssfrr/TestSetExtensions.jl

jd-lara: I don't think that's necessary, the README hasn't been updated since 2017

daniel-thom: You are correct.  Thanks.

PR #280, opened 2019-12-03
Jd/chronology

jd-lara:
This addresses some of the concerns about the user-facing side of the simulation definition. Please refer to the test file `test_simulation_sequence.jl` for an example.   

There are some tests missing for the simulation object, I will add more to the same PR. The changes are quite large so this will give time to review.


src/core/core_structs/chronology.jl
```diff
@@ -1,10 +1,18 @@
...
+        new(from_steps, to_executions)
+    end
 end
 
-struct RecedingHorizon <: AbstractChronology end
-
-struct Sequential <: AbstractChronology end
+struct RecedingHorizon <: AbstractChronology
+    step::Int64
+    function RecedingHorizon(;from_step::Int64=1)
```
daniel-thom: By specifying the type as an Int you prevent the user from passing a non-Int that could be converted to an Int, like a Float64.  You're probably better off not specifying the type on inner constructor parameters.

daniel-thom: Do these need to be Int64 rather than just Int?  Whenever I see that I'm assuming that the developer is expecting numbers bigger than Int32 can hold.

src/core/core_structs/simulation.jl
```diff
@@ -6,6 +6,7 @@ mutable struct SimulationRef
     date_ref::Dict{Int64, Dates.DateTime}
     current_time::Dates.DateTime
     reset::Bool
+    daterange::NTuple{2, Dates.DateTime} #Inital Time of the first forecast and Inital Time of the last forecast
```
daniel-thom: This might be better expressed as an alias, like Min_Max.

src/simulation/stage_update.jl
```diff
@@ -31,7 +31,7 @@ end
 
 # This makes the choice in which variable to get from the results.
 function _get_stage_variable(chron::Type{RecedingHorizon},
```
daniel-thom: not part of this review, but you aren't using `chron`.  Could just be `::Type{Consecutive}`.  And below, too.

test/test_simulation sequence.jl
```diff
@@ -0,0 +1,26 @@
...
+
+    for field in fieldnames(SimulationSequence)
+          @test !isempty(getfield(sequence, field))
+    end
+
+    sim = Simulation(name = "test",
+                 steps = 2,
+                 stages = stages_definition,
+                 stages_sequence = sequence,
+                 simulation_folder= "/Users/jdlara/Desktop/",
```
daniel-thom: Need to remove your local directory name.

src/core/core_structs/simulation_sequence.jl
```diff
@@ -0,0 +1,28 @@
+mutable struct SimulationSequence
+    initial_time::Dates.DateTime
+    horizons::Dict{String, Int64}
+    intervals::Dict{String, <:Dates.TimePeriod}
+    feed_forward_chronologies::Dict{Pair{String, String}, <:AbstractChronology}
+    feedforward::Dict{Tuple{String, Symbol}, <:AbstractAffectFeedForward}
+    ini_cond_chronology::Dict{String, <:AbstractChronology}
```
daniel-thom: nits:  feedforward vs feed_forward.  Be consistent.
If the user is going to use ini_cond_chronology then spell out the whole thing.  If internal only, never mind.

src/core/simulation.jl
```diff
@@ -160,8 +200,9 @@ function _build_simulation!(sim_ref::SimulationRef,
...
         end
     end
-    _validate_steps(stages, steps, stage_initial_times)
-    _check_chronology_ref(stages)
-    dates, validation = _get_dates(stages)
-    return dates, validation, _build_stages(sim_ref, stages, verbose = verbose; kwargs...)
+    #_validate_steps(stages, steps, stage_initial_times)
+    #_check_chronology_ref(stages)
+    #dates, validation = _get_dates(stages)
+    #_build_stages(sim_ref, stages, verbose = verbose; kwargs...)
```
daniel-thom: Why are these commented out?

jd-lara: working on it

src/core/core_structs/chronology.jl
jd-lara: I think we always use Int64, I don't think it matters too much.

src/simulation/stage_update.jl
jd-lara: Not updated yet.

src/core/chronology.jl
```diff
@@ -0,0 +1,45 @@
...
+end
+
+struct RecedingHorizon <: AbstractChronology
+    step::Int64
+    function RecedingHorizon(;from_step::Int64=1)
+        new(step)
+    end
+end
+
+function check_chronology(synch::Synchronize,
```
Lilyhanig: sometimes synchronized gets shortened to "synch" and sometimes it gets shortened to "sync". I think it should always be "sync".

test/test_chronology.jl
```diff
@@ -5,8 +5,8 @@ else
 end
 
 function test_chronology()
-    stages = Dict(1 => Stage(template_uc, 24, Dates.Hour(24), 1, c_sys5_uc, GLPK_optimizer,  Dict(0 => Sequential())),
-                2 => Stage(template_ed, 12, Dates.Minute(5), 24, c_sys5_ed, GLPK_optimizer, Dict(1 => Synchronize(24,1), 0 => Sequential()), TimeStatusChange(:ON_ThermalStandard)))
+    stages = Dict(1 => Stage(template_uc, 24, Dates.Hour(24), 1, c_sys5_uc, GLPK_optimizer,  Dict(0 => Consecutive())),
```
Lilyhanig: shouldn't this change to be the new stages formulation?

jd-lara: This isn't updated yet

PR #281, opened 2019-12-12
12 12 testing

Lilyhanig:


test/Project.toml
```diff
@@ -13,6 +13,7 @@ Memento = "f28f55f0-a522-5efc-85c2-fe41dfb9b2d9"
 OSQP = "ab2f91bb-94b4-55e3-9ba0-7f65df51de79"
 ParameterJuMP = "774612a8-9878-5177-865a-ca53ae2495f9"
 PowerModels = "c36e90e8-916a-50a6-bd94-075b64ef4655"
+PowerSimulations = "e690365d-45e2-57bb-ac84-44ba829e73c4"
```
jd-lara: There should be no changes to toml files

Lilyhanig: oops, didn't mean to add Project.toml

src/utils/simulation_results.jl
```diff
@@ -164,16 +164,25 @@ function check_file_integrity(path::String)
...
         ))
     end
 end
+
+function get_variable_names(sim::Simulation, stage::Any)
+     return collect(keys(sim.stages[stage].internal.psi_container.variables))
+ end
+ function get_reference(sim_results::SimulationResultsReference, stage::String, step::Int, variable::Symbol) 
+     file_paths = sim_results.ref["stage-$stage"][variable]
+     file_paths = filter(file_paths -> file_paths.Step == "step-$step", file_paths)[:, 3]
```
daniel-thom: What is `[:, 3]`?  The 3 is not obvious to me.

src/utils/simulation_results.jl
```diff
@@ -164,16 +164,25 @@ function check_file_integrity(path::String)
...
+     return file_paths
+ end
+
+  get_psi_container(sim::Simulation, stage::Any) = sim.stages[stage].internal.psi_container
```
daniel-thom: Would it be better to check if `stage` is present before trying to dereference it?

src/core/simulation.jl
```diff
@@ -128,18 +138,18 @@ function _get_simulation_initial_times!(sim::Simulation)
             stage_initial_times[stage_number] = PSY.generate_initial_times(stage_system,
                                                         get_interval(get_sequence(sim), stage_name),
                                                         get_horizon(get_sequence(sim), stage_name))
-            isempty(stage_initial_times[stage_number]) ? throw(ArgumentError("Simulation interval 
-                                    and forecast interval definitions are not compatible")) : nothing ; 
+            isempty(stage_initial_times[stage_number]) ? throw(IS.ConflictingInputsError("Simulation interval ($(get_interval(get_sequence(sim), stage_name))) and
```
daniel-thom: nit:  repeated calls to get_interval(get_sequence).  Also, go please ahead and remove the ternary operator.

src/utils/printing.jl
```diff
@@ -126,6 +130,27 @@ function Base.show(io::IO, ::MIME"text/plain", results::Results)
     println(io, "Stage()")
  end
 
+function Base.show(io::IO, period::Dates.Millisecond)
+    if period % Dates.Millisecond(1000) !== Dates.Millisecond(0)
```
daniel-thom: `!=`

src/utils/printing.jl
```diff
@@ -126,6 +130,27 @@ function Base.show(io::IO, ::MIME"text/plain", results::Results)
     println(io, "Stage()")
  end
 
+function Base.show(io::IO, period::Dates.Millisecond)
```
daniel-thom: I think that this should account for the size of the number and not just divisibility.  For example, I think that 45.2 seconds should be reported in seconds rather than milliseconds.

src/utils/simulation_results.jl
Lilyhanig: if stage isn't present a KeyError already gets thrown

daniel-thom: This was defined as an exported function.  Getting a KeyError on an exported function exposes underlying implementation details to the user, which you wouldn't want to do.  If the function is internal only then it doesn't really matter.

src/routines/solve_routines.jl
```diff
@@ -102,7 +102,8 @@ function execute!(sim::Simulation; verbose::Bool = false, kwargs...)
     elseif sim.internal.reset == false
         error("Re-build the simulation")
     end
-    !sim.internal.compiled_status && error("Simulation not built, build the simulation to execute")
+    isnothing(sim.internal) && error("Simulation not built, build the simulation to execute")
+    #!sim.internal.compiled_status && error("Simulation not built, build the simulation to execute")
```
daniel-thom: Does this need to be enabled?

src/utils/printing.jl
```diff
@@ -130,27 +130,7 @@ function Base.show(io::IO, ::MIME"text/plain", results::Results)
...
-    elseif period % Dates.Millisecond(604800000) !== Dates.Millisecond(0)
-        period = convert(Dates.Day, period)
-        println(io, period) # day
-    else
-        period = convert(Dates.Week, period)
-        println(io, period) # minute
-    end
-end
-
+# TODO test
```
daniel-thom: Stale?

src/core/simulation.jl
```diff
@@ -246,3 +261,12 @@ function build!(sim::Simulation; verbose::Bool = false, kwargs...)
     sim.internal.compiled_status = true
     return
 end
+
+function _check_folder(folder::String)
+    !isdir(folder) && throw(IS.ConflictingInputsError("Specified folder is not valid"))
+    testdir(folder) = try
```
daniel-thom: This creates a lambda function and returns it.  Is that what you want?

src/core/simulation.jl
```diff
@@ -228,16 +241,18 @@ function _build_stages!(sim::Simulation, verbose::Bool = true; kwargs...)
 end
 
 function build!(sim::Simulation; verbose::Bool = false, kwargs...)
+    _check_inputs(sim)
     _check_chronologies(sim)
-    raw_dir, models_dir, results_dir = _prepare_workspace(sim.name, sim.simulation_folder)
+    _check_folder(sim.simulation_folder)
+    raw_dir, models_dir, results_dir = sim.simulation_folder, sim.simulation_folder, sim.simulation_folder #_prepare_workspace(sim.name, sim.simulation_folder)
```
daniel-thom: Can you remove this line and use the `sim.` variables below?

test/test_simulation_execute.jl
```diff
@@ -1,45 +1,159 @@
...
+            time_file_path_1 = joinpath(dirname(reference_1), "time_stamp.feather") #first line, file path
+            time_file_path_2 = joinpath(dirname(reference_2), "time_stamp.feather")
+            time_1 = convert(Dates.DateTime, Feather.read(time_file_path_1)[1,1]) # first time
+            time_2 = convert(Dates.DateTime, Feather.read(time_file_path_2)[1,1])
+            time_change = time_2 - time_1
+            interval = sim.sequence.intervals[name] # change to getter function
+            @test Dates.Hour(time_change) == Dates.Hour(interval)
+        end
+    end
+#=
```
daniel-thom: Can you include comments describing why these tests are commented out?

src/core/chronology.jl
```diff
@@ -23,13 +23,13 @@ function check_chronology(sync::Synchronize,
     from_stage_sync = sync.from_steps
 
     if from_stage_sync > from_stage_horizon
-        error("The lookahead length $(from_stage_horizon) in stage is insufficient to syncronize with $(from_stage_sync) feed_forward steps")
+        throw(IS.ConflictingInputsError("The lookahead length $(from_stage_horizon) in stage is insufficient to syncronize with $(from_stage_sync) feed_forward steps"))
```
jd-lara: Nice.

src/core/simulation_sequence.jl
```diff
@@ -16,6 +16,9 @@ mutable struct SimulationSequence
                                  feed_forward = Dict{Tuple{String, Symbol, Symbol}, AbstractAffectFeedForward}(),
                                  ini_cond_chronology = Dict{String, AbstractChronology}(),
                                  cache = Dict{String, Vector{AbstractCache}}())
+        for (k, interval) in intervals
+            intervals[k] = IS.time_period_conversion(interval)
```
jd-lara: It isn't completely clear to me why the conversion is being done here. I saw the code in IS but I can't totally get why it was necessary.

src/PowerSimulations.jl
```diff
@@ -112,6 +112,9 @@ export get_all_constraint_index
 export get_all_var_index
 export get_con_index
 export get_var_index
+export get_variable_names
+export get_reference
+export get_psi_container
```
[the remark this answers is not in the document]
jd-lara: @Lilyhanig and @daniel-thom all the getter functions are meant to be internal. The general "rule" is that the exported functions are the ones required to execute an analysis. Any functions used for modeling or developing should be internal.

src/routines/get_results.jl
```diff
@@ -125,9 +125,9 @@ function get_time_stamps(op_problem::OperationsProblem)
 end
 
 function get_time_stamps(stage::Stage, start_time::Dates.DateTime)
-    interval = PSY.get_forecasts_resolution(stage.sys)
-    horizon = PSY.get_forecasts_horizon(stage.sys)
-    range = collect(start_time:interval:start_time+ interval.*horizon)
+    resolution = PSY.get_forecasts_resolution(stage.sys)
+    horizon = stage.internal.psi_container.time_steps[end]
+    range = collect(start_time:resolution:start_time + resolution*horizon)
```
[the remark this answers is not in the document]
jd-lara: my bad here.

src/routines/solve_routines.jl
```diff
@@ -93,12 +97,14 @@ results, include a vector of the variable names to be included
 """
 
 function execute!(sim::Simulation; verbose::Bool = false, kwargs...)
-    !sim.internal.compiled_status && error("Simulation not build, build the simulation to execute")
     if sim.internal.reset
         sim.internal.reset = false
     elseif sim.internal.reset == false
         error("Re-build the simulation")
     end
+    !sim.internal.compiled_status && error("Simulation not built, build the simulation to execute")
```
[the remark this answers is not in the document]
jd-lara: my bad. I think this got fixed later.

src/utils/simulation_results.jl
```diff
@@ -164,16 +164,25 @@ function check_file_integrity(path::String)
         expected_hash, file_name = split(line)
         actual_hash = compute_sha256(file_name)
         if expected_hash != actual_hash
-            @error "hash mismatch for file" file_name expected_hash actual_hash
+            @debug "hash mismatch for file" file_name expected_hash actual_hash
```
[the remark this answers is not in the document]
jd-lara: agreed with @daniel-thom

src/utils/simulation_results.jl
jd-lara: This shouldn't be an exported function.

src/core/simulation.jl
jd-lara: The behavior here also looks incorrect to me.

src/core/simulation.jl
```diff
@@ -79,7 +80,18 @@ end
 get_ini_cond_chronology(s::Simulation, number::Int64) = get(s.sequence.ini_cond_chronology, s.sequence.order[number], nothing)
 
 
-
+function _check_inputs(sim::Simulation)
```
jd-lara: This function should have a better name since it is only checking the chronology inputs to the simulation.

src/core/simulation.jl
jd-lara: _prepare_workspace isn't being called anymore, why?

src/core/simulation_sequence.jl
Lilyhanig: So, for / and + datetimes get messed up.... Julia can't automatically handle 1 hour divided by 1 minute..... we talked for a while about the fastest solution and decided that converting every period internally to milliseconds will mean that everything will work together instead of implementing a bunch of conversions or method overloads. I added the print method in IS so that when you do "sim.intervals" it still prints the time nicely in minutes/hours/days instead of milliseconds

src/core/simulation.jl
```diff
@@ -227,17 +232,30 @@ function _build_stages!(sim::Simulation, verbose::Bool = true; kwargs...)
...
     _check_chronologies(sim)
-    raw_dir, models_dir, results_dir = _prepare_workspace(sim.name, sim.simulation_folder)
-    sim.internal = SimulationInternal(raw_dir, models_dir, results_dir, sim.steps, keys(sim.sequence.order))
+    _check_folder(sim.simulation_folder)
+    sim.internal = SimulationInternal(sim.steps, keys(sim.sequence.order))
     stage_initial_times = _get_simulation_initial_times!(sim)
     for (stage_number, stage_name) in sim.sequence.order
         stage = get(sim.stages, stage_name, nothing)
-        stage_interval = sim.sequence.intervals[stage_name]
+        stage_interval = IS.time_period_conversion(sim.sequence.intervals[stage_name])
```
sourabhdalvi: This is done when Sequence is created no need to do it here

src/core/simulation_sequence.jl
sourabhdalvi: you can also do `intervals = convert(Dict{String, Dates.Millisecond}, intervals) `

test/test_simulation_build.jl
```diff
@@ -1,30 +1,185 @@
...
 
+function test_sequence_build(file_path::String)
+
+    stages_definition = Dict("UC" => Stage(GenericOpProblem, template_uc, c_sys5_uc, GLPK_optimizer),
+                                "ED" => Stage(GenericOpProblem, template_ed, c_sys5_ed, GLPK_optimizer))
+    @testset "Simulation Sequence Tests" begin
+
+    sequence = SimulationSequence(order = Dict(1 => "UC", 2 => "ED"),
+                    intra_stage_chronologies = Dict(("UC"=>"ED") => Synchronize(from_steps = 24, to_executions = 1)),
+                    horizons = Dict("UC" => 24, "ED" => 12),
```
sourabhdalvi: remove `to_executions`

test/test_simulation_execute.jl
sourabhdalvi: Please uncomment the test or update them,  even if they are failing because the functionality is not implemented

test/test_simulation_loading.jl
```diff
@@ -42,55 +58,117 @@ function test_load_simulation()
...
-            res = load_simulation_results(sim_results, ix, step, variable)
+        for name in stage_names
+            variable = PSI.get_variable_names(sim, name)
+            results = load_simulation_results(sim_results, name)
+            res = load_simulation_results(sim_results, name, step, variable)
             @test results.variables == res.variables
         end
     end
+    #=
+    @testset "Testing to verify raw output results correctly match aggregated for Receding Horizon" begin
```
sourabhdalvi: Update this test and uncomment

src/core/simulation_sequence.jl
Lilyhanig: I think the benefit of looping over the function is that so if there's ever a time where it's decided "maybe we should do seconds instead of milliseconds" or whatever, than changing the function will change it everywhere @sourabhdalvi

PR #283, opened 2019-12-15
Pass device models

claytonpbarrows:
- adds device constraint modification/creation for upward and downward reserves for:
   - thermal generation
   - renewable generation
   - storage


src/devices_models/devices/common/range_constraint.jl
```diff
@@ -49,7 +49,7 @@ function device_range(psi_container::PSIContainer,
             end
             expression_lb = JuMP.AffExpr(0.0, variable[name, t] => 1.0)
             for val in range_data.additional_terms_lb[ix]
-                JuMP.add_to_expression!(expression_lb, get_variable(psi_container, val)[name, t])
+                JuMP.add_to_expression!(expression_lb, get_variable(psi_container, val)[name, t], -1.0)
```
jd-lara: Why did this change to -1? the sign should be assigned in the creation of range_data.

src/devices_models/devices/common/range_constraint.jl
```diff
@@ -123,7 +123,7 @@ function device_semicontinuousrange(psi_container::PSIContainer,
             end
             expression_lb = JuMP.AffExpr(0.0, varcts[name, t] => 1.0)
             for val in range_data.additional_terms_lb[ix]
-                JuMP.add_to_expression!(expression_lb, get_variable(psi_container, val)[name, t])
+                JuMP.add_to_expression!(expression_lb, get_variable(psi_container, val)[name, t], -1.0)
```
jd-lara: same here.

src/devices_models/devices/common/range_constraint.jl
```diff
@@ -200,7 +200,7 @@ function reserve_device_semicontinuousrange(psi_container::PSIContainer,
...
```
jd-lara: same here.

src/devices_models/devices/common/timeseries_constraint.jl
```diff
@@ -22,20 +22,43 @@ Constructs upper bound for given variable and time series data and a multiplier.
 """
 function device_timeseries_ub(psi_container::PSIContainer,
                               ts_data::Vector{Tuple{String, Int64, Float64, Vector{Float64}}},
+                              range_data::DeviceRange, #TODO: there is some repititve info in range_data and ts_data, and consistend order/size are assumed, we should combine into a single argument
```
jd-lara: a ts_data object hasn't been implemented yet in the same way that there is a device_range object. Eventually, when we resolve #219 this will be more clear.

There shouldn't be any changes to this function yet.

src/devices_models/devices/common/timeseries_constraint.jl
```diff
@@ -163,6 +217,7 @@ Constructs lower bound for given variable using a parameter. The constraint is
 """
 function device_timeseries_param_lb(psi_container::PSIContainer,
                                     ts_data::Vector{Tuple{String, Int64, Float64, Vector{Float64}}},
+                                    range_data::DeviceRange, #TODO: there is some repititve info in range_data and ts_data, and consistend order/size are assumed, we should combine into a single argument
```
jd-lara: same answer here. This will be resolved when we implement #219

src/devices_models/devices/common/timeseries_constraint.jl
```diff
@@ -114,23 +145,46 @@ Constructs upper bound for given variable using a parameter. The constraint is
 """
 function device_timeseries_param_ub(psi_container::PSIContainer,
```
jd-lara: Same here, we need to create a ts_data object that contains everything, there shouldn't be changes here either.

test/Project.toml
```diff
@@ -13,6 +13,7 @@ Memento = "f28f55f0-a522-5efc-85c2-fe41dfb9b2d9"
 OSQP = "ab2f91bb-94b4-55e3-9ba0-7f65df51de79"
 ParameterJuMP = "774612a8-9878-5177-865a-ca53ae2495f9"
 PowerModels = "c36e90e8-916a-50a6-bd94-075b64ef4655"
+PowerSimulations = "e690365d-45e2-57bb-ac84-44ba829e73c4"
```
jd-lara: This file shouldn't be checked in

src/devices_models/devices/common/range_constraint.jl
claytonpbarrows: For downward reserves, the reserve provision needs to be subtracted from the device output and constrained to be above Pmin. However, the reserve provision value is still positive, albeit downward.

jd-lara: Makes sense.

test/Project.toml
claytonpbarrows: Why not? the test environment needs it. @daniel-thom also mentioned that it shouldn't need to be in the Project, but without it, the user needs to `dev .` in the test environment to do anything. This resolves that silliness... does it create a problem?

src/devices_models/devices/common/timeseries_constraint.jl
claytonpbarrows: Agreed on the need for #219 . As for the changes to this function, they support the desired features.  I suggest that we keep these changes and resolve the remainder of #219 in a new PR.

test/Project.toml
daniel-thom: This line should be removed.

The test environment does not need this if you run Pkg.test().  The PowerSimulations package is added implicitly.  This is only a problem if you `include("test/runtests.jl")`.  If the user wants to use that non-Julia-standardized workflow (which I happen to like) then they need to perform `]dev .`.  Doing it this way means that the test will download PowerSimulations.jl instead of using the current directory.

src/devices_models/devices/common/timeseries_constraint.jl
```diff
@@ -22,20 +22,43 @@ Constructs upper bound for given variable and time series data and a multiplier.
...
+            if add_lower_bound
+                expression_lb = JuMP.AffExpr(0.0, variable[name, t] => 1.0)
+                for val in range_data.additional_terms_lb[ix]
+                    JuMP.add_to_expression!(expression_lb, 
+                                            get_variable(psi_container, val)[name, t], -1.0)
+                end
+                con_lb[name, t] = JuMP.@constraint(psi_container.JuMPmodel,
+                                                expression_lb >= 0.0)
+            end
+        end
```
daniel-thom: There is a lot of code duplication across these functions which may cause future maintenance headaches.  Can some of the logic be consolidated?

src/devices_models/devices/renewable_generation.jl
```diff
@@ -106,18 +106,19 @@ end
 
 function activepower_constraints!(psi_container::PSIContainer,
                                 devices::IS.FlattenIteratorWrapper{R},
-                                device_formulation::Type{<:AbstractRenewableDispatchFormulation},
+                                model::DeviceModel{R, <:AbstractRenewableDispatchFormulation},
```
daniel-thom: Can this be `model::DeviceModel{R, AbstractRenewableDispatchFormulation}` instead?  Same comment for other functions that take a DeviceModel.

src/devices_models/devices/common/timeseries_constraint.jl
jd-lara: These functions need refactoring for #219. Definitely they can be simplified.

src/services_models/reserves.jl
```diff
@@ -7,11 +7,19 @@ This function add the variables for reserves to the model
...
                                   service::SR,
                                   devices::Vector{<:PSY.Device}) where SR<:PSY.Reserve
+
+    function get_ub_val(d::PSY.Device)
+        return d.tech.activepowerlimits.max
+    end
+    function get_ub_val(d::PSY.RenewableGen)
+        return d.tech.rating
+    end
+
```
daniel-thom: Should we define a PowerSystems function for this?

src/devices_models/devices/renewable_generation.jl
claytonpbarrows: removing the `<:` causes errors...

PR #285, opened 2019-12-17
Sd/sim tests

jd-lara:
Restructured Few things in Simulations 

## Chronologies
- The `_assign_chronologies`  function wasn't called the `build!`
https://github.com/NREL/PowerSimulations.jl/blob/fd17b3b2e4bbe07c9e180fc2e114152d71203678/src/core/simulation.jl#L311
- `Consecutive` can be setup to simulate like Receding Horizon by setting`interval` == `resolution`. I think we should just remove `Receding Horizon` or Receding Horizon just results in a Consecutive that has interval same as resolution.
## Execution count 
- `Synchronize` with only one field `from_periods`. The idea is `from_periods` describes the number of time periods of data to use in the next stage. Given `from_periods`, `interval` & `horizon` you can tell how many times the to_stage can be executed using  (`from_periods` - `horizon` + `interval`)/`interval`. Same can be done for the from_stage based ceil(`from_periods`/`interval`).

- If there is no `Receding Horizon` and only consecutive this method works
## Step Resolution
 - Step Resolution should be able to `interval` * `resolution` * `executions` ? This still hasn't been resolved.
## Initial Time
 - Support for contiguous forecast.
https://github.com/NREL/PowerSimulations.jl/blob/fd17b3b2e4bbe07c9e180fc2e114152d71203678/src/core/simulation.jl#L152-L157
## Parameters 
  - Parameters Container used to only have one index (device name) which would lead to wrong commitment applied in 2nd stage interval is longer than one hour
  - Suggested solution: Adding time index to Parameter Array but this introduces the complexity of figuring out which 1st/previous stage variable to use as parameter value. The current solutions is implemented in `feed_forward_update`  but it require to_stage interval to be passed. https://github.com/NREL/PowerSimulations.jl/blob/fd17b3b2e4bbe07c9e180fc2e114152d71203678/src/core/feed_forward.jl#L290-L296
 - Currently Parameter value is updated using  variable values from  the previous Stage's  JuMP model, this methods can pass the wrong parameter values if the previous stage has `Receding Horizon` inter-stage Chronology or `Consecutive` with multiple executions. In both cases the information that is wanted is no longer in the JuMP model and will need to use cache to store variable values.
 - Suggested solution: Update to `_attach_feed_forward`  which checks if the from stage has multiple executions  and creates Cache which is used for FeedForward with `add_ff_cache`. 
https://github.com/NREL/PowerSimulations.jl/blob/fd17b3b2e4bbe07c9e180fc2e114152d71203678/src/core/cache.jl#L22-L25
https://github.com/NREL/PowerSimulations.jl/blob/fd17b3b2e4bbe07c9e180fc2e114152d71203678/src/core/simulation.jl#L81-L86
 - This also creates issues with the `cache_dict` in cases where we can end-up with non-unique keys. For which we have created `CacheKey` which is similar to the `ICKey`.
https://github.com/NREL/PowerSimulations.jl/blob/fd17b3b2e4bbe07c9e180fc2e114152d71203678/src/core/aux_structs.jl#L6-L9
## Cache
- Build `TimeStatusChange` Cache now supports (in theory) using a VariableRef. 
  1. The previous default value (count =999 & value=1) can cause some really high values for initial conditions e.g. if a unit stays online for 17 consecutive hours the count = 1016 and this can cause problem. 
  2. I do understand that if we start counting from zero you could be at a state where some units can be turned off or on (but this is why you would need initial conditions). Suggested solution: build_cache method that can per-populate the cache_dict from maybe an external file ( like a csv). 
-  Update `TimeStatusChange` previously used the single parameter value to count the Up/Down hours but we need to count the status over all the time periods. The count should be stored in Milliseconds or some time units (not implemented but should be).
 - Also, Should we store the Cache/Data in the Stage its Generated in or the stage it will be used in ? There can be examples in which Cache/Data from 1st stage is used in Stage 2 & 3, in this case it might be easier to Store Cache Globally and have some mapping that will retrieve the correct Cache.
## Initial Conditions
 - Initial Conditions `cache` field now store a CacheKey . 
   1. The issue with IC is that the `UpdateRef` in the current implementation expects `P_Thermal`(Stage 2) but the Cache is build using `ON_Thermal` (Stage 1). 
   2. In this method `P_Thermal` (Stage2) is already constrained by `ON_Thermal` (Stage 1) as a parameter in the FeedForward. If all the information required to calculate the IC is in Stage 1 why do the heavy lifting in Stage 2 ?
   3. Suggested Solution: Update IC using a VariableRef and Store the Cache in Stage 1. But in a Example where Stage 1 = UC and Stage 2 = SCUC ; you then want to calculate IC from Stage1  from the `ON_Thermal` from Stage 2. So you never really rely on a continuous variable (`P_th`) that is constrained by a Parameter
    4. If we Think this is the way to do IC then we need to also change `make_initial_conditions.jl` file so it has the correct CacheKey.


src/core/aux_structs.jl
```diff
@@ -3,4 +3,11 @@ struct UpdateRef{T}
     access_ref::Union{Symbol,String}
 end
 
+struct CacheKey{C<:CacheQuantity}
+    quantity::Type{C}
+    ref::UpdateRef
+end
+
+CacheKey(::Nothing) = nothing
```
jd-lara: Why is this constructor needed?

sourabhdalvi: I deleted that

src/core/cache.jl
```diff
@@ -11,18 +14,51 @@ function TimeStatusChange(parameter::Symbol)
...
+end
 
 function build_cache!(cache::TimeStatusChange, op_problem::OperationsProblem)
     build_cache!(cache, op_problem.psi_container)
 end
 
 function build_cache!(cache::TimeStatusChange, psi_container::PSIContainer)
     parameter = get_value(psi_container, cache.ref)
-    value_array = JuMP.Containers.DenseAxisArray{Dict{Symbol, Float64}}(undef, axes(parameter)...)
+    value_array = JuMP.Containers.DenseAxisArray{Dict{Symbol, Float64}}(undef, parameter.axes[1])
```
daniel-thom: Why is this hard-coded to 1?

src/core/simulation.jl
```diff
@@ -227,20 +259,59 @@ function _build_stages!(sim::Simulation, verbose::Bool = true; kwargs...)
...
+            to_stage_res =PSY.get_forecasts_resolution(PSI.get_sys(
+                                    get_stage(sim, stage_name)))
+            from_stage_res = PSY.get_forecasts_resolution(PSI.get_sys(
+                                    get_stage(sim, key.first)))
+            to_stage_horizon = get_horizon(get_sequence(sim), stage_name)
+            to_stage_interval = get_interval(get_sequence(sim), stage_name)
+            _count = (chron.from_periods*from_stage_res - to_stage_horizon*to_stage_res + to_stage_interval) /to_stage_interval
+            if execution_count != 0.0
+                if _count != execution_count
+                    @error("Stage $stage_name has two conflicting execution counts $_count != $execution_count")
```
daniel-thom: Execution continues here.  Should an exception be thrown?  Same on line 289.

src/core/abstract_types.jl
```diff
@@ -1,5 +1,6 @@
 abstract type AbstractAffectFeedForward end
 abstract type AbstractCache end
+abstract type CacheQuantity end
```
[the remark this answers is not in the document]
sourabhdalvi: Sure

src/core/cache.jl
sourabhdalvi: Current Implementation will lead to parameter's having 2 axes but only the 1st (Device names) is required in creating this type of  Cache

src/core/simulation.jl
sourabhdalvi: Yes, I'll change this

src/core/simulation.jl
```diff
@@ -50,14 +45,13 @@ mutable struct Simulation
 
     function Simulation(;name::String,
                         steps::Int64,
-                        step_resolution::Dates.TimePeriod,
                         stages=Dict{String, Stage{AbstractOperationsProblem}}(),
                         stages_sequence=nothing,
                         simulation_folder::String,
                         verbose::Bool = false, kwargs...)
+    #step_resolution = IS.time_period_conversion(step_resolution)
```
daniel-thom: Stale?

src/core/simulation.jl
```diff
@@ -77,9 +71,41 @@ function get_simulation_time(s::Simulation, stage_number::Int64)
...
+        end
+    end
+end
 
+function add_cache!(ff::F, sim::Simulation, 
+                            stage::Stage, 
+                            device_model::DeviceModel{I, D}) where {
+                                F<:AbstractAffectFeedForward,
+                                I<:PSY.StaticInjection,
+                                D<:AbstractDeviceFormulation}
```
daniel-thom: I'd like us to follow this rule:

When a function accepts a parameter whose type is an abstract type and you don't need to use the actual concrete type inside the function, declare the parameter with the abstract type.  Prefer the use of `T` for a parameterized type instead of picking a letter in the type.

```
function add_cache!(
    ff::AbstractAffectFeedForward,
    sim::Simulation, 
    stage::Stage, 
    device_model::DeviceModel{T, AbstractDeviceFormulation}
) where T <: PSY.StaticInjection
```

test/test_simulation_execute.jl
```diff
@@ -1,45 +1,174 @@
...
+            time_file_path_1 = joinpath(dirname(reference_1), "time_stamp.feather") #first line, file path
+            time_file_path_2 = joinpath(dirname(reference_2), "time_stamp.feather")
+            time_1 = convert(Dates.DateTime, Feather.read(time_file_path_1)[1,1]) # first time
+            time_2 = convert(Dates.DateTime, Feather.read(time_file_path_2)[1,1])
+            time_change = time_2 - time_1
+            interval = sim.sequence.intervals[name] # change to getter function
+            @test Dates.Hour(time_change) == Dates.Hour(interval)
+        end
+    end
+    ### These tests are commented out until the parameter update method is updated
```
daniel-thom: No longer commented out.  And below.

test/test_simulation_loading.jl
```diff
@@ -42,55 +57,116 @@ function test_load_simulation()
...
+        for name in stage_names
+            variable = PSI.get_variable_names(sim, name)
+            results = load_simulation_results(sim_results, name)
+            res = load_simulation_results(sim_results, name, step, variable)
             @test results.variables == res.variables
         end
     end
+    
+    @testset "Testing to verify raw output results correctly match aggregated for 
+    ceding Horizon" begin
```
daniel-thom: Unexpected return?  RecedingHorizon?

src/core/feed_forward.jl
```diff
@@ -271,14 +291,22 @@ function feed_forward_update(sync::Chron,
...
+
+    execution_count = get_execution_count(to_stage) 
+    from_stage_res = PSY.get_forecasts_resolution(PSI.get_sys(from_stage))
+    to_stage_res = PSY.get_forecasts_resolution(PSI.get_sys(to_stage))
+    for device_name in axes(param_array)[1] , time in axes(param_array)[2]
+        remainder = (time*to_stage_res + (execution_count)*to_interval) % from_stage_res
+        if remainder == Dates.Millisecond(0)
+            var_count = Int((time*to_stage_res + (execution_count)*to_interval) / from_stage_res )
+        else
+            var_count = Int(ceil((time*to_stage_res + (execution_count)*to_interval) / from_stage_res))
```
jd-lara: Why are we using ceil here? We need to make sure the counts are consistent with the inputs. These hacks are what I try to avoid by having extra inputs and guaranteeing that the counts are consistent. Please avoid the use of ceil in these cases.

jd-lara:
> Restructured Few things in Simulations
> 
> ## Chronologies
> * The `_assign_chronologies`  function wasn't called the `build!`
>   https://github.com/NREL/PowerSimulations.jl/blob/fd17b3b2e4bbe07c9e180fc2e114152d71203678/src/core/simulation.jl#L311
> * `Consecutive` can be setup to simulate like Receding Horizon by setting`interval` == `resolution`. I think we should just remove `Receding Horizon` or Receding Horizon just results in a Consecutive that has interval same as resolution.

Consecutive and receding horizon have very different implications on what is happening. The receding horizon implies that the results from the lookahead are not implemented and consecutive imply that the whole horizon gets implemented. 

> 
> ## Execution count
> * `Synchronize` with only one field `from_periods`. The idea is `from_periods` describes the number of time periods of data to use in the next stage. Given `from_periods`, `interval` & `horizon` you can tell how many times the to_stage can be executed using  (`from_periods` - `horizon` + `interval`)/`interval`. The same can be done for the from_stage based ceil(`from_periods`/`interval`).

NEVER use ceil for these calculations. The counts need to be consistent if we use this hack then a simulation will run and we can't make sure that the inputs are consistent for this. What you are enabling here is inconsistency in the inputs that might allow the simulation to run with incoherent discretizations. 

> * If there is no `Receding Horizon` and only consecutive this method works

Don't remove Receding Horizon. 

> 
> ## Step Resolution
> * Step Resolution should be able to `interval` * `resolution` * `executions` ? This still hasn't been resolved.
> 

Yes, the points of indicating it is that there is clarity on what it is from the user simulation definition and also allows to have inputs checks. 

> ## Initial Time
> * Support for contiguous forecast.
>   https://github.com/NREL/PowerSimulations.jl/blob/fd17b3b2e4bbe07c9e180fc2e114152d71203678/src/core/simulation.jl#L152-L157
> 
> ## Parameters
> * Parameters Container used to only have one index (device name) which would lead to wrong commitment applied in 2nd stage interval is longer than one hour
> * Suggested solution: Adding time index to Parameter Array but this introduces the complexity of figuring out which 1st/previous stage variable to use as parameter value. The current solutions is implemented in `feed_forward_update`  but it require to_stage interval to be passed. https://github.com/NREL/PowerSimulations.jl/blob/fd17b3b2e4bbe07c9e180fc2e114152d71203678/src/core/feed_forward.jl#L290-L296

This shouldn't be implemented this way. Please roll back this change and implement a separate chronology to do this where the parameters created have the number of time indexes that want to be synchronized. This default behavior will create a lot of unnecessary parameters. Moreover, these changes + the ceil usage can lead to inadequate synchronization of the discretizations. 

The desired behavour here is the cases there is a "time decomposition" approach to solve a simulation which is different from synchronize. 

> * Currently Parameter value is updated using  variable values from  the previous Stage's  JuMP model, these methods can pass the wrong parameter values if the previous stage has `Receding Horizon` inter-stage Chronology or `Consecutive` with multiple executions. In both cases, the information that is wanted is no longer in the JuMP model and will need to use cache to store variable values.

Please provide an example of this behviour. From the description it looks to me like an improper simulation set-up. 

> * Suggested solution: Update to `_attach_feed_forward`  which checks if the from stage has multiple executions and creates Cache which is used for FeedForward with `add_ff_cache`.
>   https://github.com/NREL/PowerSimulations.jl/blob/fd17b3b2e4bbe07c9e180fc2e114152d71203678/src/core/cache.jl#L22-L25
>   
>   https://github.com/NREL/PowerSimulations.jl/blob/fd17b3b2e4bbe07c9e180fc2e114152d71203678/src/core/simulation.jl#L81-L86
> * This also creates issues with the `cache_dict` in cases where we can end-up with non-unique keys. For which we have created `CacheKey` which is similar to the `ICKey`.
>   https://github.com/NREL/PowerSimulations.jl/blob/fd17b3b2e4bbe07c9e180fc2e114152d71203678/src/core/aux_structs.jl#L6-L9
> 
> ## Cache
> * Build `TimeStatusChange` Cache now supports (in theory) using a VariableRef.
>   
>   1. The previous default value (count =999 & value=1) can cause some really high values for initial conditions e.g. if a unit stays online for 17 consecutive hours the count = 1016 and this can cause problem.

Yes, magic numbers are a problem. See the next comment. 

>   2. I do understand that if we start counting from zero you could be at a state where some units can be turned off or on (but this is why you would need initial conditions). Suggested solution: build_cache method that can per-populate the cache_dict from maybe an external file ( like a csv).

The cache here needs to be consistent with the initial conditions. The problem is that the current implementation doesn't populate the cache based on the initial conditions. 

> * Update `TimeStatusChange` previously used the single parameter value to count the Up/Down hours but we need to count the status over all the time periods. The count should be stored in Milliseconds or some time units (not implemented but should be).

The underlying optimization problem doesn't use explicit time. It needs to be done in counts of the resolution. Since the caches are meant to be used by the optimization problem, they need to be in the same units as the problem. There is no way to implement this with time units. 

TimeStatusChange stores two values. 

> * Also, Should we store the Cache/Data in the Stage its Generated in or the stage it will be used in ? There can be examples in which Cache/Data from 1st stage is used in Stage 2 & 3, in this case it might be easier to Store Cache Globally and have some mapping that will retrieve the correct Cache.

The caches are all stored in the simulation object globally. If a cache from a specific stage is needed this just requires re-writing the function to take the simulation object here:

https://github.com/NREL/PowerSimulations.jl/blob/d3c54b4ac2799629e522091c88982e47779a1c02/src/routines/stage_update.jl#L80

> 
> ## Initial Conditions
> * Initial Conditions `cache` field now store a CacheKey .
>   
>   1. The issue with IC is that the `UpdateRef` in the current implementation expects `P_Thermal`(Stage 2) but the Cache is build using `ON_Thermal` (Stage 1).
>   2. In this method `P_Thermal` (Stage2) is already constrained by `ON_Thermal` (Stage 1) as a parameter in the FeedForward. If all the information required to calculate the IC is in Stage 1 why do the heavy lifting in Stage 2 ?

The IC condition of step 2 of stage 1 requires information from the last stage in step 1 to calculate the initial condition. For instance, the initial ramping condition of the UC in step 2 needs the ED solution from step 1. 

>   3. Suggested Solution: Update IC using a VariableRef and Store the Cache in Stage 1. But in a Example where Stage 1 = UC and Stage 2 = SCUC ; you then want to calculate IC from Stage1  from the `ON_Thermal` from Stage 2. So you never really rely on a continuous variable (`P_th`) that is constrained by a Parameter

The reason for keeping the cache in stage 2 is because this prevents relying on the solution of stage 1 to update the values. For instance, if I just use 24 steps out of the 48 step lookahead from stage 1. Also, if we have a Hour Ahead UC in between the DA-UC and the ED, keeping the cache in the ED stage is simpler. 

>   4. If we Think this is the way to do IC then we need to also change `make_initial_conditions.jl` file so it has the correct CacheKey.

I still don't see the need to the cache key.

src/core/simulation.jl
jd-lara: agreed with Dan with the relaxation that the letter could be other things.

src/core/simulation.jl
jd-lara: IS implemented methods to perform proper time period conversions, we should use those.

src/core/cache.jl
jd-lara: This needs to be reverted to the previous code.

src/core/abstract_types.jl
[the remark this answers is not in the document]
jd-lara: So we have Cache and CacheType? What's the difference?

src/core/simulation.jl
```diff
@@ -77,9 +71,41 @@ function get_simulation_time(s::Simulation, stage_number::Int64)
...
+        horizon = sim.sequence.horizons[key]
+        interval = sim.sequence.intervals[key]
+        horizon_time = resolution * horizon
+        if horizon_time < interval
+            throw(IS.ConflictingInputsError("horizon ($horizon_time) is shorter than interval ($interval) for $key"))
+        end
+    end
+end
 
+function add_cache!(ff::F, sim::Simulation,
```
jd-lara: don't use ff (I might have done this too)

src/core/simulation.jl
```diff
@@ -89,17 +115,24 @@ function _check_chronologies(sim::Simulation)
 end
 
 function _assign_chronologies(sim::Simulation)
+    function find_val(d, value)
```
jd-lara: More this function to utilities? @daniel-thom what do you think about this syntax?

src/core/simulation_stages.jl
```diff
@@ -4,13 +4,13 @@ mutable struct StageInternal
     executions::Int64
     execution_count::Int64
     psi_container::Union{Nothing, PSIContainer}
-    cache_dict::Dict{Type{<:AbstractCache}, AbstractCache}
+    cache_dict::Dict{CacheKey, <:AbstractCache}
```
jd-lara: Are you storing both the cachekey and the cache in the stage?

src/core/simulation_stages.jl
```diff
@@ -67,14 +70,20 @@ function get_stage_variable(::Type{Consecutive},
     return JuMP.value(variable[device_name, step])
 end
 
-function get_stage_variable(::Type{Synchronize},
+function get_stage_variable(chron::Type{Synchronize},
```
jd-lara: Where are you using chron here? substitute for `::Type{Synchronize}`

src/devices_models/devices/thermal_generation.jl
```diff
@@ -475,3 +475,13 @@ function cost_function(psi_container::PSIContainer,
     add_to_cost(psi_container, devices, Symbol("ON_$(T)"), :fixed)
     return
 end
+
+####################### Feed Forward Cache Automation ###############################################
+
+function add_cache!(ff::SemiContinuousFF, sim::Simulation,
```
jd-lara: Why do you add this everywhere?

src/core/simulation.jl
daniel-thom: I'm fine with how this is.  It would be better as a utility function, such as `find_key_with_value`

sourabhdalvi:
Update on the PR
## Chronologies
- Added a new Chronology `SynchronizeTimeBlocks` to allow for each time period's parameter update. With `Synchronize` the functionality is the same as before where all the time period's have the same parameter value.
## Parameters
- For all feed forward types parameters are created for each time period for now. To reduce the number of parameter used, we can figure out the number of constraints that are always going to have the same parameter value to reduce the total number of parameters used. 
- All `ceil` function have been removed, to update the parameter values a variable map is now created in `build!` and stored in the `stage.internal`.
- This mapping points to the 1st stage variable whose values should be updated in the parameter for the next simulation
## Cache
- Still not clear to me how Receding Horizon in the 1st stage would work without cache

jd-lara:
> Update on the PR
> 
> ## Chronologies
> * Added a new Chronology `SynchronizeTimeBlocks` to allow for each time period's parameter update. With `Synchronize` the functionality is the same as before where all the time period's have the same parameter value.

ok

> 
> ## Parameters
> * For all feed forward types parameters are created for each time period for now. To reduce the number of parameter used, we can figure out the number of constraints that are always going to have the same parameter value to reduce the total number of parameters used.

This needs to be reverted in this PR to the original design. It is inadvisable to leave this fix for a future solution that isn't implemented. 

> * All `ceil` function have been removed, to update the parameter values a variable map is now created in `build!` and stored in the `stage.internal`.

ok

> * This mapping points to the 1st stage variable whose values should be updated in the parameter for the next simulation
> 
> ## Cache
> * Still not clear to me how Receding Horizon in the 1st stage would work without cache

I will do this once I am back.

PR #286, opened 2019-12-20
Device timeseries

claytonpbarrows:
This PR is a draft to simplify code in [timeseries_constraint.jl](https://github.com/NREL/PowerSimulations.jl/blob/dev0-2/src/devices_models/devices/common/timeseries_constraint.jl). It implements a new container `DeviceTimeSeries` and collapses the construction of time series data and constraint data into a single loop.

TODO:
 - propagate changes to electric_loads.jl

### DIFF
--- a/src/devices_models/devices/common/constraints_structs.jl
+++ b/src/devices_models/devices/common/constraints_structs.jl
@@ -1,31 +1,12 @@
 """ Data Container to construct range constraints"""
 struct DeviceRange
-    names::Vector{String}
-    values::Vector{MinMax}
-    additional_terms_ub::Vector{Vector{Symbol}}
-    additional_terms_lb::Vector{Vector{Symbol}}
-end
-
-function DeviceRange(count::Int64)
-    names = Vector{String}(undef, count)
-    limit_values = Vector{MinMax}(undef, count)
-    additional_terms_ub = fill(Vector{Symbol}(), count)
-    additional_terms_lb = fill(Vector{Symbol}(), count)
-    return DeviceRange(names, limit_values, additional_terms_ub, additional_terms_lb)
+    limits::MinMax
+    additional_terms_ub::Vector{Symbol}
+    additional_terms_lb::Vector{Symbol}
 end
 
 struct DeviceTimeSeries
-    names::Vector{String}
-    bus_numbers::Vector{Int64}
-    multipliers::Vector{Float64}
-    ts_vectors::Vector{Vector{Float64}}
+    bus_number::Int64
+    multiplier::Float64
+    timeseries::Vector{Float64}
 end
-
-function DeviceTimeSeries(count::Int64)
-    names = Vector{String}(undef, count)
-    bus_numbers = Vector{Int64}(undef, count)
-    multipliers = Vector{Float64}(undef, count)
-    ts_vectors = Vector{Vector{Float64}}(undef, count)
-    return DeviceTimeSeries(names, bus_numbers, multipliers, ts_vectors)
-end
-    
--- a/src/devices_models/devices/common/constraints_structs.jl
+++ b/src/devices_models/devices/common/constraints_structs.jl
@@ -17,7 +17,7 @@ struct DeviceTimeSeries
     timeseries::Vector{Float64}
     range::Union{Nothing, DeviceRange}
     function DeviceTimeSeries(name, bus_number, multiplier, timeseries, range)
-        !isnothing(range) && @assert name == range.name
+        @assert isnothing(range) || name == range.name
         return new(name, bus_number, multiplier, timeseries, range)
     end
 end
--- a/src/devices_models/devices/common/constraints_structs.jl
+++ b/src/devices_models/devices/common/constraints_structs.jl
@@ -15,7 +15,7 @@ struct DeviceTimeSeries
     bus_number::Int64
     multiplier::Float64
     timeseries::Vector{Float64}
-    range::Union{Nothing,DeviceRange}
+    range::Union{Nothing, DeviceRange}
     function DeviceTimeSeries(name, bus_number, multiplier, timeseries, range)
         !isnothing(range) && @assert name == range.name
         return new(name, bus_number, multiplier, timeseries, range)
--- a/src/devices_models/devices/common/constraints_structs.jl
+++ b/src/devices_models/devices/common/constraints_structs.jl
@@ -13,3 +13,19 @@ function DeviceRange(count::Int64)
     additional_terms_lb = fill(Vector{Symbol}(), count)
     return DeviceRange(names, limit_values, additional_terms_ub, additional_terms_lb)
 end
+
+struct DeviceTimeSeries
+    names::Vector{String}
+    bus_numbers::Vector{Int64}
+    multipliers::Vector{Float64}
+    ts_vectors::Vector{Vector{Float64}}
+end
+
+function DeviceTimeSeries(count::Int64)
+    names = Vector{String}(undef, count)
+    bus_numbers = Vector{Int64}(undef, count)
+    multipliers = Vector{Float64}(undef, count)
+    ts_vectors = Vector{Vector{Float64}}(undef, count)
+    return DeviceTimeSeries(names, bus_numbers, multipliers, ts_vectors)
+end
+    
--- a/src/devices_models/devices/common/constraints_structs.jl
+++ b/src/devices_models/devices/common/constraints_structs.jl
@@ -16,4 +16,8 @@ struct DeviceTimeSeries
     multiplier::Float64
     timeseries::Vector{Float64}
     range::Union{Nothing,DeviceRange}
+    function DeviceTimeSeries(name, bus_number, multiplier, timeseries, range)
+        !isnothing(range) && @assert name == range.name
+        return new(name, bus_number, multiplier, timeseries, range)
+    end
 end
--- a/src/devices_models/devices/common/constraints_structs.jl
+++ b/src/devices_models/devices/common/constraints_structs.jl
@@ -1,12 +1,19 @@
 """ Data Container to construct range constraints"""
 struct DeviceRange
+    name::String
     limits::MinMax
     additional_terms_ub::Vector{Symbol}
     additional_terms_lb::Vector{Symbol}
 end
 
+function DeviceRange(name::String, limits::MinMax)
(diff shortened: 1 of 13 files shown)


src/devices_models/devices/common/constraints_structs.jl
```diff
@@ -13,3 +13,19 @@ function DeviceRange(count::Int64)
     additional_terms_lb = fill(Vector{Symbol}(), count)
     return DeviceRange(names, limit_values, additional_terms_ub, additional_terms_lb)
 end
+
+struct DeviceTimeSeries
+    names::Vector{String}
+    bus_numbers::Vector{Int64}
+    multipliers::Vector{Float64}
+    ts_vectors::Vector{Vector{Float64}}
+end
```
daniel-thom: Could it be this instead?

```
struct DeviceTimeSeries
    names::String
    bus_numbers::Int64
    multipliers::Float64
    ts_vectors::Vector{Float64}
end
```

Then you could have vectors of these structs.

src/devices_models/devices/common/timeseries_constraint.jl
```diff
@@ -1,57 +1,55 @@
 @doc raw"""
     device_timeseries_ub(psi_container::PSIContainer,
-                     ts_data::Vector{Tuple{String, Int64, Float64, Vector{Float64}}}
+                     ts_data::Dict{String,DeviceTimeSeries},
```
jd-lara: The struct `DeviceTimeSeries` should contain the information in range data and the function should just take the struct with the required information. There is no need to have both structs as inputs.

src/devices_models/devices/common/range_constraint.jl
```diff
@@ -106,27 +106,27 @@ function device_semicontinuousrange(psi_container::PSIContainer,
     varbin = get_variable(psi_container, binvar_name)
     ub_name = _middle_rename(cons_name, "_", "ub")
     lb_name = _middle_rename(cons_name, "_", "lb")
+    names = collect(keys(range_data))
```
jd-lara: This call is not necessary if we keep the names vector. Any reason to remove the names vector and make this a dict?

src/devices_models/devices/common/range_constraint.jl
```diff
@@ -135,7 +135,7 @@ end
 
 @doc raw"""
     reserve_device_semicontinuousrange(psi_container::PSIContainer,
-                                    range_data::DeviceRange,
+                                    range_data::Dict{String,DeviceRange},
```
jd-lara: missing space after the comma.

src/devices_models/devices/common/range_constraint.jl
```diff
@@ -165,14 +165,14 @@ where limits in range_data.
...
-* range_data::DeviceRange : contains names and vector of min/max
+* range_data::Dict{String,DeviceRange} : contains names and vector of min/max
 * cons_name::Symbol : name of the constraint
 * var_name::Symbol : the name of the continuous variable
 * binvar_name::Symbol : the name of the binary variable
 """
 #This function looks suspicious and repetittive. Needs verification
 function reserve_device_semicontinuousrange(psi_container::PSIContainer,
-                                            range_data::DeviceRange,
+                                            range_data::Dict{String,DeviceRange},
```
jd-lara: space after comma

src/devices_models/devices/common/constraints_structs.jl
```diff
@@ -1,15 +1,12 @@
...
+    additional_terms_lb::Vector{Symbol}
 end
 
-function DeviceRange(count::Int64)
-    names = Vector{String}(undef, count)
-    limit_values = Vector{MinMax}(undef, count)
-    additional_terms_ub = fill(Vector{Symbol}(), count)
-    additional_terms_lb = fill(Vector{Symbol}(), count)
-    return DeviceRange(names, limit_values, additional_terms_ub, additional_terms_lb)
+struct DeviceTimeSeries
```
jd-lara: If you keep it as a struct per device, you only need to pass a single struct, I suggest making a single struct as the input. I am not clear why the bus number is needed. 

```suggestion
struct DeviceTimeSeries
name::String
limits::MinMax
additional_terms_ub::Vector{Symbol}
additional_terms_lb::Vector{Symbol}
multiplier::Float64
timeseries::Vector{Float64}
end
```
Even if there is a little repetition, it is far cleaner to define a single struct to the constraints that contains the required information.

src/devices_models/devices/common/timeseries_constraint.jl
```diff
@@ -64,8 +63,8 @@ end
 
 @doc raw"""
     device_timeseries_lb(psi_container::PSIContainer,
-                     ts_data::Dict{String,DeviceTimeSeries},
-                     range_data::Dict{String,DeviceRange},
+                     ts_data::Dict{String, DeviceTimeSeries},
+                     range_data::Dict{String, DeviceRange},
```
jd-lara: don't need range data here

src/devices_models/devices/common/timeseries_constraint.jl
```diff
@@ -266,15 +264,15 @@ where (name, data) in range_data.
...
+* range_data::Dict{String, DeviceRange} : container of device range constraint modification data
 * cons_name::Symbol : name of the constraint
 * var_name::Symbol :  name of the variable
 * binvar_name::Symbol : name of binary variable
 """
 function device_timeseries_ub_bin(psi_container::PSIContainer,
-                                    ts_data::Dict{String,DeviceTimeSeries},
-                                    range_data::Dict{String,DeviceRange},
+                                    ts_data::Dict{String, DeviceTimeSeries},
+                                    range_data::Dict{String, DeviceRange},
```
jd-lara: also here, no need to pass range_data

src/devices_models/devices/common/constraints_structs.jl
```diff
@@ -1,12 +1,19 @@
...
     additional_terms_ub::Vector{Symbol}
     additional_terms_lb::Vector{Symbol}
 end
 
+function DeviceRange(name::String, limits::MinMax)
+    return DeviceRange(name, limits, Vector{Symbol}(), Vector{Symbol}())
+end
+
 struct DeviceTimeSeries
+    name::String
```
jd-lara: Looks good to me but I slightly prefer having a single struct. @daniel-thom this will store the name twice, is that ok with you?

src/devices_models/devices/common/constraints_structs.jl
```diff
@@ -1,12 +1,19 @@
...
     bus_number::Int64
     multiplier::Float64
     timeseries::Vector{Float64}
+    range::Union{Nothing,DeviceRange}
```
jd-lara: This is where the name would be stored twice

src/devices_models/devices/common/range_constraint.jl
```diff
@@ -183,27 +191,31 @@ function reserve_device_semicontinuousrange(psi_container::PSIContainer,
...
     lb_name = _middle_rename(cons_name, "_", "lb")
-    names = collect(keys(range_data))
+    names = (d.name for d in range_data)
     #MOI has a semicontinous set, but after some tests is not clear most MILP solvers support it.
     #In the future this can be updated
     con_ub = add_cons_container!(psi_container, ub_name, names, time_steps)
     con_lb = add_cons_container!(psi_container, lb_name, names, time_steps)
 
-    for (name, data) in range_data
+    for data in range_data
```
jd-lara: If the character count gets too long we can consolidate the for loops as 
```for t in time, data in range_data ``` and avoid an extra indentation.

File: src/devices_models/devices/common/timeseries_constraint.jl [hunk 12]
```diff
@@ -136,48 +135,47 @@ Constructs upper bound for given variable using a parameter. The constraint is
 
 # Arguments
 * psi_container::PSIContainer : the psi_container model built in PowerSimulations
-* ts_data::Dict{String,DeviceTimeSeries} : container of device time series data and scaling factors
-* range_data::Dict{String,DeviceRange} : container of device range constraint modification data
+* ts_data::Dict{String, DeviceTimeSeries} : container of device time series data and scaling factors
+* range_data::Dict{String, DeviceRange} : container of device range constraint modification data
```
jd-lara: update docstring

src/devices_models/devices/common/timeseries_constraint.jl
```diff
@@ -188,8 +186,8 @@ end
 
 @doc raw"""
     device_timeseries_param_lb(psi_container::PSIContainer,
-                                    ts_data::Dict{String,DeviceTimeSeries},
-                                    range_data::Dict{String,DeviceRange},
+                                    ts_data::Dict{String, DeviceTimeSeries},
```
jd-lara: update docsting

src/devices_models/devices/common/constraints_structs.jl
daniel-thom: If DeviceRange will be used independently of DeviceTimeSeries then I don't have a problem with name being stored twice.  I would suggest putting an `assert name == range.name` in the DeviceTimeSeries constructor.

jd-lara: ok. 

P.S. @claytonpbarrows you are missing a space after comma here too :P

claytonpbarrows: I added the assertion statement in an inner constructor. @daniel-thom is that the best way to do it?
And to answer your question directly... yes, `DeviceRange` is used independently of `DeviceTimeSeries`

claytonpbarrows: @daniel-thom I kept the assert inside to keep the logic correct... if the range isn't nothing, make sure that the names match...

src/devices_models/devices/common/constraints_structs.jl
```diff
@@ -1,15 +1,23 @@
...
+end
+
+struct DeviceTimeSeries
+    name::String
+    bus_number::Int64
+    multiplier::Float64
+    timeseries::Vector{Float64}
+    range::Union{Nothing, DeviceRange}
+    function DeviceTimeSeries(name, bus_number, multiplier, timeseries, range)
+        !isnothing(range) && @assert name == range.name
```
daniel-thom:
```suggestion
        @assert isnothing(range) || name == range.name
```

daniel-thom: But I'm fine with the way it is.

PR #289, opened 2020-01-02
Hydro update

sourabhdalvi:
-  Updating hydro constructors
###  Adding a new implementation of  Budget / energy limit constraint
-  the constraint now enforces that the average of all the power dispatch variables has to be less than dispatch variable passed.
https://github.com/NREL/PowerSimulations.jl/blob/57a339ba75701919e5cf7ae7340ff153f575fa66/src/devices_models/devices/hydro_generation.jl#L365-L366
 - Current limitation:
  - It won't work if few of the variables belong to a different hour or have different limits, currently all variables in the problem are subject to the same limit.
  -  Only work if the stage has a horizon of a Hour or Day ( in which case the limit can be interpreted as a daily or hourly energy limit. Not sure if there is a clean solution for this
  -  Does not carry forward overage or leftover energy/water from one simulation to the next. Suggested solution : adding initial conditions  for energy and slack variable to the constraint
### Adding a new Feed Forward `EnergyLimitFF`
- It's similar to `UpperBoundFF` but a slightly different in implementation as its upper bound on the sum of variable, not sure if its enough justification to create a new type.
https://github.com/NREL/PowerSimulations.jl/blob/57a339ba75701919e5cf7ae7340ff153f575fa66/src/core/feed_forward.jl#L26-L30


jd-lara:
> * Updating hydro constructors
> 
> ### Adding a new implementation of Budget / energy limit constraint
> * the constraint now enforces that the average of all the power dispatch variables has to be less than dispatch variable passed.
>   https://github.com/NREL/PowerSimulations.jl/blob/57a339ba75701919e5cf7ae7340ff153f575fa66/src/devices_models/devices/hydro_generation.jl#L365-L366
> * Current limitation:
> * It wont work if few of the varibales below to a different hour or have different limits, currently all variables in the problem are subject to this limit.
Can you clarify? below what? 

> * Only work if the stage has a horizon of a Hour or Day ( in which case the limit can be interpreted as a daily or hourly energy limit. Not sure if there is a clean solution for this

Which stage? Why? 

> * Does not carry forward overage or leftover energy/water from one simulation to the next. Suggested solution : adding initial conditions  for energy and slack variable to the constraint

You can implement a cache for this. 

> 
> ### Adding a new Feed Forward `EnergyLimitFF`
> * It's similar to `UpperBoundFF` but a slightly different in implementation as its upper bound on the sum of variable, not sure if its enough justification to create a new type.
>   https://github.com/NREL/PowerSimulations.jl/blob/57a339ba75701919e5cf7ae7340ff153f575fa66/src/core/feed_forward.jl#L26-L30

src/devices_models/devices/hydro_generation.jl
```diff
@@ -315,101 +315,55 @@ function cost_function(psi_container::PSIContainer,
...
-    param = _add_param_container!(psi_container, param_reference, names, no_of_budgets)
-
-    for data in budget_data, i in 1:no_of_budgets
-        name = data[1]
-        forecast = data[4][i]
-        multiplier = data[3]
-        param[name] = PJ.add_parameter(psi_container.JuMPmodel, forecast)
-        constraint[name] = JuMP.@constraint(psi_container.JuMPmodel,
-                    sum([variable[name, t] for t in time_chunks[:, i]]) <= multiplier*param[name])
+function energylimit_constraints!(psi_container::PSIContainer,
```
jd-lara: This function should be implemented regardless of the feedforward, why are you passing the feedforward here?

src/devices_models/devices/hydro_generation.jl
```diff
@@ -315,101 +315,55 @@ function cost_function(psi_container::PSIContainer,
...
+        end
     end
 
     return
 end
 
-
-function device_budget_ub(psi_container::PSIContainer,
-                            budget_data::Vector{Tuple{String, Int64, Float64, Vector{Float64}}},
+function device_energylimit(psi_container::PSIContainer,
```
jd-lara: Same here, why are you passing a parameter ref here? you need to detect if the model uses parameters or not like all the other model implementations and then implement the constraint appropriately

sourabhdalvi:
> > * Current limitation:
> > * It wont work if few of the variables belong to a different hour or have different limits, currently all variables in the problem are subject to this limit.
> 
> Can you clarify? below what?

I meant if the variables belong to two different hour or there is look-ahead such that you need to enforce two different limits this wont do it. 
> 
> > * Only work if the stage has a horizon of a Hour or Day ( in which case the limit can be interpreted as a daily or hourly energy limit. Not sure if there is a clean solution for this
> 
> Which stage? Why?

The to_stage to which your feeding forward the energy limit
> > * Does not carry forward overage or leftover energy/water from one simulation to the next. Suggested solution : adding initial conditions  for energy and slack variable to the constraint
> 
> You can implement a cache for this.

Agree

src/devices_models/devices/hydro_generation.jl
sourabhdalvi: - It's a feed-forward constraint no ? At least for this use case where you decide in the 1st stage what the optimal use of water/energy is and then feed forward the this limit to the 2nd stage.
- I dont know the reason you pass the feedforward model in all the constraint func but I assumed that your intention was to do something like this
- The constraint that should be the default is tracking water/energy in the reservoir 
    - 
- If you think the energy limit needs to be a separate constraint regardless of feed-forward we can do it.

PR #290, opened 2020-01-03
refactoring service construction

claytonpbarrows:
- re-orders service construction for consistency with other variable/constraint creation.


src/services_models/reserves.jl
```diff
@@ -5,57 +5,75 @@ struct RangeReserve <: AbstractReservesFormulation end
 This function add the variables for reserves to the model
 """
 function activeservice_variables!(psi_container::PSIContainer,
-                                  service::SR,
-                                  devices::Vector{<:PSY.Device}) where SR<:PSY.Reserve
+                                  services::IS.FlattenIteratorWrapper{SR},
+                                  service_devices::Dict{NamedTuple{(:type, :name),Tuple{DataType,String}},PSY.ServiceContributingDevices}) where SR<:PSY.Reserve
```
jd-lara: Why are you creating the variables for all the services here?

src/services_models/services_constructor.jl
```diff
@@ -10,31 +10,33 @@ function construct_services!(psi_container::PSIContainer,
...
-        contributing_devices = _filter_service_mapping(service_model.service_type, services_mapping)
-        for value in values(contributing_devices)
-            add_service!(psi_container,
-                         value.service,
-                         service_model,
-                         value.contributing_devices,
-                         devices_template; kwargs...)
-        end
+        service_devices = _filter_service_mapping(service_model.service_type, services_mapping)
+        construct_service!(psi_container,
```
jd-lara: we can change the name here if desired.

src/services_models/reserves.jl
claytonpbarrows: I'm only creating the variables for all services of a particular type. Mostly to be consistent with all of the device construction.

claytonpbarrows:
This is required because the previous implementation created a parameter container each time a service was created. However, only one parameter container for each type of service was stored. So, in the case where there are multiple services of a different type, the parameter update errored out because it was trying to index into the parameter container using a reserve name that didn't exist (because it had been overwritten by the creation of another reserve of the same type). This also simplifies the handling of parameters, making it consistent with the treatment of other devices and removing one of the `include_parameters` methods.

jd-lara:
Do you have a test for the aforementioned case? I understand the problem but would like to see if there is an alternative way to address it.

jd-lara:
@claytonpbarrows I rebased to make the tests run and see the codecov report.

jd-lara:
@claytonpbarrows the issue you describe is in the way UpdateRef was originally conceptualized. The problem is the use of ParameterRef as the key. I made a PR with a quick fix in #292, however, a final fix should be addressed in PSIMA-1 (cc. @daniel-thom)

jd-lara:
@claytonpbarrows can we close this? I would like to free up some CI VM's

src/core/definitions.jl
```diff
@@ -25,3 +25,6 @@ const JuMPAffineExpressionVector = Vector{GAE{V}} where V<:JuMP.AbstractVariable
 const JuMPConstraintArray = JuMP.Containers.DenseAxisArray{JuMP.ConstraintRef}
 const JuMPParamArray = JuMP.Containers.DenseAxisArray{PJ.ParameterRef}
 const DSDA = Dict{Symbol, JuMP.Containers.DenseAxisArray}
+
+# Tolerance of comparisons
+const EqTolerance = 1.0e-10
```
jd-lara: Rename to `ComparisonTolerance` maybe?

src/services_models/reserves.jl
```diff
@@ -6,56 +6,59 @@ This function add the variables for reserves to the model
...
+                                                   ts_vector[t])
+            constraint[name, t] = JuMP.@constraint(psi_container.JuMPmodel,
+                                        sum(reserve_variable[:,t]) >= param[name, t])
         end
     else
         for t in time_steps
-            constraint[t] = JuMP.@constraint(psi_container.JuMPmodel,
-                                    sum(reserve_variable[:,t]) >= ts_vector[t]*requirement)
+            constraint[name, t] = JuMP.@constraint(psi_container.JuMPmodel,
+                                    sum(reserve_variable[:,t]) >= ts_vector[t])
```
jd-lara: Here this change will remove the assumption that the time series is 1-0 as scaling factors. This would be inconsistent with other implementations.

src/services_models/services_constructor.jl
```diff
@@ -10,31 +10,42 @@ function construct_services!(psi_container::PSIContainer,
...
+                            service_model,
+                            devices_template; kwargs...)
     end
     return
 end
 
-function add_service!(psi_container::PSIContainer, service::SR,
+function construct_service!(psi_container::PSIContainer, 
+                      services::IS.FlattenIteratorWrapper{SR},
+                      service_devices::Dict{NamedTuple{(:type, :name),Tuple{DataType,String}},PSY.ServiceContributingDevices},
```
jd-lara: We need to pass here the contributing devices and perform the parsing inside. This object is not easy to track.

src/services_models/services_constructor.jl
```diff
@@ -10,31 +10,42 @@ function construct_services!(psi_container::PSIContainer,
...
     for service_model in values(services_template)
-        contributing_devices = _filter_service_mapping(service_model.service_type, services_mapping)
-        for value in values(contributing_devices)
-            add_service!(psi_container,
-                         value.service,
-                         service_model,
-                         value.contributing_devices,
-                         devices_template; kwargs...)
-        end
+        service_devices = _filter_service_mapping(service_model.service_type, services_mapping)
```
jd-lara: We should be able to run this line inside the services loop in line 41

src/core/psi_container.jl
```diff
@@ -221,6 +221,7 @@ model_initial_time(psi_container::PSIContainer) = psi_container.initial_time
 #Internal Variables, Constraints and Parameters accessors
 get_variables(psi_container::PSIContainer) = psi_container.variables
 get_constraints(psi_container::PSIContainer) = psi_container.constraints
+get_constraints(psi_container::PSIContainer, name::Symbol) = psi_container.constraints[name]
```
sourabhdalvi: Isn't this the same as `get_constraint` ?

claytonpbarrows: indeed... I didn't see that.

src/services_models/reserves.jl
```diff
@@ -6,56 +6,60 @@ This function add the variables for reserves to the model
...
+                                                   ts_vector[t] * requirement)
+            constraint[name, t] = JuMP.@constraint(psi_container.JuMPmodel,
+                                        sum(reserve_variable[:,t]) >= param[name, t])
         end
     else
         for t in time_steps
-            constraint[t] = JuMP.@constraint(psi_container.JuMPmodel,
-                                    sum(reserve_variable[:,t]) >= ts_vector[t]*requirement)
+            constraint[name, t] = JuMP.@constraint(psi_container.JuMPmodel,
+                                    sum(reserve_variable[:,t]) >= ts_vector[t] * requirement)
```
jd-lara: space after comma

src/services_models/services_constructor.jl
```diff
@@ -10,31 +10,42 @@ function construct_services!(psi_container::PSIContainer,
...
+                            service_model,
+                            devices_template; kwargs...)
     end
     return
 end
 
-function add_service!(psi_container::PSIContainer, service::SR,
+function construct_service!(psi_container::PSIContainer, 
+                      services::IS.FlattenIteratorWrapper{SR},
+                      services_mapping::Dict{NamedTuple{(:type, :name),Tuple{DataType,String}},PSY.ServiceContributingDevices},
```
jd-lara: services_mapping type needs to be simplified.

PR #296, opened 2020-01-07
1 06 write results

Lilyhanig:
re-add the plot functionality (not by fuel)
add test for the creation of the plot recipe structs


src/PowerSimulations.jl
```diff
@@ -113,15 +113,14 @@ export get_con_index
...
-# export report
+export sort_data
+export get_stacked_plot_data
+export get_bar_plot_data
+export get_stacked_generation_data
+export get_stacked_aggregation_data
+export bar_plot
+export stack_plot
+export report
 # export make_fuel_dictionary
```
jd-lara: add a comment here about why this is commented out in the code please.

src/PowerSimulations.jl
```diff
@@ -229,6 +228,11 @@ include("utils/dual_results.jl")
 include("utils/simulation_results_reference.jl")
 include("utils/simulation_results.jl")
 include("utils/printing.jl")
+include("utils/plot_results.jl")
```
jd-lara: separate with a commented line `#Plots`

test/test_plot_results.jl
```diff
@@ -0,0 +1,68 @@
+IS.configure_logging(console_level = Logging.Info)
```
jd-lara: This test uses a simulation to make the plots. Is there a way to test using only Operations? I say this because it would be the third test that relies on fully running simulations and increases a lot the testing times.

src/utils/call_plots.jl
```diff
@@ -0,0 +1,180 @@
...
+function bar_plot(res::OperationsProblemResults; kwargs...)
+    default = hcat([Colors.RGBA(0.7,0.1,0.1,0.95)], # maroon
+              [Colors.RGBA(0,0,0,0.8)], [:lightblue], # Dark gray
+              [Colors.RGBA(0.33,0.42,0.18,0.9)], [:pink], # olive green
+              [Colors.RGBA(0.93,0.46,0,1)], [Colors.RGBA(0.56,0.28,0.54,1)], # orange, orchid
+              [Colors.RGBA(0.9,0.5,0.6,0.80)],  # dark pink
+              [Colors.RGBA(1, 1, 0.5, 0.6)], # light yellow
+              [Colors.RGBA(0.27, 0.5, 0.7, 0.9)], # steel blue
+              [Colors.RGBA(1, 0.757, 0.15, 01)], # canary yellow
+              [Colors.RGBA(0.8, 0.6, 0.3, 1)], [:red]) # khaki
```
daniel-thom: Are there any constants defined for the colors in the plot library?  If not, can you define them?

daniel-thom: Can you change the indentation here (and elsewhere) to line up with the new coding style guide?

src/utils/call_plots.jl
```diff
@@ -0,0 +1,180 @@
...
+    seriescolor = get(kwargs, :seriescolor, default)
+    key_name = string.(collect(keys(res.variables)))
+    for name in key_name
```
daniel-thom: Performance doesn't matter here, but it would be good to generally not create temporary arrays when you don't need them.  You could just iterate over the call to keys().

src/utils/call_plots.jl
```diff
@@ -0,0 +1,180 @@
...
+end
+
+function bar_plot(res::PSI.DualResults; kwargs...)
+    results = OperationsProblemResults(res.variables, res.total_cost, 
+    res.optimizer_log, res.time_stamp)
+    bar_plot(results; kwargs...)
+end
+function bar_plot(res::PSI.SimulationResults; kwargs...)
+    results = OperationsProblemResults(res.variables, res.total_cost, 
+    res.optimizer_log, res.time_stamp)
```
daniel-thom: indentation

src/utils/call_plots.jl
```diff
@@ -0,0 +1,180 @@
...
+    [Colors.RGBA(0,0,0,0.8)], [:lightblue], # Dark gray
+    [Colors.RGBA(0.33,0.42,0.18,0.9)], [:pink], # olive green
+    [Colors.RGBA(0.93,0.46,0,1)], [Colors.RGBA(0.56,0.28,0.54,1)], # orange, orchid
+    [Colors.RGBA(0.9,0.5,0.6,0.80)],  # dark pink
+    [Colors.RGBA(1, 1, 0.5, 0.6)], # light yellow
+    [Colors.RGBA(0.27, 0.5, 0.7, 0.9)], # steel blue
+    [Colors.RGBA(1, 0.757, 0.15, 01)], # canary yellow
+    [Colors.RGBA(0.8, 0.6, 0.3, 1)], [:red]) # khaki
+    seriescolor = get(kwargs, :seriescolor, default)  
+    key_name = string.(collect(keys(res.variables)))
```
daniel-thom: Ditto regarding iteration over keys()

src/utils/plot_results.jl
```diff
@@ -54,7 +54,10 @@ function get_stacked_plot_data(res::OperationsProblemResults, variable::String;
 
     data_matrix = convert(Matrix, variable)
     labels = collect(names(variable))
-    legend = string.(labels)
+    legend = [names(variable)[1]]
```
daniel-thom: In the existing code it seems odd that `variable` is passed into the function and then immediately overwritten.

What does `names()` do?  That is a Base function to get the names exported by a module.

daniel-thom: `names` is called twice. You could use `labels` in the second case.  Also, should you verify that the length of `labels` is greater than 2?  Or is that guaranteed?

src/utils/plot_results.jl
```diff
@@ -70,13 +73,16 @@ function get_bar_plot_data(res::OperationsProblemResults, variable::String; kwar
     if isnothing(sort)
         variable = variable[:, alphabetical]
     else
-        variable = variable[:,sort]
+        variable = variable[:, sort]
```
daniel-thom: I know that `sort` is existing code, but you shouldn't override Base function.

test/test_plot_results.jl
daniel-thom: You shouldn't reconfigure logging here.  If you want to enable Info prints then you should set the relevant environment variables.

test/test_plot_results.jl
```diff
@@ -0,0 +1,68 @@
...
+if !isdir(joinpath(pwd(), "testing_plots"))
```
daniel-thom:
```suggestion
file_path = joinpath(pwd(), "testing_plots")
if !isdir(file_path)
    mkdir(file_path)
end
```

test/test_plot_results.jl
```diff
@@ -0,0 +1,68 @@
...
+                        cache = Dict("ED" => [TimeStatusChange(:ON_ThermalStandard)]),
+                        ini_cond_chronology = Dict("UC" => Consecutive(), "ED" => Consecutive())
+                        )
+
+    sim = Simulation(name = "test",
+                        steps = 1, step_resolution =Hour(24),
+                        stages = stages_definition,
+                        stages_sequence = sequence,
+                        simulation_folder= file_path,
+                        verbose = true)
```
daniel-thom: indentation

test/test_plot_results.jl
```diff
@@ -0,0 +1,68 @@
+IS.configure_logging(console_level = Logging.Info)
+if !isdir(joinpath(pwd(), "testing_plots"))
+    file_path = mkdir(joinpath(pwd(), "testing_plots"))
+else
+    file_path = joinpath(pwd(), "testing_plots")
+end
+
+function test_plots(file_path::String)
```
daniel-thom: For readability, avoid overriding the global variable.

test/test_plot_results.jl
```diff
@@ -0,0 +1,68 @@
...
+                        verbose = true)
+    build!(sim)
+    sim_results = execute!(sim; verbose = true)
+    res_UC = load_simulation_results(sim_results, "UC")
+    res_ED = load_simulation_results(sim_results, "ED")
+
+    @testset "testing bar plot" begin
+        for res in [res_UC, res_ED]
+            results = PSI.OperationsProblemResults(res.variables, res.total_cost, res.optimizer_log, res.time_stamp)
+            key_name = collect(keys(results.variables))
```
daniel-thom: No need to `collect`.

test/test_plot_results.jl
```diff
@@ -0,0 +1,68 @@
...
+                for name in key_name
+                        variable_bar = get_bar_plot_data(results, string(name))
+                        sort = sort!(names(results.variables[name]))
```
daniel-thom: Ditto regarding `sort`

test/test_plot_results.jl
daniel-thom: Can the test load the outputs from a saved simulation?

jd-lara: @daniel-thom I added this to all the tests that require Simulations. What's the alternative approach?

Lilyhanig: @daniel-thom do you have an example of how to set the environmental variables? I'm not sure exactly what you mean

daniel-thom: The `run_tests` function in test/runtests.jl configures a global logger for all unit tests.  It sets the console logging to Error.  It sounds like you may just want to set that to Info.

If you don't want to make that change then you can set an environment variable to override the default.  However, this method doesn't work if you are running `Pkg.test()`.  It only works if you are doing `include(test/runtests.jl)` in the REPL.

```
ENV["SYS_CONSOLE_LOG_LEVEL"] = "Info"
include(test/runtests.jl)
```

Reference:  https://github.com/NREL/InfrastructureSystems.jl/blob/master/docs/src/man/tests.md

jd-lara:
@Lilyhanig is this PR superseded by #299? If so please close it

Lilyhanig:
this has a lot of commits and some things have been changed so I'm just going to open a new PR

PR #299, opened 2020-01-08
add tests for write results

Lilyhanig:
add test for all write_results functions across operation, simulation, and simulation with dual constraints in the results.

### DIFF
--- a/src/routines/solve_routines.jl
+++ b/src/routines/solve_routines.jl
@@ -114,7 +114,7 @@ function execute!(sim::Simulation; verbose::Bool = false, kwargs...)
     isnothing(sim.internal) && error("Simulation not built, build the simulation to execute")
     sim.internal.raw_dir, sim.internal.models_dir, sim.internal.results_dir = _prepare_workspace(sim.name, sim.simulation_folder)
     _build_stage_paths!(sim, verbose = verbose; kwargs...)
-
+    dual_constraints = get(kwargs, :dual_constraints, nothing)
     steps = get_steps(sim)
     for s in 1:steps
         verbose && println("Step $(s)")
@@ -130,7 +130,6 @@ function execute!(sim::Simulation; verbose::Bool = false, kwargs...)
                 raw_results_path = joinpath(sim.internal.raw_dir, run_name, replace_chars("$(sim.internal.current_time)", ":", "-"))
                 mkpath(raw_results_path)
                 update_stage!(stage, s, sim)
-                dual_constraints = get(kwargs, :dual_constraints, nothing)
                 _run_stage(stage, sim.internal.current_time, raw_results_path; duals = dual_constraints)
                 sim.internal.run_count[s][stage_number] += 1
                 sim.internal.date_ref[stage_number] = sim.internal.date_ref[stage_number] + stage_interval
@@ -140,6 +139,6 @@ function execute!(sim::Simulation; verbose::Bool = false, kwargs...)
         end
 
     end
-    sim_results = SimulationResultsReference(sim)
+    sim_results = SimulationResultsReference(sim; dual_constraints = dual_constraints)
     return sim_results
 end
--- a/src/utils/dual_results.jl
+++ b/src/utils/dual_results.jl
@@ -23,9 +23,9 @@ end
 
 # internal function to parse through the reference dictionary and grab the file paths
 function _reading_references(results::Dict, duals::Array, stage::String, step::Array,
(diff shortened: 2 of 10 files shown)


src/routines/solve_routines.jl
```diff
@@ -114,7 +114,7 @@ function execute!(sim::Simulation; verbose::Bool = false, kwargs...)
     isnothing(sim.internal) && error("Simulation not built, build the simulation to execute")
     sim.internal.raw_dir, sim.internal.models_dir, sim.internal.results_dir = _prepare_workspace(sim.name, sim.simulation_folder)
     _build_stage_paths!(sim, verbose = verbose; kwargs...)
-
+    dual_constraints = get(kwargs, :dual_constraints, nothing)
```
jd-lara: constraints_duals, please change the kwarg too.

src/routines/solve_routines.jl
```diff
@@ -140,6 +139,6 @@ function execute!(sim::Simulation; verbose::Bool = false, kwargs...)
         end
 
     end
-    sim_results = SimulationResultsReference(sim)
+    sim_results = SimulationResultsReference(sim; dual_constraints = dual_constraints)
```
jd-lara: same here

src/utils/dual_results.jl
```diff
@@ -23,9 +23,9 @@ end
 
 # internal function to parse through the reference dictionary and grab the file paths
 function _reading_references(results::Dict, duals::Array, stage::String, step::Array,
```
jd-lara: _read_references* In the future prefer to use plain form verbs instead of the present participle.

src/utils/dual_results.jl
```diff
@@ -35,23 +35,21 @@ function _reading_references(results::Dict, duals::Array, stage::String, step::A
         for (ix,time) in enumerate(step_df.Date)
             file_path = step_df[ix, :File_Path]
             var = Feather.read("$file_path")
-            correct_var_length = size(1:(size(var, 1) - extra_time), 1)
-            results[name] = vcat(results[name],var[1:correct_var_length,:])
+            results[name] = vcat(results[name],var[1:time_length, :])
```
jd-lara: space after comma

src/utils/dual_results.jl
```diff
@@ -35,23 +35,21 @@ function _reading_references(results::Dict, duals::Array, stage::String, step::A
...
+                             references::Dict, time_length::Int64)
     for name in dual
         date_df = references[stage][name]
         results[name] = DataFrames.DataFrame()
         for (ix,time) in enumerate(date_df.Date)
             file_path = date_df[ix, :File_Path]
             var = Feather.read(file_path)
-            correct_var_length = size(1:(size(var, 1) - extra_time_length), 1)
-            results[name] = vcat(results[name],var[1:correct_var_length,:])
+            results[name] = vcat(results[name],var[1:time_length,:])
```
jd-lara: space after comma

test/Project.toml
```diff
@@ -1,4 +1,5 @@
 [deps]
+CSV = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
```
jd-lara: CSV is only added to the test environment. Is this the intention?

test/runtests.jl
```diff
@@ -52,7 +52,7 @@ function get_logging_level(env_name::String, default)
 end
 
 function run_tests()
-    console_level = get_logging_level("SYS_CONSOLE_LOG_LEVEL", "Error")
+    console_level = get_logging_level("SYS_CONSOLE_LOG_LEVEL", "Info")
```
jd-lara: It is best to change the logging level only when running the simulations tests. @daniel-thom is it possible to reset the logging level for the simulation tests?

test/test_dual_constraints.jl
```diff
@@ -0,0 +1,52 @@
+path = joinpath(pwd(), "test_dual_constraints")
```
jd-lara: we can fold the tests for results into a single file. No need to have a separate test file for getting the duals.

claytonpbarrows:
@Lilyhanig, sorry I hastily merged this. Then I un-merged to wait for your changes per @jd-lara requests.

test/Project.toml
Lilyhanig: yes because there is a function to write to a csv (but no way to read it back) if the user really wants to read the results in csv form, but it shouldn't be a dependency in PowerSimulations, but it should be in the test environment for that function

PR #302, opened 2020-01-09
Prevent Travis-CI from timing out

kdheepak:
This line cause this backgrounded process to `sleep` for 9 minutes, and then print "still running" along with the seconds since it started running.

```
while sleep 9m; do echo "=====[ $SECONDS seconds still running ]====="; done &
```

Anything run after it need not output anything to the console. Once done, `kill %1` will kill the last background process. 

We need to do the above because on Travis any job that hasn't printed anything to the screen in 10 minutes will time out (Any job that is running for more than 50 minutes will be timed out regardless of whether it is printing anything to the screen or not).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,13 @@ addons:
 #before_script: # homebrew for mac
 #  - if [ $TRAVIS_OS_NAME = osx ]; then brew install gcc; fi
 
+script:
+  - while sleep 9m; do echo "=====[ $SECONDS seconds still running ]====="; done &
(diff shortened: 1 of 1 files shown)


jd-lara:
@kdheepak I am not sure this works, Travis is timing out

jd-lara:
@daniel-thom do we still need this after the latest changes to the Logging?

daniel-thom:
Probably.  We no longer do anything with solver output.  If a solver is configured to output text regularly then nothing will timeout.  However, if output is disabled and a solver takes 10 minutes to complete then a timeout will occur.  It would be better to guarantee periodic text output.

PR #304, opened 2020-01-10
changes to plots and plot tests, consolidate simulation tests

Lilyhanig:


test/test_simulation_build.jl
```diff
@@ -1,28 +1,27 @@
-IS.configure_logging(console_level = Logging.Info)
```
jd-lara: rename file to `test_simulation.jl`

test/test_simulation_loading.jl
```diff
@@ -1,28 +1,26 @@
-IS.configure_logging(console_level = Logging.Info)
```
jd-lara: rename to `test_simulation_solution_loading.jl`

Lilyhanig:
@claytonpbarrows this can be merged now

PR #305, opened 2020-01-13
Adding IntegralLimitFF implementation for HydroDispatch

sourabhdalvi:
- Allowing HydroDispatch constructor to create IntegralFF constraint or Energy Limit constraint depending on the FeedForward passed

This is the same PR as the previous one but clean git history

### DIFF
--- a/src/devices_models/devices/hydro_generation.jl
+++ b/src/devices_models/devices/hydro_generation.jl
@@ -349,7 +349,11 @@ function _get_budget(psi_container::PSIContainer,
         tech = PSY.get_tech(device)
         # This is where you would get the water/energy storage capacity
         # which is then multiplied by the forecast value to get you the energy budget
-        energy_capacity = use_forecast_data ? PSY.get_storage_capacity(device) : PSY.get_activepower(device)
+        if use_forecast_data
+            energy_capacity = PSY.get_storage_capacity(device)
+        else
+            energy_capacity = PSY.get_activepower(device)
+        end
         if use_forecast_data
             ts_vector = TS.values(PSY.get_data(PSY.get_forecast(PSY.Deterministic,
                                                                 device,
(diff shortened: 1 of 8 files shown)


src/devices_models/devices/hydro_generation.jl
```diff
@@ -330,12 +349,12 @@ function _get_budget(psi_container::PSIContainer,
         tech = PSY.get_tech(device)
         # This is where you would get the water/energy storage capacity
         # which is then multiplied by the forecast value to get you the energy budget
-        energy_capacity = use_forecast_data ? PSY.get_storagecapacity(device) : PSY.get_activepower(device)
+        energy_capacity = use_forecast_data ? PSY.get_storage_capacity(device) : PSY.get_activepower(device)
```
jd-lara: we need to eliminate ternary operators.

src/devices_models/devices/hydro_generation.jl
```diff
@@ -374,18 +400,18 @@ function device_budget_param_ub(psi_container::PSIContainer,
...
+    param = add_param_container!(psi_container, param_reference, set_name, 1:no_of_budgets)
 
     for data in budget_data, i in 1:no_of_budgets
         name = data[1]
         forecast = data[4][i]
         multiplier = data[3]
-        param[name] = PJ.add_parameter(psi_container.JuMPmodel, forecast)
-        constraint[name] = JuMP.@constraint(psi_container.JuMPmodel,
-                    sum([variable[name, t] for t in time_chunks[:, i]]) <= multiplier*param[name])
+        param[name,i] = PJ.add_parameter(psi_container.JuMPmodel, forecast)
```
jd-lara: space after commas

src/devices_models/devices/hydro_generation.jl
daniel-thom: IMHO, that was a completely acceptable use of ternary operators.  Discussed here:  https://github.com/NREL/PowerSystems.jl/blob/master/docs/src/developer/style.md#one-line-conditionals

PR #307, opened 2020-01-15
Jd/sim interface

### LINKED ISSUE #297: Update Simulation Interface
- Remove "to_executions" from the "Synchronize" inputs and calculate base on intervals in SimulationSequence:
   - `Synchronize` currently takes `from_steps` and `to_executions` arguments. We currently don't support any use cases where `to_executions` can't be inferred from the information included in the `intervals` argument of `SimulationSequence`.

 - Remove "step_resolution" from "Simulation" arguments and calculate based on the information in SimulationSequence.

jd-lara:
closes #297 and addresses #284 partially


src/core/operations_problem.jl
```diff
@@ -376,31 +375,31 @@ function build_op_problem!(op_problem::OperationsProblem{M}; kwargs...) where M<
 end
 
 function _build!(psi_container::PSIContainer, template::OperationsProblemTemplate, sys::PSY.System; kwargs...)
-    verbose = get(kwargs, :verbose, true)
     transmission = template.transmission
 
     # Order is required
     #Build Services
+    # TODO: Add info print
```
claytonpbarrows: Can we also add an info print where the PSI container is created? More specifically, where the empty JuMP model is created. I'd like this because some of the solvers (e.g. Xpress#jg/new_interface) return solver log prints at that point, and in our console log it's a bit abrupt to have this empty solver log print without any context. I think this would help.

test/Project.toml
```diff
@@ -13,13 +13,15 @@ Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
 Memento = "f28f55f0-a522-5efc-85c2-fe41dfb9b2d9"
 OSQP = "ab2f91bb-94b4-55e3-9ba0-7f65df51de79"
 ParameterJuMP = "774612a8-9878-5177-865a-ca53ae2495f9"
+Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
 PowerModels = "c36e90e8-916a-50a6-bd94-075b64ef4655"
+PowerSimulations = "e690365d-45e2-57bb-ac84-44ba829e73c4"
```
claytonpbarrows: Doesn't this break the CI testing? @daniel-thom told me not to add PSI here.

daniel-thom: Correct. This should not be here.  When you run `]test` PSI gets added implicitly.

src/core/chronology.jl
```diff
@@ -4,40 +4,40 @@ Defines a logical sequence for simulation within one stage.
...
-                to_executions::Int64
-                )
-Defines the co-ordination of time between Two stages.  
+    Synchronize(steps::Int64)
+Defines the co-ordination of time between Two stages.
 
 # Arguments
-- `from_steps::Int64`: Number of time periods to grab data from
-- `to_executions::Int64`: Number of times to run using the same data
+- `steps::Int64`: Number of time periods to grab data from
```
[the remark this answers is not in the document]
jd-lara: Good catch

src/core/chronology.jl
```diff
@@ -4,40 +4,40 @@ Defines a logical sequence for simulation within one stage.
...
+    function Synchronize(; steps)
+        new(steps)
     end
 end
 
 """
-    RecedingHorizon(step::Int64
-                    )                             
-""" # TODO: Add DocString    
+    RecedingHorizon(step::Int64)
```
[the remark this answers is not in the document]
jd-lara: RH has a different behavior than Synchronize 1

src/core/operations_problem.jl
jd-lara: Would it make more sense to have a @debug statement? It seems that your request is introducing a general behavior to address a specific solver output print. @daniel-thom thoughts?

test/Project.toml
jd-lara: This was an oversight when I updated all the other toml files.

src/core/operations_problem.jl
daniel-thom: As long as a message isn't getting logged repeatedly I tend to err on the side of logging more.  If it's helpful to the user, log it.

PR #308, opened 2020-01-17
adding new HydroDispatch

sourabhdalvi:
- Adding new formulation for `HydroDispatch` which can track energy stored in reservoir, account for inflow and spillage.
- The new formulation is called `HydroDispatchReservoirFlow` (can be named something else), the need for this emerges if we apply energy balance constraint to reservoir storage then the energy limit constraint shouldn't be applied.
- If a energy limits need to be applied, if should be enforced on the `E_HydroDispatch` variable.

### DIFF
--- a/src/devices_models/devices/hydro_generation.jl
+++ b/src/devices_models/devices/hydro_generation.jl
@@ -36,6 +36,25 @@ function reactivepower_variables!(psi_container::PSIContainer,
     return
 end
 
+function energy_storage_variables!(psi_container::PSIContainer,
+                                  devices::IS.FlattenIteratorWrapper{H}) where H<:PSY.HydroGen
+    add_variable(psi_container,
+                 devices,
+                 Symbol("E_$(H)"),
+                 false;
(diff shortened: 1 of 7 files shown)


claytonpbarrows:
Most to least complex:
1. Inflow, spillage, storage, dispatchable
2. Inflow, storage, dispatchable
3. Inflow, dispatchable
4. Fixed

Seems like this covers 1., and we just need to make sure that we relabel `HyroFix` in PSY to `HydroROR` and make sure that we can create 3. and 4. for it.

src/devices_models/devices/hydro_generation.jl
```diff
@@ -235,6 +255,128 @@ function activepower_constraints!(psi_container::PSIContainer,
     return
 end
 
+######################## output constraints without Time Series ############################
+function _get_inflow_time_series(psi_container::PSIContainer,
```
claytonpbarrows: This feels like we should be able to do it in the `_get_time_series` function. But I think that's a separate PR.

PR #310, opened 2020-01-21
Change format of PSIContainer keys

daniel-thom:
This PR changes the use of hard-coded symbols in PSIContainer keys.  It also converts arbitrary references to PSIContainer member variables to accessor functions.  All tests pass but I'm still reviewing some things for accuracy.  Please focus on high-level changes for now.


src/core/definitions.jl
```diff
@@ -42,3 +42,47 @@ const SIMULATION_BUILD_KWARGS = [
...
+const REACTIVE = "reactive"
+const REACTIVE_RANGE = "reactiverange"
+const FP_FT = "FpFT"
+const FP_TF = "FpTF"
+const FQ_FT = "FqFT"
+const FQ_TF = "FqTF"
+
+# Parameters
+const ACTIVE_POWER = "activepower"
+const ENERGY = "energy"
```
daniel-thom: Please feel free to suggest new names for any of these.  One thing I considered was whether to prefix each type differently (VAR_START, CONSTRAINT_RAMP, etc.).

src/devices_models/devices/thermal_generation.jl
```diff
@@ -299,25 +307,24 @@ function ramp_constraints!(psi_container::PSIContainer,
                            feed_forward::Union{Nothing, AbstractAffectFeedForward}) where {T<:PSY.ThermalGen,
                                                     D<:AbstractThermalFormulation,
                                                     S<:PM.AbstractPowerModel}
-    key = ICKey(DevicePower, T)
-    if !(key in keys(psi_container.initial_conditions))
-        error("Initial Conditions for $(T) Rate of Change Constraints not in the model")
-    end
```
daniel-thom: This should not have been deleted. Will fix.

src/core/aux_structs.jl
```diff
@@ -3,4 +3,8 @@ struct UpdateRef{T}
     access_ref::Union{Symbol,String}
 end
 
+function UpdateRef{T}(name::AbstractString, ::Type{U}) where {T, U <: PSY.Device}
```
daniel-thom: I'm thinking that we should one or more explicit constructors for UpdateRef.  Users will need better advice on how to create UpdateRef instances.
1. This current one is used for storing ParameterJuMP variables in the format (<basename>_<device_type>).
2. The code creates UpdateRef instances with a string where that string is an accessor function for a forecast. We could create a specific constructor for this purpose.
3. Symbols are passed in other cases that I'm still going through. I'm not yet sure if we need those.

jd-lara: I agree with the idea to have functions that return the UpdateRef objects. These were never meant to be used directly by users. Since we have discussed the use of get and set functions I never intended for a user to have to write an UpdateRef object manually.

src/core/psi_container.jl
```diff
@@ -222,8 +262,146 @@ model_initial_time(psi_container::PSIContainer) = psi_container.initial_time
...
+    psi_container::PSIContainer,
+    variable_type::AbstractString,
+    device_type::Type{<:PSY.Device},
+    value,
+)
+    set_variable!(psi_container, variable_name(variable_type, device_type), value)
+    return
+end
+
+function set_variable!(
```
jd-lara: we should change this name to `assign` or `fill` because in the optimization lingo this can get confused with setting the variable value

src/core/psi_container.jl
```diff
@@ -222,8 +262,146 @@ model_initial_time(psi_container::PSIContainer) = psi_container.initial_time
...
+    psi_container::PSIContainer,
+    constraint_type::AbstractString,
+    device_type::Type{<:PSY.Device},
+    value,
+)
+    set_constraint!(psi_container, constraint_name(constraint_type, device_type), value)
+    return
+end
+
+function set_constraint!(
```
jd-lara: same here about the function name to avoid confusion on the terminology.

daniel-thom:
@jd-lara I'm done with the changes we discussed.  If you approve them then we can merge and continue with the new work in a feature branch.

src/devices_models/devices/AC_branches.jl
```diff
@@ -104,13 +92,8 @@ function branch_rate_bounds!(psi_container::PSIContainer,
...
               DeviceRange(name, limit_values, services_ub, Vector{Symbol}()))
     end
-    set_variable_bounds(psi_container,
-                        constraint_data,
-                        Symbol("FpFT_$(B)"))
-
-    set_variable_bounds(psi_container,
-                        constraint_data,
-                        Symbol("FpTF_$(B)"))
+    set_variable_bounds!(psi_container, constraint_data, FLOW_REAL_POWER_FROM_TO, B)
```
jd-lara: `FLOW_REAL_POWER_FROM_TO` could be a constraint or variable name. This is an example of such a case.

PR #312, opened 2020-01-25
Updating Hydro Formulations

sourabhdalvi:
- Adding new formulation for `HydroDispatch` which can track energy stored in reservoir, account for inflow and spillage.
- Changed `HydroDispatchSeasonalFlow` - > `HydroDispatchReservoirFlow`
- The new formulation is called `HydroDispatchReservoirStorage` (can be named something else), the need for this emerges if we apply energy balance constraint to reservoir storage then the energy limit constraint shouldn't be applied.
- Added capability to model HydroFixed as HydroROR

### DIFF
--- a/src/devices_models/devices/hydro_generation.jl
+++ b/src/devices_models/devices/hydro_generation.jl
@@ -499,9 +499,18 @@ function cost_function(psi_container::PSIContainer,
                        system_formulation::Type{<:PM.AbstractPowerModel}) where D<:AbstractHydroFormulation
     add_to_cost(psi_container,
                 devices,
-                Symbol("P_HydroDispatch"),
+                variable_name(REAL_POWER, PSY.HydroDispatch),
                 :fixed,
                 -1.0)
+
+    return
+end
+
+function cost_function(psi_container::PSIContainer,
+                       devices::IS.FlattenIteratorWrapper{H},
+                       device_formulation::Type{D},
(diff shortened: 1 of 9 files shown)


claytonpbarrows:
@sourabhdalvi it looks like you're still missing the `device_budget_param_ub` definitions...

sourabhdalvi:
> @sourabhdalvi it looks like you're still missing the `device_budget_param_ub` definitions...
I forgot to rename those function calls.

src/devices_models/devices/hydro_generation.jl
```diff
@@ -28,14 +30,51 @@ function reactivepower_variables!(psi_container::PSIContainer,
...
     return
 end
 
+function energy_variables!(psi_container::PSIContainer,
+                                 devices::IS.FlattenIteratorWrapper{H}) where H<:PSY.HydroGen
+    add_variable(psi_container,
+                 devices,
+                 variable_name(ENERGY, H),
+                 false;
+                 ub_value = d -> d.storage_capacity,
```
claytonpbarrows: `ub_value = d -> get_storage_capacity(d)`

src/devices_models/devices/hydro_generation.jl
```diff
@@ -28,14 +30,51 @@ function reactivepower_variables!(psi_container::PSIContainer,
...
+                 lb_value = d -> 0.0,
+                 init_value = d -> d.initial_storage)
```
claytonpbarrows: `init_value = d -> initial_storage(d)`

PR #313, opened 2020-01-25
changing updateref keys in nodal_expression!

claytonpbarrows:
Per email from @daniel-thom:
The function nodal_expression! for each `RenewableGen`, `ThermalGen`, and `HydroGen` does something like the code below:
 ```julia
        include_parameters(canonical,
                        ts_data_active,
                        UpdateRef{L}("get_maxactivepower"),
                        :nodal_balance_active,
                        -1.0)
        include_parameters(canonical,
                        ts_data_reactive,
                        UpdateRef{L}("get_maxactivepower"),
                        :nodal_balance_reactive,
                        -1.0)
```
Because the keys are the same the 2nd parameter container overwrites the first.  This probably mucks up the results.

### DIFF
--- a/src/devices_models/devices/hydro_generation.jl
+++ b/src/devices_models/devices/hydro_generation.jl
@@ -278,11 +278,11 @@ function nodal_expression!(psi_container::PSIContainer,
     if parameters
         include_parameters(psi_container,
                            ts_data_active,
-                           UpdateRef{H}("get_rating"),
+                           UpdateRef{H}("get_maxactivepower"),
                            :nodal_balance_active)
         include_parameters(psi_container,
                            ts_data_reactive,
-                           UpdateRef{H}("get_rating"),
+                           UpdateRef{H}("get_maxreactivepower"),
                            :nodal_balance_reactive)
         return
     end
--- a/src/devices_models/devices/hydro_generation.jl
+++ b/src/devices_models/devices/hydro_generation.jl
@@ -281,11 +281,11 @@ function nodal_expression!(psi_container::PSIContainer,
     if parameters
         include_parameters(psi_container,
                            ts_data_active,
-                           UpdateRef{H}("get_maxactivepower"),
+                           UpdateRef{H}("get_activepower"),
                            :nodal_balance_active)
         include_parameters(psi_container,
                            ts_data_reactive,
-                           UpdateRef{H}("get_maxreactivepower"),
+                           UpdateRef{H}("get_reactivepower"),
                            :nodal_balance_reactive)
         return
     end
--- a/src/devices_models/devices/hydro_generation.jl
+++ b/src/devices_models/devices/hydro_generation.jl
@@ -116,14 +116,17 @@ function _get_time_series(psi_container::PSIContainer,
         bus_number = PSY.get_number(PSY.get_bus(device))
         name = PSY.get_name(device)
         tech = PSY.get_tech(device)
-        # pf = sin(acos(PSY.get_powerfactor(PSY.get_tech(device))))
-        active_power = use_forecast_data ? PSY.get_rating(tech) : PSY.get_activepower(device)
+        pf = PSY.get_reactivepower(device) / PSY.get_rating(tech)       
         if use_forecast_data
+            active_power = PSY.get_rating(tech)
+            reactive_power = PSY.get_rating(tech) * pf
             ts_vector = TS.values(PSY.get_data(PSY.get_forecast(PSY.Deterministic,
                                                                 device,
                                                                 initial_time,
                                                                 "get_rating")))
         else
+            active_power = PSY.get_activepower(device)
+            reactive_power = PSY.get_reactivepower(device) 
             ts_vector = ones(time_steps[end])
         end
         range_data = DeviceRange(name, get_constraint_values(device))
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/devices_models/devices/hydro_generation.jl [hunk 1]
```diff
@@ -278,11 +278,11 @@ function nodal_expression!(psi_container::PSIContainer,
     if parameters
         include_parameters(psi_container,
                            ts_data_active,
-                           UpdateRef{H}("get_rating"),
+                           UpdateRef{H}("get_maxactivepower"),
```
claytonpbarrows: There is no `get_maxactivepower()` for <:HydroGen

src/devices_models/devices/hydro_generation.jl
```diff
@@ -278,11 +278,11 @@ function nodal_expression!(psi_container::PSIContainer,
...
                            :nodal_balance_active)
         include_parameters(psi_container,
                            ts_data_reactive,
-                           UpdateRef{H}("get_rating"),
+                           UpdateRef{H}("get_maxreactivepower"),
```
claytonpbarrows: There is no `get_maxreactivepower()` for `<:HydroGen`

src/devices_models/devices/renewable_generation.jl
```diff
@@ -171,11 +171,11 @@ function nodal_expression!(psi_container::PSIContainer,
     if parameters
         include_parameters(psi_container,
                            ts_data_active,
-                           UpdateRef{R}("get_rating"),
+                           UpdateRef{R}("get_maxactivepower"),
```
claytonpbarrows: There is no `get_maxactivepower()` for `<:RenewableGen`

src/devices_models/devices/renewable_generation.jl
```diff
@@ -171,11 +171,11 @@ function nodal_expression!(psi_container::PSIContainer,
...
                            :nodal_balance_active)
         include_parameters(psi_container,
                            ts_data_reactive,
-                           UpdateRef{R}("get_rating"),
+                           UpdateRef{R}("get_maxreactivepower"),
```
claytonpbarrows: There is no `get_maxreactivepower()` for `<:RenewableGen`

claytonpbarrows:
The challenge here is that the time series data that is passed to the `include_parameters` function is gathered based on different scaling values depending on what's happening (see `_get_time_series()`). So, at the point where we are creating `UpdateRef{R}("accessor_function"), we don't know what accessor function was used to gather the time series. 
 1. Does it matter? i.e. do we need consistency between the UpdateRef name and the time series data access?
   a. If not, then I propose to use "get_activepower" and "get_reactivepower" 
   b. If so, then I propose to create the UpdateRef instances when we gather the time series.

jd-lara:
We might need to define a better key for the parameters dict and not rely only on UpdateRef since there might be UpdateRef rules that are the same for several cases.

daniel-thom:
@claytonpbarrows Tangential question:  This code path for `ElectricLoads` stores time series data for `max_activepower` in variables `ts_data_active` and `ts_data_reactive`. Refer to where `get_forecast` is called in `_get_time_series` in `src/devices_models/devices/electric_loads.jl`.  Is that correct?

claytonpbarrows:
@daniel-thom, that's correct. The story follows:
We usually only have time series data for active power (mostly because that's what is primarily monetized). Also, it's a reasonable assumption that the ratio between peak active power and peak reactive power consumption at a particular location is relatively constant. So, in this case we use the same scaling factor time series profile and multiply it by the peak active power and peak reactive power to get the respective time series values.

src/devices_models/devices/hydro_generation.jl
```diff
@@ -116,22 +116,25 @@ function _get_time_series(psi_container::PSIContainer,
...
+            reactive_power = PSY.get_reactivepower(device) 
             ts_vector = ones(time_steps[end])
         end
         range_data = DeviceRange(name, get_constraint_values(device))
         _device_services!(range_data, device, model)
         push!(constraint_data, range_data)
         push!(active_timeseries, DeviceTimeSeries(name, bus_number, active_power, ts_vector,
                                                  range_data))
-        push!(reactive_timeseries, DeviceTimeSeries(name, bus_number, active_power,
+        push!(reactive_timeseries, DeviceTimeSeries(name, bus_number, reactive_power,
```
claytonpbarrows: @daniel-thom, just a heads up that if you aren't going to use the rest of these changes, you should probably pick up this one.

src/devices_models/devices/renewable_generation.jl
```diff
@@ -95,15 +95,18 @@ function _get_time_series(psi_container::PSIContainer,
...
+            reactive_power = PSY.get_rating(tech) * pf
             forecast = PSY.get_forecast(PSY.Deterministic,
                                         device,
                                         initial_time,
                                         "get_rating",
                                         length(time_steps))
             ts_vector = TS.values(PSY.get_data(forecast))
         else
+            active_power = PSY.get_activepower(device)
+            reactive_power = PSY.get_reactivepower(device)
```
claytonpbarrows: @daniel-thom also, this one.

jd-lara:
@daniel-thom and @claytonpbarrows do #317 closes this?

claytonpbarrows:
closing in lieu of #327

PR #318, opened 2020-01-30
speed up tests by consolidation

Lilyhanig:
- consolidated
- moved build! checks to unit tests
- is checking everything for 14 bus and 5 bus necessary? see network constructors
- consolidated execute! in one spot


jd-lara:
Nodes5 only has lines as branches. Nodes14 has transformers and tap transformers. This is critical for network_constructions tests

test/test_operations_solve.jl
```diff
@@ -24,7 +27,7 @@ end
 @testset "Solving ED with PTDF Models" begin
     template = OperationsProblemTemplate(StandardPTDFModel, devices, branches, services);
     parameters_value = [true, false]
-    systems = [c_sys5, c_sys14, c_sys14_dc]
+    systems = [c_sys5, c_sys14_dc] #c_sys14
```
jd-lara: @Lilyhanig add 14 buses here too

test/test_operations_solve.jl
```diff
@@ -61,7 +64,7 @@ end
 end
 
 @testset "Solving ED With PowerModels with linear convex models" begin
-    systems = [c_sys5, c_sys14]
+    systems = [c_sys5] #c_sys14
```
jd-lara: enable 14 bus here too

test/test_base_PowerModels_interface.jl
```diff
@@ -24,12 +24,12 @@ case5_dc_data = PM.replicate(case5_dc_data, 2)
...
+    #pm = PowerSimulations.build_nip_model(case5_data, DCPPowerModel)
+    #JuMP.num_variables(pm.model) == 34 # Repetitive?
+    #pm = PowerSimulations.build_nip_model(case5_data, ACPPowerModel)
+    #JuMP.num_variables(pm.model) == 96
     pm = PowerSimulations.build_nip_model(case5_dc_data, DCPPowerModel)
-    JuMP.num_variables(pm.model) == 36
+    @test JuMP.num_variables(pm.model) == 36
     pm = PowerSimulations.build_nip_model(case5_dc_data, DCPPowerModel)
-    JuMP.num_variables(pm.model) == 48
+    @test JuMP.num_variables(pm.model) == 36 # 48 was originally here, verify??
```
jd-lara: Why did this test change?

test/test_base_PowerModels_interface.jl
```diff
@@ -24,12 +24,12 @@ case5_dc_data = PM.replicate(case5_dc_data, 2)
 end
 
 @testset "PM with type extensions" begin
-    pm = PowerSimulations.build_nip_model(case5_data, DCPPowerModel)
-    JuMP.num_variables(pm.model) == 34
-    pm = PowerSimulations.build_nip_model(case5_data, ACPPowerModel)
-    JuMP.num_variables(pm.model) == 96
+    #pm = PowerSimulations.build_nip_model(case5_data, DCPPowerModel)
```
jd-lara: These tests are different, one includes dc data and the other one doesn't

test/test_base_PowerModels_interface.jl
Lilyhanig: it wasn't actually testing, there was no @test ... when I added the @test it failed because it was the wrong value, that's why I added the comment, thoughts on what it should be? (Is the test wrong or is it actually failing?)

test/test_base_PowerModels_interface.jl
Lilyhanig: both did not include dc data (case5_data) so it was repetitive, I can change it to be case5_dc_data... I kind of figured that was the intention but wasn't sure.

test/test_base_PowerModels_interface.jl
jd-lara: This might have changed with PM. When I make the update to PM v0.14 I will test it again.

PR #329, opened 2020-02-04
fix to plots found, fix to mkdir issue in tests in appveyor

Lilyhanig:
- found a color issue for PlotlyJS for big datasets that is now fixed
- fixed test issue with appveyor for test_operations_solve


test/test_operations_solve.jl
```diff
@@ -286,7 +286,6 @@ function test_write_functions(file_path)
 
     @testset "test write result functions" begin
         new_path = joinpath(file_path, "seven")
-        @test_throws SystemError PSI.write_results(res, new_path) # not yet a directory
```
daniel-thom: I'm fine with the change, but why was this failing?

PR #330, opened 2020-02-04
fixing active and reactive power access and update refs

claytonpbarrows:


src/devices_models/devices/electric_loads.jl
```diff
@@ -248,7 +249,7 @@ function nodal_expression!(
         include_parameters(
             psi_container,
             ts_data_reactive,
-            UpdateRef{L}(REACTIVE_POWER, "get_maxactivepower"),
+            UpdateRef{L}(REACTIVE_POWER, "get_maxreactivepower"),
```
jd-lara: we need to update the conditional here for both active and reactive power (this is the example of reactive. 

Case 1: Parameters and Forecast `UpdateRef{L}(REACTIVE_POWER, "get_maxreactivepower")`
Case 2: Parameters and No - Forecast `UpdateRef{L}(REACTIVE_POWER, "get_reactivepower")`

src/devices_models/devices/hydro_generation.jl
```diff
@@ -185,20 +185,23 @@ function _get_time_series(
...
         name = PSY.get_name(device)
         tech = PSY.get_tech(device)
-        # Hydro gens dont't have a power factor field, so the pf calc is commented
+        # Hydro gens dont't have a power factor field, so the pf is calculated
         # pf = sin(acos(PSY.get_powerfactor(PSY.get_tech(device))))
-        active_power =
-            use_forecast_data ? PSY.get_rating(tech) : PSY.get_activepower(device)
-        reactive_power =
-            use_forecast_data ? PSY.get_rating(tech) : PSY.get_reactivepower(device)
+        pf = PSY.get_reactivepower(device) / PSY.get_rating(tech) # TODO: remove if pf gets added to hydro struct
```
jd-lara: we don't have a constant power factor model for Hydro. The reactive power modeling for hydro must follow something similar to thermal. 

The key aspect to consider in Hydro modeling is that we can generally (except when hydro storage and reservoir coordination is present) think of it as renewable in terms of active power and as thermal in terms of reactive power.

src/devices_models/devices/hydro_generation.jl
```diff
@@ -185,20 +185,23 @@ function _get_time_series(
...
+        pf = iszero(pf) ? pf : 1.0 # TODO: remove if pf gets added to hydro struct
+
         if use_forecast_data
+            active_power = PSY.get_rating(tech)
+            reactive_power = PSY.get_rating(tech) * pf
```
jd-lara: ditto here

src/devices_models/devices/hydro_generation.jl
```diff
@@ -545,13 +547,13 @@ function nodal_expression!(
         include_parameters(
             psi_container,
             ts_data_active,
-            UpdateRef{H}(ACTIVE_POWER, "get_rating"),  # TODO: fix in PR #316
+            UpdateRef{H}(ACTIVE_POWER, "get_activepower"),
```
jd-lara: Same logic as before 

Case 1: Parameters and Forecast `UpdateRef{H}(ACTIVE_POWER, "get_maxactivepower")`
Case 2: Parameters and No - Forecast `UpdateRef{H}(ACTIVE_POWER, "get_activepower")`

src/devices_models/devices/hydro_generation.jl
```diff
@@ -545,13 +547,13 @@ function nodal_expression!(
...
             :nodal_balance_active,
         )
         include_parameters(
             psi_container,
             ts_data_reactive,
-            UpdateRef{H}(REACTIVE_POWER, "get_rating"),  # TODO: fix in PR #316
+            UpdateRef{H}(REACTIVE_POWER, "get_reactivepower"),
```
jd-lara: This doesn't apply here see above comment

src/devices_models/devices/renewable_generation.jl
```diff
@@ -203,13 +204,13 @@ function nodal_expression!(
         include_parameters(
             psi_container,
             ts_data_active,
-            UpdateRef{R}(ACTIVE_POWER, "get_rating"),
+            UpdateRef{R}(ACTIVE_POWER, "get_activepower"),
```
jd-lara: undo this change. The full dispatch version is based on the model `P_re <= time_series_parameter*rating`

src/devices_models/devices/electric_loads.jl
claytonpbarrows: what about case3: no paramaters, no forecast?

src/devices_models/devices/hydro_generation.jl
claytonpbarrows: I don't follow

jd-lara: Hydro doesn't have limitations on reactive power based on time-series. There isn't a reason to create a parameter here.

PR #333, opened 2020-02-07
Jd/simulation sequence fix

jd-lara:
I am opening this PR that partially addresses the issue of the sequence because the fix will become a large change to the internals of simulations and I am aiming at making the review process more manageable. 

The current simulation tests will fail and can't be disabled temporarily, but please focus on the uses of the files `test_simulation_sequence.jl` and `test_simulation.jl`

The formatter is also failing but I don't want to run it and add too much noise to the PR. I will run it once approved.


test/test_simulation_sequence.jl
```diff
@@ -0,0 +1,63 @@
+@testset "Simulation Sequence" begin
+    feedforward_chronologies = Dict(
+        ("UC" => "HAUC") => Synchronize(periods = 24),
+        ("HAUC" => "ED") => RecedingHorizon(),
+        ("ED" => "AGC") => RecedingHorizon(),
+    )
+    ini_cond_chronology = InterStage()
```
claytonpbarrows: What is the alternative to InterStage()? Has it been defined yet?

test/test_simulation_sequence.jl
```diff
@@ -0,0 +1,63 @@
...
+    test_sequence = SimulationSequence(
+        order = order,
+        feedforward_chronologies = feedforward_chronologies,
+        step_resolution = Hour(24),
+        horizons = horizons,
+        intervals = intervals,
+        ini_cond_chronology = ini_cond_chronology,
+    )
+
+    @test length(findall(x -> x == 4, test_sequence.execution_order)) == 24 * 60
```
claytonpbarrows: Now that execution_order is a vector of stage numbers, do we still need stage numbers? i.e. could we instead just use the stage names and keep a vector of those? That could simplify the interface.

jd-lara: I prefer to keep it to streamline the construction and running of the stages. Also, this will force the users to explicitly state the order they expect instead of having to infer it from the order provided. We can come back to this once the whole process is finished.

test/test_simulation_sequence.jl
jd-lara: `IntraStage()`

src/core/simulation_sequence.jl
```diff
@@ -1,51 +1,193 @@
...
+        index -= 1
+    end
+
+    index = length(execution_order)
+    stages = sort!(collect(keys(interval_run_counts)))
+    last_stage = stages[end]
+    _fill_stage(execution_order, index, stages[1], interval_run_counts, last_stage)
+end
+
+function _get_execution_order_vector(
```
daniel-thom: Suggest it's worth having a unit test for this function.

src/core/initial_conditions.jl
```diff
@@ -1,3 +1,7 @@
+struct InterStage <: IniCondChronology end
+struct IntraStage <: IniCondChronology end
```
daniel-thom: Naming:  in these cases I think appending Chronology to the structs (InterStageChronology) increases readability.

src/core/feedforward.jl
```diff
@@ -1,7 +1,88 @@
...
+    return
+end
+
+function check_chronology(sim::Simulation, key::Pair, ::Consecutive)
+    from_stage_horizon = sim.sequence.horizons[key.first]
+    from_stage_interval = sim.sequence.intervals[key.first]
+    if from_stage_horizon != from_stage_interval
+        @warn("Consecutive Chronology Requires the same interval and horizon, the parameter horizon = $(from_stage_horizon) in stage $(key.first) will be replaced with $(from_stage_interval). If this is not the desired behviour consider changing your chronology to RecedingHorizon")
+    end
+    sim.sequence.horizons[key.first] = sim.sequence.intervals[key.first]
```
daniel-thom: This line should be moved inside the if block.  Since you're mutating `sim`, the function should be `check_chronology!`.

src/core/simulation.jl
```diff
@@ -198,6 +190,12 @@ function _get_simulation_initial_times!(sim::Simulation)
                 end
             end
         end
+        if !isnothing(sim_ini_time)
+            if isempty([x for x in stage_initial_times[stage_number] if x == sim_ini_time])
```
daniel-thom:
```
if !isnothing(sim_ini_time) && !mapreduce(x -> x == sim_ini_time, *, stage_initial_times[stage_number])
```

src/core/simulation.jl
```diff
@@ -319,25 +320,27 @@ end
 """ # TODO: Add DocString
 function build!(sim::Simulation; kwargs...)
     check_kwargs(kwargs, SIMULATION_BUILD_KWARGS, "build!")
-    _check_sequence(sim)
-    _check_chronologies(sim)
+    if isnothing(sim.sequence) || isempty(sim.stages)
+        throw(ArgumentError("The simulation object requires a valid definition of stages and SimulationSequence"))
+    end
+    _check_forecasts_sequence(sim)
```
daniel-thom: Should the call to PYS.check_forecast_consistency below be moved into this function?

src/core/simulation.jl
```diff
@@ -236,6 +234,7 @@ function _check_steps(
     for (stage_number, stage_name) in sim.sequence.order
         forecast_count = length(stage_initial_times[stage_number])
         stage = get(sim.stages, stage_name, nothing)
+        @assert !isnothing(stage)
```
daniel-thom: Instead of this assert you could just change the line above to `stage = sim.stages[stage_name]`.

src/core/simulation.jl
```diff
@@ -261,12 +260,13 @@ function _build_stages!(sim::Simulation; kwargs...)
         @info("Building Stage $(stage_number)-$(stage_name)")
         horizon = sim.sequence.horizons[stage_name]
         stage = get(sim.stages, stage_name, nothing)
+        @assert !isnothing(stage)
```
daniel-thom: Same comment as above

test/test_simulation_sequence.jl
```diff
@@ -0,0 +1,81 @@
...
+        intervals = Dict("DAUC" => Hour(24)),
+        ini_cond_chronology = InterStage(),
+    )
+
+    @test isa(test_sequence.ini_cond_chronology, IntraStage)
+    @test test_sequence.execution_order == [1]
+
+end
+
+@testset "testing if Horizon and interval result in a discountinous simulation" begin
```
daniel-thom: typo: discontinuous

src/core/simulation_sequence.jl
jd-lara: there are 2 tests already for the output of the function.

src/core/simulation_sequence.jl
```diff
@@ -1,51 +1,193 @@
...
+            end
+        end
+        execution_order[index] = stage
+        index -= 1
+    end
+
+    index = length(execution_order)
+    stages = sort!(collect(keys(interval_run_counts)))
+    last_stage = stages[end]
+    _fill_stage(execution_order, index, stages[1], interval_run_counts, last_stage)
```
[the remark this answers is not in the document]
jd-lara: only 2 variables don't change `interval_run_counts` and `last_stage`

src/core/simulation.jl
jd-lara: That's already called when making the PSI container. We would en up calling it twice.

src/core/simulation_sequence.jl
```diff
@@ -1,51 +1,193 @@
...
+)
+    reverse_order = sort(collect(keys(order)), rev = true)
+    interval_run_counts = Dict{Int64,Int64}()
+    for k in reverse_order[1:end-1]
+        stage_name = order[k]
+        previous_stage_name = order[k-1]
+        stage_interval = intervals[stage_name]
+        previous_stage_interval = intervals[previous_stage_name]
+        try
+            interval_run_counts[k] = previous_stage_interval / stage_interval
```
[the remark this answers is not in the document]
jd-lara: Intervals are in TimePeriod units. In this way, we guarantee always getting the error. I am not sure the behavior we should expect from the % operator with Periods.

test/test_simulation_sequence.jl
```diff
@@ -0,0 +1,81 @@
...
+            ("ED", :devices, :Generators) => SemiContinuousFF(
+                binary_from_stage = PSI.ON,
+                affected_variables = [PSI.ACTIVE_POWER],
+            ),
+        ),
+        cache = Dict("ED" => [TimeStatusChange(PSY.ThermalStandard, PSI.ON)]),
+        ini_cond_chronology = InterStageChronology(),
+    )
+
+    test_sequence = SimulationSequence(
```
claytonpbarrows: I don't see a test that actually builds this sequence.

src/core/simulation_sequence.jl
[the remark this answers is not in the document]
daniel-thom: The reference to execution_order also doesn't change.

src/core/simulation_sequence.jl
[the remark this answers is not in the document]
daniel-thom: I'm not sure that your code is correct.  Here are some example with mod.

```
julia> Minute(10) % Minute(2)
0 minutes

julia> Minute(10) % Minute(3)
1 minute
```

src/core/simulation.jl
```diff
@@ -105,33 +105,25 @@ function _check_sequence(sim::Simulation)
...
-        to_stage_interval = sim.sequence.intervals[key.second]
-        check_chronology(
-            chron,
-            (from_stage => to_stage),
-            (from_stage_horizon => to_stage_horizon),
-            (from_stage_interval => to_stage_interval),
-        )
+function _check_feedforward_chronologies(sim::Simulation)
+    if isempty(sim.sequence.feedforward_chronologies)
+        @info("No Feedforward Chronologies defined")
```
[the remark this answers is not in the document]
jd-lara: Those cases are already checked for problematic cases in here

https://github.com/NREL/PowerSimulations.jl/blob/689b3ada71ba40312aa72939af00cc348df96533/src/core/simulation_sequence.jl#L99-L108

test/test_simulation_sequence.jl
jd-lara: Simulation sequence tests only test the simulation sequence object. I will add the single-stage construction test.

src/core/simulation_sequence.jl
```diff
@@ -1,51 +1,193 @@
...
         order::Dict{Int64,String},
-        intra_stage_chronologies = Dict{Pair{String,String},AbstractChronology}(),
-        feed_forward = Dict{Tuple{String,Symbol,Symbol},AbstractAffectFeedForward}(),
-        ini_cond_chronology = Dict{String,AbstractChronology}(),
+        feedforward_chronologies = Dict{Pair{String,String},FeedForwardChronology}(),
+        feedforward = Dict{Tuple{String,Symbol,Symbol},AbstractAffectFeedForward}(),
+        ini_cond_chronology = InterStage(),
         cache = Dict{String,Vector{AbstractCache}}(),
     )
+        _check_stage_order(order)
```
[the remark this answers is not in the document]
jd-lara: Already addressed here 
https://github.com/NREL/PowerSimulations.jl/blob/689b3ada71ba40312aa72939af00cc348df96533/src/core/simulation_sequence.jl#L91-L97

PR #341, opened 2020-02-12
add template functions

claytonpbarrows:
creates functions to return `OperationsProblemTemplate`s for standard formulations for:
 - Unit Commitment
 - Economic Dispatch


jd-lara:
The package needs tests

PR #343, opened 2020-02-13
add multiple result comparison for plots

Lilyhanig:


PowerSimulationsPlots/src/plot_recipes.jl
```diff
@@ -3,6 +3,21 @@ import PlotlyJS
 import WebIO
 WebIO.install_jupyter_nbextension()
 
+function _format_multiple_plots(length::Int, plots::Array)
```
daniel-thom: Why is this function necessary?  Does `plots` have extra entries that you are truncating?  If so, could you write
```plots = plots[1:length]```

File: PowerSimulationsPlots/src/call_plots.jl [hunk 2]
```diff
@@ -78,6 +78,26 @@ PLOTLY_DEFAULT = vcat(
     CURTAILMENT_288,
 )
 
+function _filter_variables(results::PSI.Results; kwargs...)
+    filter_results = Dict()
+    reserves = get(kwargs, :reserves, false)
+    if reserves
+        for (key, var) in results.variables
+            if "$key"[1:2] == "P_" || "$key"[1:5] == "Spin_" || "$key"[1:4] == "Reg_" || "$key"[1:5] == "Flex_"
```
daniel-thom: This feels brittle. Are these strings coming from PSI const definitions for variables?  Seems like you could split by underscore and then compare the values.  And that could be made part of PSI.

PowerSimulationsPlots/src/plot_recipes.jl
Lilyhanig: couldn't figure this out for now....

jd-lara: Add a comment on why the hacky workaround.

PR #344, opened 2020-02-13
simulation solution refactor

jd-lara:
All the capabilities are back in place, and the tests are passing. This PR is reorganizing certain aspects of the initial conditions calculations and information passing. Now, based on the initial conditions created, there are checks that the appropriate caches are present. 

It also implements different initial conditions rule updating functions according to the InitialConditionChronology definition. This required adding a chronology to the interval to know if the initial conditions update from the same stage need to be done in a consecutive or receding horizon way. This syntax also enables creating problems that run single-stage multiple times. 

This PR also includes the use of a simulation-wide cache and other code errors. This new design resolved issues of choosing when a cache is defined over parameters or variables.


test/test_simulation_build.jl
```diff
@@ -13,7 +13,10 @@ function test_sequence_build(file_path::String)
         order = Dict(1 => "UC", 2 => "ED"),
         feedforward_chronologies = Dict(("UC" => "ED") => Synchronize(periods = 24)),
         horizons = Dict("UC" => 24, "ED" => 12),
-        intervals = Dict("UC" => Hour(24), "ED" => Hour(1)),
+        intervals = Dict(
```
jd-lara: @claytonpbarrows see here that now the intervals need to define a chronology too. This allows us to know how information passes between runs of the same stage.

src/core/simulation.jl
```diff
@@ -225,20 +231,57 @@ function _check_steps(
...
+    for (stage_number, stage_name) in sim.sequence.order
+        receiving_stage = get_stage(sim, stage_name)
+        for (k, v) in get_initial_conditions(receiving_stage.internal.psi_container)
+            # No cache needed for the initial condition -> continue
+            isnothing(v[1].cache_type) && continue
+            c = nothing
+            # Search other stages
+            for source_stage in values(sim.stages)
+                c = get_cache(source_stage, v[1].cache_type)
+                break
```
daniel-thom: Looping one time and then breaking feels odd.  And it doesn't match the comment.  If this is what you want then you could write `source_stage = iterate(values(sim.stages))[1]`.  Then you could delete `c = nothing`.

src/core/simulation.jl
```diff
@@ -333,3 +376,58 @@ function build!(sim::Simulation; kwargs...)
...
+# or custom InitialConditionType #
+
+""" Updates the initial conditions of the stage"""
+function initial_condition_update!(
+    stage::Stage,
+    ini_cond_key::ICKey,
+    chronology::IntraStageChronology,
+    sim::Simulation,
+)
+    ini_cond_vector = get_initial_conditions(stage.internal.psi_container)[ini_cond_key]
```
daniel-thom: nit: Don't use vector in variable names.
```
ini_conds = get_initial_conditions(stage.internal.psi_container)[ini_cond_key]
```
or
```
for ic in get_initial_conditions(stage.internal.psi_container)[ini_cond_key]
```

test/test_simulation_build.jl
```diff
@@ -216,6 +234,65 @@ function test_sequence_build(file_path::String)
         stage_initial_times = PSI._get_simulation_initial_times!(sim)
         @test_throws IS.ConflictingInputsError PSI._check_steps(sim, stage_initial_times)
     end
+
+    @testset "Innapropiate cache definition" begin
```
daniel-thom: Inappropriate

src/core/simulation.jl
jd-lara: if c keeps being nothing because the `get_cache` functions return nothing it means that it is not defined. I don't think the proposed change achieves that.

jd-lara: fixed in 2c30f4a.

test/test_simulation_build.jl
claytonpbarrows: @jd-lara does an interval with Consecutive() chronology always get initial conditions from the last period of the previous execution? What if your interval and horizon arent equal?

jd-lara: This needs correction to make it grab the results from the end of the interval.

test/test_simulation_build.jl
```diff
@@ -216,6 +234,65 @@ function test_sequence_build(file_path::String)
         stage_initial_times = PSI._get_simulation_initial_times!(sim)
         @test_throws IS.ConflictingInputsError PSI._check_steps(sim, stage_initial_times)
     end
+
+    @testset "Inapropiate cache definition" begin
```
daniel-thom: Still spelled wrong.  Inappropriate

src/core/simulation.jl
```diff
@@ -225,23 +231,60 @@ function _check_steps(
...
+            # No cache needed for the initial condition -> continue
+            isnothing(v[1].cache_type) && continue
+            c = nothing
+            # Search other stages
+            for source_stage in values(sim.stages)
+                c = get_cache(source_stage, v[1].cache_type)
+                !isnothing(c) && break
+            end
+            if isnothing(c)
+                throw(IS.ArgumentError("Cache $(v[1].cache_type) not defined for initial condition $(k.ic_type) in stage $receiving_stage "))
```
daniel-thom: IS.ArgumentError does not exist.

src/core/simulation.jl
```diff
@@ -225,23 +231,60 @@ function _check_steps(
...
 end
 
+function _check_required_ini_cond_caches(sim::Simulation, ::IntraStageChronology)
+    for (stage_number, stage_name) in sim.sequence.order
+        stage = get_stage(sim, stage_name)
+        for (k, v) in get_initial_conditions(stage.internal.psi_container)
+            isnothing(v[1].cache_type) && continue
+            c = get_cache(stage, v[1].cache_type)
+            if isnothing(c)
+                throw(IS.ArgumentError("Cache $(v[1].cache_type) not defined for initial condition $(k.ic_type) in stage $stage_name"))
```
daniel-thom: Ditto

test/test_simulation_build.jl
jd-lara: solved in c752b37. Now Consecutive() chooses the last step in the interval.

jd-lara:
> sim.sequence.initial_time

Fixed in 74df002. Thanks for noting it

daniel-thom:
In almost all places where you call `get_stage` you don't check whether the return is nothing.  In at least one place you check and then throw an exception.  Whenever I see that I wonder if it could be a problem.  You'd likely be better off putting the throw in the function itself.  I don't see any place where you take some other action at the call site if the return is nothing.

daniel-thom:
The function `_assign_feedforward_chronologies` creates a local function called `find_val` that is unused.  Can you delete it?

src/core/simulation.jl
```diff
@@ -225,23 +242,47 @@ function _check_steps(
...
     isnothing(caches) && return
-    cache_dict = Dict{Type{<:AbstractCache}, AbstractCache}()
+    stage = get_stage(sim, stage_name)
     for c in caches
-        sim.stages[stage_name].internal.cache_dict[typeof(c)] = c
-        build_cache!(c, sim.stages[stage_name].internal.psi_container)
+        cache_key = CacheKey(c)
+        push!(stage.internal.caches, cache_key)
+        if !haskey(sim.internal.simulation_cache, cache_key)
+            @debug "Cache $(cache_key) added to he simulation"
```
daniel-thom: typo:  `the simulation`

src/core/simulation_stages.jl
```diff
@@ -3,20 +3,22 @@ mutable struct StageInternal
...
-    synchronized_executions::Dict{Int, Int} # Number of executions per upper level stage step
+    end_of_interval_step::Int
+    # This line keeps track of the executions of a stage relative to other stages.
+    # This might be needed in the future to run multiple stages. For now it is disabled
+    #synchronized_executions::Dict{Int, Int} # Number of executions per upper level stage step
     psi_container::Union{Nothing, PSIContainer}
-    cache_dict::Dict{Type{<:AbstractCache}, AbstractCache}
-    # Can probably be eliminated and use getter functions from
-    # Simulation object. Need to determine if its always available in the stage update steps.
+    caches::Set{<:CacheKey{<:AbstractCache, <:PSY.Device}}
```
daniel-thom: Can you add a comment on `caches`, particularly explaining why it is a Set?

src/core/simulation_sequence.jl
```diff
@@ -132,50 +133,57 @@ end
 mutable struct SimulationSequence
     horizons::Dict{String, Int}
     step_resolution::Dates.TimePeriod
-    intervals::Dict{String, <:Dates.TimePeriod}
+    intervals::Dict{String, Tuple{<:Dates.TimePeriod, <:FeedForwardChronology}}
     order::Dict{Int, String}
     feedforward_chronologies::Dict{Pair{String, String}, <:FeedForwardChronology}
     feedforward::Dict{Tuple{String, Symbol, Symbol}, <:AbstractAffectFeedForward}
     ini_cond_chronology::IniCondChronology
     cache::Dict{String, Vector{<:AbstractCache}}
```
daniel-thom: Can you add a comment that this string is stage name?

src/core/simulation.jl
```diff
@@ -225,23 +242,47 @@ function _check_steps(
...
+    caches = get_stage_cache_definition(sim, stage_name)
     isnothing(caches) && return
-    cache_dict = Dict{Type{<:AbstractCache}, AbstractCache}()
+    stage = get_stage(sim, stage_name)
     for c in caches
-        sim.stages[stage_name].internal.cache_dict[typeof(c)] = c
-        build_cache!(c, sim.stages[stage_name].internal.psi_container)
+        cache_key = CacheKey(c)
+        push!(stage.internal.caches, cache_key)
+        if !haskey(sim.internal.simulation_cache, cache_key)
```
daniel-thom: This code block should probably be a function in SimulationInternal.
`function update_cache_value(internal::SimulationInternal, c, value)`

That would keep the implementation details of SimulationInternal within itself.

src/core/simulation.jl
```diff
@@ -74,21 +76,35 @@ get_sequence(s::Simulation) = s.sequence
...
-get_name(s::Simulation, stage::Stage) = get(s.sequence.order, get_number(stage), nothing)
+get_ini_cond_chronology(s::Simulation) = s.sequence.ini_cond_chronology
+get_stage_name(s::Simulation, stage::Stage) = get_stage_name(s.sequence, stage)
+get_name(s::Simulation) = s.name
+get_simulation_folder(s::Simulation) = s.simulation_folder
+get_execution_order(s::Simulation) = s.sequence.execution_order
+get_current_execution_index(s::Simulation) = s.sequence.current_execution_index
+get_stage_cache_definition(s::Simulation, stage::String) =
+    get(s.sequence.cache, stage, nothing)
+get_cache(s::Simulation, key::CacheKey) = get(s.internal.simulation_cache, key, nothing)
```
daniel-thom: Would it be better to throw an exception here if key is not stored?  It depends on whether the calling code will take some other action in that case.

src/routines/stage_update.jl
```diff
@@ -42,74 +42,37 @@ function parameter_update!(
...
+    _update_initial_conditions!(stage, sim)
     return
 end
 
+#############################Interfacing Functions##########################################
+## These are the functions that the user will have to implement to update a custom stage ###
+""" Generic Stage update function for most problems with no customization"""
 function update_stage!(
     stage::Stage{M},
     step::Int,
```
daniel-thom: Does this function still need `step`?

src/core/simulation.jl
```diff
@@ -333,3 +364,161 @@ function build!(sim::Simulation; kwargs...)
...
+function _update_caches!(sim::Simulation, stage::Stage)
+    for cache in stage.internal.caches
+        update_cache!(sim, cache, stage)
+    end
+    return
+end
+
+################################Cache Update################################################
+# TODO: Need to be careful here if 2 stages modify the same cache. This function might need
+# dispatch on the Statge{OpModel} to assing different actions. e.g. HAUC and DAUC
```
daniel-thom: typo:  assign

src/core/simulation.jl
```diff
@@ -333,3 +364,161 @@ function build!(sim::Simulation; kwargs...)
...
+    for t in 1:get_end_of_interval_step(stage), name in variable.axes[1]
+        device_status = JuMP.value(variable[name, t])
+        if c.value[name][:status] == device_status
+            c.value[name][:count] += 1.0
+        elseif c.value[name][:status] != device_status
+            c.value[name][:count] = 1.0
+            c.value[name][:status] = device_status
+        end
+    end
+
```
daniel-thom: It would probably be a good idea to log or record these actions.

src/core/simulation_stages.jl
jd-lara: I made it a set because I don't need any order and I need to make sure that there is only one of each.

src/core/simulation.jl
jd-lara: I look into the uses of `get_cache` and check if I should throw an error when not present.

src/core/simulation.jl
jd-lara: I will add a debug statement and we should implement a better logging solution later.

src/core/simulation_sequence.jl
```diff
@@ -132,50 +133,57 @@ end
...
     execution_order::Vector{Int}
+    current_execution_index::Int64
 
     function SimulationSequence(;
         horizons::Dict{String, Int},
         step_resolution::Dates.TimePeriod,
-        intervals::Dict{String, <:Dates.TimePeriod},
+        intervals::Dict{String, <:Tuple{<:Dates.TimePeriod, <:FeedForwardChronology}},
```
claytonpbarrows: So, can the `intervalas[key][2]` be any `FeedForwardChrolongy`? I don't understand what the implications of a `Synchronize()` entry would be...

src/core/simulation_sequence.jl
```diff
@@ -132,50 +133,57 @@ end
...
         feedforward = Dict{Tuple{String, Symbol, Symbol}, AbstractAffectFeedForward}(),
         ini_cond_chronology = InterStageChronology(),
         cache = Dict{String, Vector{AbstractCache}}(),
     )
         _check_stage_order(order)
+        _intervals = Dict{String, Tuple{<:Dates.TimePeriod, <:FeedForwardChronology}}()
+        for (k, v) in intervals
+            _intervals[k] = (IS.time_period_conversion(intervals[k][1]), intervals[k][2])
+        end
+        step_resolution = IS.time_period_conversion(step_resolution)
```
claytonpbarrows: Given the new, more explicit, definition of the sequence and order, do we still need the `step_resolution` argument? Can it be inferred from the interval of the first stage?

src/core/simulation_sequence.jl
jd-lara: Synchronize implies that the next execution of the same stage uses information from the last step in the interval to update its data.

src/core/simulation_sequence.jl
jd-lara: No, we can't get rid of step resolution if the first stage is supposed to run more than once per step.

PR #349, opened 2020-02-19
add glossary

Lilyhanig:
- this could potentially be done differently I think, with a mustache and json file like the docstrings in PowerSystems/InfrastructureSystems.... maybe more work than it's worth.

### DIFF
--- a/src/core/glossary.md
+++ b/src/core/glossary.md
@@ -1,0 +1,29 @@
+# Glossary
+
+## Simulation Sequence Components
+
+**cache:** Dictionary of each stage and its vector of caches. Each cache tracks the last time status of a device changed in a simulation.
+
+**feed_forward:** The vertical inter-stage relationship between the binary variable status of one stage and the affected variables of the next stage.
+- Example: `feedforward = Dict(("stage-2", :devices, :Generators) => SemiContinuousFF(binary_from_stage = PSI.ON, affected_variables = [PSI.ACTIVE_POWER])`
+
+**feed_forward_chronologies:** The vertical inter-stage relationship dictating how variable results impact the next stage's variable parameters.
+- Example: `feedforward_chronologies = Dict(("stage-1" => "stage-2") => Synchronize(periods = 24))`
+
+**horizons:** The integer count of resolution time periods for a full step resolution of the simulation. *(Horizon = 12) x (resolution = 1 Hour) = 12 Hours*
+- Example: `horizons = Dict("stage-1" => 24, "stage-2" => 12)`
+
+**ini_cond_chronology:** The structure dictating how initial conditions get updated from previous results in the simulation.
+- Example: `ini_cond_chronology = InterStageChronology()`
+
+**intervals:** The increment of time per stage in each simulation solve, and how results get fed forward into initial conditions between intervals.
+- Example: `intervals = Dict("UC" => (Hour(24), Consecutive()), "ED" => (Hour(1), Consecutive()))`
+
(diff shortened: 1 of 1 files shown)


src/core/glossary.md
```diff
@@ -0,0 +1,29 @@
+# Glossary
+
+## Simulation Sequence Components
+
+**cache:** Dictionary of each stage and its vector of caches. Each cache tracks the last time status of a device changed in a simulation.
```
jd-lara: Don't mention the type of object, just the definition for instance in this case: 

`Cache is the storage of the solution of a variable from a stage to be used by other stages in the simulation`

src/core/glossary.md
```diff
@@ -0,0 +1,29 @@
...
+
+**feed_forward:** The vertical inter-stage relationship between the binary variable status of one stage and the affected variables of the next stage.
+- Example: `feedforward = Dict(("stage-2", :devices, :Generators) => SemiContinuousFF(binary_from_stage = PSI.ON, affected_variables = [PSI.ACTIVE_POWER])`
+
+**feed_forward_chronologies:** The vertical inter-stage relationship dictating how variable results impact the next stage's variable parameters.
```
jd-lara: mention that this creates a time relationship.

src/core/glossary.md
```diff
@@ -0,0 +1,29 @@
+# Glossary
+
+## Simulation Sequence Components
+
+**cache:** Dictionary of each stage and its vector of caches. Each cache tracks the last time status of a device changed in a simulation.
+
+**feed_forward:** The vertical inter-stage relationship between the binary variable status of one stage and the affected variables of the next stage.
```
jd-lara: Not a relationship, it is the variable that is used as a parameter for a later stage.

src/core/glossary.md
```diff
@@ -0,0 +1,29 @@
...
+- Example: `feedforward = Dict(("stage-2", :devices, :Generators) => SemiContinuousFF(binary_from_stage = PSI.ON, affected_variables = [PSI.ACTIVE_POWER])`
+
+**feed_forward_chronologies:** The vertical inter-stage relationship dictating how variable results impact the next stage's variable parameters.
+- Example: `feedforward_chronologies = Dict(("stage-1" => "stage-2") => Synchronize(periods = 24))`
+
+**horizons:** The integer count of resolution time periods for a full step resolution of the simulation. *(Horizon = 12) x (resolution = 1 Hour) = 12 Hours*
```
jd-lara: Horizon

src/core/glossary.md
```diff
@@ -0,0 +1,29 @@
...
+- Example: `horizons = Dict("stage-1" => 24, "stage-2" => 12)`
+
+**ini_cond_chronology:** The structure dictating how initial conditions get updated from previous results in the simulation.
```
jd-lara: don't use the input to the simulation object. use just the words. 

Initial Condition Chronology

src/core/glossary.md
```diff
@@ -0,0 +1,29 @@
...
+- Example: `ini_cond_chronology = InterStageChronology()`
+
+**intervals:** The increment of time per stage in each simulation solve, and how results get fed forward into initial conditions between intervals.
```
jd-lara: Interval*

src/core/glossary.md
daniel-thom: Jose recently changed all uses to `feedforward`

PR #355, opened 2020-02-20
2 19 docstrings

Lilyhanig:


jd-lara:
Does this depend on the newer IS?

PR #356, opened 2020-02-21
2 21 serialize

Lilyhanig:


src/utils/simulation_results.jl
```diff
@@ -98,26 +127,47 @@ function load_simulation_results(
...
+- `stage_number::String`: The stage of the results getting parsed
 
+# Example
+```julia
+sim_output = execute!(simulation)
+results = load_simulation_results(sim_output, "stage_name")
+```
+# Accepted Key Words
+- `write::Bool`: if true, the aggregated results get written back to the results file in the folder structure
+- `file_type::File Type = Feather`: default is feather file for writing the aggregated results back to the folder.
```
daniel-thom: Why would someone want to do this?

Lilyhanig: @daniel-thom the simulation writes the results to a complicated folder structure, someone definitely might want to write a chunk of aggregated results back to the "results" folder.

jd-lara: we have a distinction between results and raw_output. Results are the post-processed output according to the user's specification.

PR #362, opened 2020-02-28
Jd/add params to results

jd-lara:


src/core/operations_problem.jl
```diff
@@ -470,6 +468,41 @@ function _build!(
...
+function get_dual_values(op_m::OperationsProblem; kwargs...)
+    cons = get(kwargs, :constraints_duals, nothing)
+    isnothing(cons) && return Dict{Symbol, Any}()
+    results_dict = Dict{Symbol, DataFrames.DataFrame}()
+    for c in cons
+        v = get_constraint(op_m.psi_container, c)
+        results_dict[c] = axis_array_to_dataframe(v)
+    end
+    return results_dict
+end
```
daniel-thom: This returns different types. Maybe the following is more readable?

```
    cons = get(kwargs, :constraints_duals, [])
    return Dict(c => axis_array_to_dataframe(get_constraint(op_m.psi_container, c)) for c in cons)
```

src/core/operations_problem.jl
```diff
@@ -470,6 +468,41 @@ function _build!(
...
+
+function get_parameters_value(op_m::OperationsProblem)
+    # TODO: Still not obvious implementation since it needs to get the multipliers from
+    # the system
+    params_dict = Dict{Symbol, DataFrames.DataFrame}()
+    parameters = get_parameters(op_m.psi_container)
+    isnothing(parameters) && return params_dict
+    isempty(parameters) && return params_dict
```
daniel-thom:
```suggestion
    isnothing(parameters) || isempty(parameters) && return params_dict
```

src/core/psi_container.jl
```diff
@@ -290,6 +288,10 @@ function _encode_for_jump(name::Symbol)
     return name
 end
 
+function decode_symbol(name::Symbol)
```
daniel-thom: The encode and decode functions should be symmetric.  We should change `_encode_for_jump` to `encode_symbol`.  And maybe `_JUMP_NAME_DELIMITER` needs a better name.

src/core/operations_problem.jl
```diff
@@ -470,6 +468,41 @@ function _build!(
...
+    for (k, v) in parameters
+        !isa(v.update_ref, UpdateRef{<:PSY.Component}) && continue
+        params_key_tuple = decode_symbol(k)
+        params_dict_key = Symbol(params_key_tuple[1], "_", params_key_tuple[3])
```
daniel-thom: Shouldn't this use the existing delimiter const?

src/core/simulation_results.jl
```diff
@@ -215,8 +303,8 @@ function load_simulation_results(
     obj_value = Dict{Symbol, Any}(:OBJECTIVE_FUNCTION => optimizer["obj_value"])
     if !isempty(dual)
         duals = _read_references(duals, dual, stage, step, references, time_length)
-        results =
-            DualResults(variables, obj_value, optimizer, time_stamp, duals, results_folder)
+        results = 0.0
```
daniel-thom: Can you insert a comment on why this is disabled?
Ditto, below.

src/core/operations_problem.jl
jd-lara: @Lilyhanig consider this in your PR to this branch, please.

src/core/operations_problem.jl
jd-lara: not in this case since it it for the printing of results

src/core/psi_container.jl
```diff
@@ -499,14 +500,23 @@ function iterate_parameter_containers(psi_container::PSIContainer)
...
+        !isa(v.update_ref, UpdateRef{<:PSY.Component}) && continue
+        params_key_tuple = decode_symbol(k)
+        params_dict_key = Symbol(params_key_tuple[1], "_", params_key_tuple[3])
+        params_dict[params_dict_key] = axis_array_to_dataframe(get_parameter_array(v))
     end
-    return results_dict
+    return params_dict
+end
+
+function is_milp(container::PSIContainer)
```
claytonpbarrows: It looks like this is used to check if there are duals available. Should we generalize to any non LP?

jd-lara: There is still value in the duals of constraints of NLP's. The issue is that this broke a lot of things in the export of results before we were exporting .feather files with a bunch of invalid values.

src/core/simulation_results.jl
```diff
@@ -241,13 +241,13 @@ for the desired step range and variables
 - `SimulationResultsReference::SimulationResultsReference`: the container for the reference dictionary created in execute!
 - `stage_number::Int = 1``: The stage of the results getting parsed: 1 or 2
 - `step::Array{String} = ["step-1", "step-2", "step-3"]`: the steps of the results getting parsed
-- `variable::Array{Symbol} = [:P_ThermalStandard, :P_RenewableDispatch]`: the variables to be parsed
+- `variable::Array{Symbol} = [:P__ThermalStandard, :P_RenewableDispatch]`: the variables to be parsed
```
daniel-thom: Does this need to be `P__RenewableDispatch`?

test/test_operations_solve.jl
```diff
@@ -233,12 +233,12 @@ res = solve_op_problem!(op_problem; constraints_duals = duals)
     end
 end
 
-path = joinpath(pwd(), "test_writing")
+path = joinpath(pwd(), "Test_writing")
```
daniel-thom: Was this change intentional?  It's kinda weird to mix capital letters and underscores.

src/core/operations_problem.jl
```diff
@@ -470,6 +468,30 @@ function _build!(
...
+    return results_dict
+end
+
+function get_dual_values(op_m::OperationsProblem; kwargs...)
+    cons = get(kwargs, :constraints_duals, [])
+    results_dict = Dict{Symbol, DataFrames.DataFrame}()
+    for c in cons
+        v = get_constraint(op_m.psi_container, c)
+        results_dict[c] = axis_array_to_dataframe(v)
+    end
```
daniel-thom: Did you forget to delete this?

src/core/psi_container.jl
```diff
@@ -537,3 +547,33 @@ function _write_psi_container(psi_container::PSIContainer, save_path::String)
...
 end
+
+function get_model_duals(op::PSIContainer, cons::Vector{Symbol})
+    results_dict = Dict{Symbol, DataFrames.DataFrame}()
+    for c in cons
+        v = get_constraint(op, c)
+        results_dict[c] = axis_array_to_dataframe(v)
+    end
+    return results_dict
+end
```
daniel-thom: This is redundant with `get_dual_values`.

PR #364, opened 2020-03-02
2 27 basepower

Lilyhanig:


src/core/operations_problem_results.jl
```diff
@@ -79,6 +79,8 @@ function load_operation_results(folder_path::AbstractString)
     dual_result = Dict{Symbol, Any}()
     dual_names = find_duals(variable_list)
     param_names = find_params(variable_list)
+    variable_list = setdiff(variable_list, dual_names)
```
jd-lara: Why 2 separate set_diff calls?

src/core/operations_problem_results.jl
```diff
@@ -157,3 +159,13 @@ end
 function write_to_CSV(results::OperationsProblemResults, folder_path::String)
     write_results(results, folder_path; file_type = CSV)
 end
+
+function _find_params(variables::Array)
```
jd-lara: I think this function is replicated from https://github.com/NREL/PowerSimulations.jl/blob/e2cd5f2df16799ed89496b4194cc9cead0716991/src/core/operations_problem_results.jl#L32

src/core/psi_container.jl
```diff
@@ -463,6 +463,16 @@ function get_parameter_array(psi_container::PSIContainer, ref)
     return get_parameter_array(get_parameter_container(psi_container, ref))
 end
 
+function get_model_duals(op::PSIContainer, cons::Vector{Symbol})
```
jd-lara: Also, duplicate of https://github.com/NREL/PowerSimulations.jl/blob/e2cd5f2df16799ed89496b4194cc9cead0716991/src/core/psi_container.jl#L550

src/core/simulation_results.jl
```diff
@@ -377,6 +381,7 @@ function load_simulation_results(
     dual = find_duals(variable)
     param = find_params(collect(keys(references[stage])))
     variable = setdiff(variable, dual)
+    variable = setdiff(variable, param)
```
jd-lara: Why 2 setdiff calls?

PR #365, opened 2020-03-03
improve code coverage

Lilyhanig:


src/core/operations_problem_results.jl
```diff
@@ -13,22 +13,6 @@ get_cost(result::OperationsProblemResults) = result.total_cost
...
 
-function get_variable(res_model::OperationsProblemResults, key::Symbol)
-    var_result = get(res_model.variable_values, key, nothing)
-    if isnothing(var_result)
-        throw(ArgumentError("No variable with key $(key) has been found."))
-    end
-    return var_result
-end
-
-function get_optimizer_log(results::OperationsProblemResults)
```
jd-lara: same here

src/core/operations_problem_results.jl
```diff
@@ -13,22 +13,6 @@ get_cost(result::OperationsProblemResults) = result.total_cost
 get_time_stamp(result::OperationsProblemResults) = result.time_stamp
 get_duals(result::OperationsProblemResults) = result.dual_values
 
-function get_variable(res_model::OperationsProblemResults, key::Symbol)
```
jd-lara: we should keep this function and add a test for it.

src/core/operations_problem_results.jl
```diff
@@ -13,22 +13,6 @@ get_cost(result::OperationsProblemResults) = result.total_cost
...
-        throw(ArgumentError("No variable with key $(key) has been found."))
-    end
-    return var_result
-end
-
-function get_optimizer_log(results::OperationsProblemResults)
-    return results.optimizer_log
-end
-
-function get_time_stamps(results::OperationsProblemResults, key::Symbol)
```
jd-lara: same here

PR #375, opened 2020-03-09
begin to change parameter struct

Lilyhanig:


src/core/feedforward.jl
```diff
@@ -203,7 +203,8 @@ function ub_ff(
     set_name = axes[1]
 
     @assert axes[2] == time_steps
-    param_ub = add_param_container!(psi_container, param_reference, set_name)
+    param_ub =
```
jd-lara: split this in two lines for clarity

PR #376, opened 2020-03-09
Results interface

claytonpbarrows:
This PR exports the interface to IS.Results objects.


src/core/results_interface.jl
```diff
@@ -0,0 +1,20 @@
...
+    return IS.get_optimizer_log(result)
+end
+
+function get_time_stamp(result::T) where {T <: IS.Results}
+    return IS.get_time_stamp(result)
+end
+
+function write_results(results::T, save_path::String; kwargs...) where {T <: IS.Results}
+    return IS.write_results(results, save_path; kwargs...)
+end
```
daniel-thom: Since these are identical signatures as what's in IS, you shouldn't need them.  Instead, you should be able to use this pattern:

```
import InfrastructureSystems: get_variables, get_total_cost, get_optimizer_log, get_time_stamp, write_results
export get_variables
export get_total_cost
export get_optimizer_log
export get_time_stamp
export write_results
```

docs/Project.toml
```diff
@@ -1,6 +1,7 @@
 [deps]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 DocumenterTools = "35a29f4d-8980-5a13-9543-d66fff28ecb8"
+PowerSimulations = "e690365d-45e2-57bb-ac84-44ba829e73c4"
```
daniel-thom: Should this have gotten added?

test/Project.toml
```diff
@@ -15,6 +15,7 @@ OSQP = "ab2f91bb-94b4-55e3-9ba0-7f65df51de79"
 ParameterJuMP = "774612a8-9878-5177-865a-ca53ae2495f9"
 Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
 PowerModels = "c36e90e8-916a-50a6-bd94-075b64ef4655"
+PowerSimulations = "e690365d-45e2-57bb-ac84-44ba829e73c4"
```
daniel-thom: Shouldn't be here.

PR #370, opened 2020-03-05
Adding Energy Storage Cache

sourabhdalvi:
Changes :
- [x] New cache type
- [x] Implement correct initial condition values for Hydro Energy Reservoir and Battery 
- [ ] Tests for EnergyStored Cache

Merging this PR is blocked by #372

PR #378, opened 2020-03-11
Fixing Cache References

sourabhdalvi:
- Changed cache dict definition in Simulation Sequence https://github.com/NREL/PowerSimulations.jl/blob/b8e2e70843c79f8e8bb30f2adf394740dbe0c7b9/src/core/simulation_sequence.jl#L169
-  This will allow multiple stages to modify information in cache


jd-lara:
> The tuple keys are really ugly. I'm approving this, but this implementation reinforces that we need to provide nice user interfaces so that they don't have to interact with these.

Agreed, we need to come up with a better interface for the cache. For now, it is important to enabling #370

PR #379, opened 2020-03-11
Adding Energy Storage Cache

sourabhdalvi:
Changes :
- [x] New cache type
- [x] Implement correct initial condition values for Hydro Energy Reservoir and Battery 
- [x] Tests for `EnergyStored` Cache
- [x] Test for Cache modified by multiple stages 

This is Copy of PR #370

### DIFF
--- a/src/PowerSimulations.jl
+++ b/src/PowerSimulations.jl
@@ -76,7 +76,7 @@ export DevicePower
 export DeviceStatus
 export TimeDurationON
 export TimeDurationOFF
-export DeviceEnergy
+export EnergyLevel
 
 # cache_models
 export TimeStatusChange
--- a/src/PowerSimulations.jl
+++ b/src/PowerSimulations.jl
@@ -80,6 +80,7 @@ export DeviceEnergy
 
 # cache_models
 export TimeStatusChange
+export EnergyStored
 
 #operation_models
 export GenericOpProblem
--- a/src/PowerSimulations.jl
+++ b/src/PowerSimulations.jl
@@ -80,7 +80,7 @@ export DeviceEnergy
 
 # cache_models
 export TimeStatusChange
-export EnergyStored
+export StoredEnergy
 
 #operation_models
 export GenericOpProblem
--- a/src/core/cache.jl
+++ b/src/core/cache.jl
@@ -21,4 +21,15 @@ function TimeStatusChange(::Type{T}, name::AbstractString) where {T <: PSY.Devic
     return TimeStatusChange(T, value_array, UpdateRef{JuMP.VariableRef}(T, name))
 end
 
+mutable struct EnergyStored <: AbstractCache
+    device_type::Type{<:PSY.Device}
+    value::JuMP.Containers.DenseAxisArray{Float64}
+    ref::UpdateRef
+end
+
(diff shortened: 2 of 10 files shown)


src/PowerSimulations.jl
```diff
@@ -80,6 +80,7 @@ export DeviceEnergy
 
 # cache_models
 export TimeStatusChange
+export EnergyStored
```
jd-lara: switch to StoredEnergy

src/core/cache.jl
```diff
@@ -21,4 +21,15 @@ function TimeStatusChange(::Type{T}, name::AbstractString) where {T <: PSY.Devic
     return TimeStatusChange(T, value_array, UpdateRef{JuMP.VariableRef}(T, name))
 end
 
+mutable struct EnergyStored <: AbstractCache
+    device_type::Type{<:PSY.Device}
+    value::JuMP.Containers.DenseAxisArray{Float64}
+    ref::UpdateRef
+end
+
+function EnergyStored(::Type{T}, name::AbstractString) where {T <: PSY.Device}
```
jd-lara: Same here please

src/core/initial_conditions.jl
```diff
@@ -99,6 +99,21 @@ function calculate_ic_quantity(
     return var_value
 end
 
+function calculate_ic_quantity(
+    initial_condition_key::ICKey{DeviceEnergy, T},
```
jd-lara: EnergyLevel  or energy quantity makes more sense

src/core/simulation.jl
```diff
@@ -425,10 +425,15 @@ function initial_condition_update!(
     ini_cond_vector = get_initial_conditions(stage.internal.psi_container)[ini_cond_key]
     for ic in get_initial_conditions(stage.internal.psi_container)[ini_cond_key]
         name = device_name(ic)
-        interval_chronology = get_stage_interval_chronology(sim, stage.name)
+        interval_chronology =
+            get_stage_interval_chronology(sim.sequence, get_stage_name(sim, stage))
         var_value =
             get_stage_variable(interval_chronology, (stage => stage), name, ic.update_ref)
-        cache = get_cache(stage, ic.cache_key)
+        if isnothing(ic.cache_type)
```
jd-lara: why this here? the get cache should return a nothing if you run the get_cache function.

sourabhdalvi: sorry, about that

sourabhdalvi: Well this function only takes `<: AbstractCache ` but `ic.cache_type` can be a nothing which breaks the that call and this is how you have it for `InterStageChronology`

src/PowerSimulations.jl
sourabhdalvi: will do

src/core/initial_conditions.jl
sourabhdalvi: you want me to change DeviceEnergy to EnergyLevel for  InitialConditionType ?

sourabhdalvi: done

src/core/simulation.jl
jd-lara: ok, just leave it like this. I need to do some refactoring of these updates to make the code cleaner.

PR #381, opened 2020-03-12
Stage kwargs

daniel-thom:


src/core/parameters.jl
```diff
@@ -10,3 +10,11 @@ Base.length(c::ParameterContainer) = length(c.parameter_array)
...
 
 const ParametersContainer = Dict{Symbol, ParameterContainer}
+
+abstract type OperationsProblemParameters end
+
+struct NetworkOperationsParameters <: OperationsProblemParameters
+    ptdf::PSY.PTDF
+end
+
+get_ptdf(p::NetworkOperationsParameters) = p.ptdf
```
daniel-thom: This is a questionable location.  Open to better ideas.

test/test_network_constructors.jl
```diff
@@ -99,8 +103,6 @@ end
     ps_model = OperationsProblem(TestOpProblem, network, c_sys5; optimizer = GLPK_optimizer)
     construct_device!(ps_model, :Thermal, thermal_model)
     construct_device!(ps_model, :Load, load_model)
-    @test_throws ArgumentError construct_network!(ps_model, network)
```
daniel-thom: This negative test check is no longer relevant.  The code will fail to run unless you pass an instance of NetworkOperationsParameters.

PR #386, opened 2020-03-16
Fixing Test bug 

sourabhdalvi:
- The RH test was looking at the 1st index in the ED simulation rather than the last index for initial condition for the next UC


jd-lara:
> * The RH test was looking at the 1st index in the ED simulation rather than the last index for initial condition for the next UC

In a RH model, the first index is used for the initial conditions of the next stage.

sourabhdalvi:
> > * The RH test was looking at the 1st index in the ED simulation rather than the last index for initial condition for the next UC
> 
> In a RH model, the first index is used for the initial conditions of the next stage.

The test runs 1 - UC at 00:00:00 followed by 12 - ED simulations from 00:00:00-00:55:00 and the results from 00:55:00 feed into the 00:01:00 UC 
The test previously was trying to compare results from 00:00:00 ED to initial conditions in 00:01:00 UC which was wrong. Now the test compares results from 00:55:00 ED to initial conditions in 00:01:00 UC. Hope this clears any confusion

PR #388, opened 2020-03-17
Add serialization of OperationsProblem and Simulation

daniel-thom:


src/core/simulation.jl
```diff
@@ -688,3 +691,145 @@ function execute!(sim::Simulation; kwargs...)
...
+they passed to the original Simulation.
+
+# Arguments
+- `directory::AbstractString`: the directory returned from the call to serialize
+- `stage_info::Dict`: Two-level dictionary containing stage parameters that cannot be
+  serialized. The outer dict should be keyed by the stage name. The inner dict must contain
+  'optimizer' and may contain 'jump_model'. These should be the same values used for the
+  original simulation.
+"""
+function Simulation(directory::AbstractString, stage_info::Dict; kwargs...)
```
daniel-thom: Accepting kwargs here is not idea.  I suggest we make a change to the Simulation constructor.
- We could make initial_times an optional parameter to it, defaulting to nothing.
- Remove kwargs from Simulation().
- Remove kwargs this constructor.  initial_times will have been serialized, so we will get it back. There won't be any ambiguity.

test/test_simulation_execute.jl
```diff
@@ -78,190 +88,209 @@ function test_load_simulation(file_path::String)
...
+    for i in 1:2
+        if i == 1
+            build!(sim)
+        else
+            rm(g_test_path, recursive = true)
+            mkdir(g_test_path)
+            sim = get_deserialized(sim, stage_info)
         end
-    end
+        sim_results = execute!(sim; constraints_duals = duals)
```
jd-lara: Execute shouldn't take constraints_duals anymore

src/core/simulation.jl
jd-lara: Ok, we can address this after merging. I will make initial_time an optional parameter.

daniel-thom: Fixed it here.

PR #389, opened 2020-03-18
Improving coverage for Hydro constructors

sourabhdalvi:
- Removed unused constructors
- Added more tests

### DIFF
--- a/src/devices_models/device_constructors/hydrogeneration_constructor.jl
+++ b/src/devices_models/device_constructors/hydrogeneration_constructor.jl
@@ -239,6 +239,8 @@ function construct_device!(
     return
 end
 
+# Currently no Hydro device supports a Unit commiment formulation
+#=
 function construct_device!(
     psi_container::PSIContainer,
     sys::PSY.System,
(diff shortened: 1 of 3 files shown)


src/devices_models/device_constructors/hydrogeneration_constructor.jl
```diff
@@ -21,7 +21,9 @@ function construct_device!(
 
     #Constraints
     activepower_constraints!(psi_container, devices, model, S, get_feedforward(model))
```
jd-lara: This constraints can't be eliminated for this model. Hydro doesn't require constant pf assumptions. For the HydroDispatchFormulations it should add reactive power bounds.

sourabhdalvi: I think @claytonpbarrows  had suggested I do that, I also if we want a reactive power constraints then we need to define a new constraint that works with the data in Hydro, as its different than Thermal if I'm correct

jd-lara: Reactive power in Hydro is the same as in Thermal

PR #390, opened 2020-03-19
Fixing Cache Update

sourabhdalvi:
- Added new field to the TimeStatusChange for establishing a units for the count 
- PSIMA-111

### DIFF
--- a/src/core/cache.jl
+++ b/src/core/cache.jl
@@ -13,12 +13,23 @@ Tracks the last time status of a device changed in a simulation
 mutable struct TimeStatusChange <: AbstractCache
     device_type::Type{<:PSY.Device}
     value::JuMP.Containers.DenseAxisArray{Dict{Symbol, Float64}}
+    units::Dates.TimePeriod
     ref::UpdateRef
+
+    function TimeStatusChange(device_type, value, units, ref)
+        units = IS.time_period_conversion(units)
+        new(device_type, value, units, ref)
+    end
 end
 
 function TimeStatusChange(::Type{T}, name::AbstractString) where {T <: PSY.Device}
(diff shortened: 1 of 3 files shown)


src/core/cache.jl
```diff
@@ -13,12 +13,23 @@ Tracks the last time status of a device changed in a simulation
 mutable struct TimeStatusChange <: AbstractCache
     device_type::Type{<:PSY.Device}
     value::JuMP.Containers.DenseAxisArray{Dict{Symbol, Float64}}
+    units::Dates.TimePeriod
     ref::UpdateRef
+
+    function TimeStatusChange(device_type, value, units, ref)
```
jd-lara: add the types to the inputs and make the defauls 

```suggestion
    function TimeStatusChange(device_type::Type{<:PSY.Device}, value::JuMP.Containers.DenseAxisArray{Dict{Symbol, Float64}}, ref::UpdateRef, units::Dates.TimePeriod = Dates.Hour(1))
```

src/core/cache.jl
```diff
@@ -13,12 +13,23 @@ Tracks the last time status of a device changed in a simulation
...
+        units = IS.time_period_conversion(units)
+        new(device_type, value, units, ref)
+    end
 end
 
 function TimeStatusChange(::Type{T}, name::AbstractString) where {T <: PSY.Device}
     value_array = JuMP.Containers.DenseAxisArray{Dict{Symbol, Float64}}(undef, 1)
-    return TimeStatusChange(T, value_array, UpdateRef{JuMP.VariableRef}(T, name))
+    return TimeStatusChange(
```
jd-lara: This won't be necessary with the previous change

PR #391, opened 2020-03-20
PSIMA-11: initial conditions interface

daniel-thom:


test/test_simulation_build.jl
```diff
@@ -1,13 +1,15 @@
-path = joinpath(pwd(), "test_sequence_build")
-!isdir(path) && mkdir(path)
+g_test_path = joinpath(pwd(), "test_sequence_build")
```
daniel-thom: The changes in this file are only there to prevent stages from being rebuilt.  They can be reverted if we don't go with that change.  I would argue that this approach is better regardless.

src/core/initial_conditions.jl
```diff
@@ -1,383 +1,128 @@
...
-        end
-    end
-
-    @assert length(ini_conds) == length_devices
-    return
-end
-
-function _make_initial_condition_active_power(
-    psi_container,
+"""
```
daniel-thom: @jd-lara These next two functions are a starting point for discussing the user interface for creating initial conditions.  This likely requires too much from the user.  What would you like to simplify?

jd-lara: We should move this function to a different PR to discuss the interface. My first impression is: 

1. users should not need to pass an `InitialConditions` to add them. 
2. It should be a single entry point like a kwarg passed into the stage/operation problem

src/core/update_initial_conditions.jl
```diff
@@ -0,0 +1,326 @@
...
+            _make_initial_condition_active_power,
+            _get_active_power_duration_value,
+            TimeStatusChange,
+        )
+    end
+
+    return
+end
+
+# TODO: This IC needs a cache for Simulation over long periods of time
```
jd-lara: This comment can be removed now. It was there before but if you can eliminate it would be great.

src/core/update_initial_conditions.jl
```diff
@@ -0,0 +1,326 @@
...
+function _make_initial_condition_energy(
+    container,
+    device::T,
+    value,
+    cache = nothing,
+) where {T <: PSY.Component}
+    return InitialCondition(device, _get_ref_energy(T, container), value, cache)
+end
+
+function _make_initial_condition_reservoir_energy(
```
jd-lara: Can these make functions be generalized to use multiple dispatches instead of having each named individually? It is not totally clear to me (right now) what the new flow of actions will be with these changes.

test/test_simulation_build.jl
```diff
@@ -1,13 +1,15 @@
...
+!isdir(g_test_path) && mkdir(g_test_path)
 
-function test_sequence_build(file_path::String)
-    stages_definition = Dict(
-        "UC" => Stage(GenericOpProblem, template_basic_uc, c_sys5_uc, GLPK_optimizer),
+function create_stages(template_uc)
+    return Dict(
+        "UC" => Stage(GenericOpProblem, template_uc, c_sys5_uc, GLPK_optimizer),
```
jd-lara: can we restore this to `template_basic_uc`?

test/test_simulation_build.jl
```diff
@@ -1,13 +1,15 @@
...
         "ED" => Stage(GenericOpProblem, template_ed, c_sys5_ed, GLPK_optimizer),
     )
+end
 
-    sequence = SimulationSequence(
+function create_sequence()
```
jd-lara: Nice change

src/core/update_initial_conditions.jl
daniel-thom: No logic in this file actually changed.  It got moved from another file.  I also changed PSIContainer -> InitialConditions.  Rather that try to optimize it, I'd prefer to make the prototype of building initial conditions from a specification that I discussed with you earlier this week.

test/test_simulation_build.jl
daniel-thom: This function is called many times in this file.  Two different global variables get passed in:  `template_basic_uc` and `template_standard_uc`.

I'm OK with reverting all changes in this file.  They are not needed because Stage will support `reset!`.  I left my changes here to let you see what the non-reset workflow looks like.  Let me know.

src/core/update_initial_conditions.jl
jd-lara: Ok, I guess this is an improvement for later

test/test_simulation_execute.jl
```diff
@@ -562,9 +562,35 @@ function test_load_simulation(file_path::String)
...
 end
+
 try
     test_load_simulation(g_test_path)
 finally
     @info("removing test files")
     rm(g_test_path, recursive = true)
 end
+
+@testset "Test Simulation with external initial conditions" begin
```
jd-lara: We should remove this test

src/PowerSimulations.jl
```diff
@@ -84,8 +84,8 @@ export StoredEnergy
 
 #operation_models
 export GenericOpProblem
-#export UnitCommitment
-#export EconomicDispatch
+export UnitCommitmentProblem
+export EconomicDispatchProblem
```
daniel-thom: I'm not sure how these got un-commented out.  Must be a merge error.  What should it be @jd-lara ?

jd-lara: I uncommented those, they had the wrong name.

PR #392, opened 2020-03-22
Jd/bug fix

jd-lara:
This PR fixes a bug introduced in the horizon setting for simulations.

### DIFF
--- a/src/core/psi_container.jl
+++ b/src/core/psi_container.jl
@@ -20,71 +20,6 @@ end
 
 const InitialConditionsContainer = Dict{ICKey, Array{InitialCondition}}
 #Defined here because of dependencies in psi_container
-function _make_jump_model(
-    JuMPmodel::Union{Nothing, JuMP.AbstractModel},
-    optimizer::Union{Nothing, JuMP.MOI.OptimizerWithAttributes},
-    parameters::Bool,
-)
-    if !isnothing(JuMPmodel)
-        if parameters
-            if !haskey(JuMPmodel.ext, :params)
-                @info("Model doesn't have Parameters enabled. Parameters will be enabled")
-                PJ.enable_parameters(JuMPmodel)
-            end
-        end
-        return JuMPmodel
-    end
-    if isa(optimizer, Nothing)
-        @debug "The optimization model has no optimizer attached"
-    end
-    @debug "Instantiating the JuMP model"
-    if !isnothing(optimizer)
-        JuMPmodel = JuMP.Model(optimizer)
-    else
-        JuMPmodel = JuMP.Model()
-    end
-    if parameters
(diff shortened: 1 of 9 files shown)


src/core/psi_container.jl
```diff
@@ -281,11 +301,11 @@ function PSIContainer(
     ini_con = get_initial_conditions(settings)
     optimizer = get_optimizer(settings)
     use_parameters = get_use_parameters(settings)
-    jump_model = _make_jump_model(jump_model, optimizer, use_parameters)
+    jump_model = _make_jump_model!(settings, jump_model, optimizer)
     if get_use_forecast_data(settings)
         time_steps = 1:get_horizon(settings)
         if length(time_steps) > 100
-            @warn("The number of time steps in the model is over 100. This will result in
+            @warn("The number of time steps in the model specification is over 100. This will result in
```
claytonpbarrows: This warning gets thrown when stages are constructed if the length of the system forecast is >100, regardless of the size of the resulting model. It's misleading to the user since the model doesn't actually end up with >100 time steps. We should move it to the beginning of the `build!(sim)` step and make sure that it only gets thrown if the JuMP Models will be large.

src/core/simulation_stages.jl
```diff
@@ -130,17 +130,15 @@ get_initial_time(s::Stage{T}) where {T <: AbstractOperationsProblem} =
...
-        0,
-        0,
-        PSIContainer(
-            stage.template.transmission,
-            stage.sys,
-            stage.internal.psi_container.settings,
-            nothing,
-        ),
+    if stage_built(stage)
+        @warn("Stage $(stage.internal.number) will be reset by the build call")
```
claytonpbarrows: I think this should be an Info print.

src/core/psi_container.jl
jd-lara: I agree with the idea in principle. However, consider that there isn't any way to "predict" what the problem size will be until we start building the JuMP model. We can add a check after building each component or something like that but it could add up to the build time. I will implement such a check, but I maintain that we need to be able to tell what size problem is being created.

jd-lara: @claytonpbarrows check 51cc9e8 and 437eb5c as ways to address this. In short, the warning is more informative and also there is a debug statement that calculates the problem size during the build problem. I made it a debug statement because I saw a 15% increase in build time with the RTS examples if we leave on by default.

claytonpbarrows: I think this is fine. If we wanted to, the check in 437eb5c4ef8652dabcd12ca4e4a358be1dbca0fd  could throw an error unless a kwarg like `force=true` was passed. I think this would alleviate some frustrating situations for users, but let's see how this works as is for now.

PR #395, opened 2020-03-24
Jd/psi container fix

jd-lara:


src/core/simulation_stages.jl
```diff
@@ -134,12 +128,9 @@ function reset!(stage::Stage)
         @info("Stage $(stage.internal.number) will be reset by the build call")
     end
```
daniel-thom: This if block should no longer be needed.  You should be able to make it an `@assert`.

src/core/simulation_stages.jl
```diff
@@ -149,13 +140,13 @@ function build!(
     horizon::Int,
     stage_interval::Dates.Period,
 )
+    stage_built(stage) && reset!(stage)
     settings = get_settings(get_psi_container(stage))
     set_horizon!(settings, horizon)
```
daniel-thom: It's not obvious why you are setting horizon and initial_time in `settings`.  A comment would help.

src/core/simulation_stages.jl
jd-lara: I think it is still needed, but will update the message.

src/core/operations_problem.jl
```diff
@@ -233,12 +228,8 @@ get_psi_container(op_problem::OperationsProblem) = op_problem.psi_container
 get_base_power(op_problem::OperationsProblem) = op_problem.sys.basepower
 
 function reset!(op_problem::OperationsProblem)
-    op_problem.psi_container = PSIContainer(
-        op_problem.template.transmission,
-        op_problem.sys,
-        op_problem.psi_container.settings,
-        nothing,
-    )
+    op_problem.psi_container =
```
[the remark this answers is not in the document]
jd-lara: I am not sure if this is strictly necessary.

PR #396, opened 2020-03-25
PSIMA-111: Multi-stage Cache

sourabhdalvi:
Hi guys, I'm currently struggling to come-up with some method that would allow us to have simulations where the  `TimeStatusChange` cache is modified by 1 or more stages.

I'm just going to list all the issue I faced and the temporary hack I have currently in place. I'll use this cache definition as  example
`    cache = Dict(
        ("UC", "HA") => TimeStatusChange(PSY.ThermalStandard, PSI.ON, 288, Minute(5)),
    ),`
1. Cache constructor
   https://github.com/NREL/PowerSimulations.jl/blob/b2bb84b1235d4c440e8e439af9f5b2bd10d85672/src/core/cache.jl#L13-L28
   - The current default constructor only stores one value for each device but with multiple stages you need to store [ max(UC_interval, HA_interval) / cache_unit ] number of values for each device.
https://github.com/NREL/PowerSimulations.jl/blob/b2bb84b1235d4c440e8e439af9f5b2bd10d85672/src/core/cache.jl#L30-L33
   - The cache `unit` field  needs to be min(UC_resolution, HA_resolution), there might be a way to get this during the stage construction when `_set_internal_caches` is called but for now just defined a new constructor that deals with both problems. 
   https://github.com/NREL/PowerSimulations.jl/blob/21c6a1605f8df8ace0674d8e6f9cdab40ec49c35/src/core/cache.jl#L35-L49
2. Initializing Cache
   -  Here the method relies on the new constructor to have created the correct number of time periods for cache. 
   https://github.com/NREL/PowerSimulations.jl/blob/21c6a1605f8df8ace0674d8e6f9cdab40ec49c35/src/core/simulation_stages.jl#L214-L219
   - It populates the first time period value in cache with initial conditions and the rest are populated with default values. This is wrong : 
       1.  You want the UC to use last time period (T) in the cache update comparison 
       2. the HA to use the 1st time period cache created by UC stage. 
    My current method always take the 1st  time step so wont work for multi-day simulation. 
3.  Updating Cache
    - With multiple stages we have to record the data (count & status) for each time-step in the cache and this can be messy if the stage resolution > cache resolution/unit. `r` here is used as a resolution factor =  stage resolution/ cache resolution.
    - https://github.com/NREL/PowerSimulations.jl/blob/88334e4031525675959993eb349b8b34276e3dce/src/core/simulation.jl#L521-L529
    - Here we also need to know what the execution count of the stage is to map the current variable time period to cache time period.
    - https://github.com/NREL/PowerSimulations.jl/blob/88334e4031525675959993eb349b8b34276e3dce/src/core/simulation.jl#L549-L560
    -  The reason I have execution count  (e) == 0 condition in there is because there is a bit of discrepancy in the way the count is reset and updated. For the UC stage  which has only one execution  it is incremented and reset in consecutive lines of code in `run_stage`
    - https://github.com/NREL/PowerSimulations.jl/blob/88334e4031525675959993eb349b8b34276e3dce/src/core/simulation_stages.jl#L183-L187
    -  So when the `_update_caches` call is made here after `run_stage`, it still shows zero execution rather than 1 but for the HA that has 24 executions it shows the correct number of execution except for the time after the last one. I tried to use `reset!(stage)` instead but that resulted in some other errors.
   -  https://github.com/NREL/PowerSimulations.jl/blob/88334e4031525675959993eb349b8b34276e3dce/src/core/simulation.jl#L734-L741

4. Initial Condition Update
    - The main issue is the initial condition now need to know which time period to access for the HA stage.(the example below is for InterStage Chronology)
    - The HA in the above case would need cache value at interval in the step resolution, here thing get complex because the cache index goes from 1:T where T = horizon * execution (horizon of the stage with smallest resolution).  Now we need a mapping of time periods in the horizon to time periods in the cache. This can be done by horizon * execution * resolution_factor . (resolution_factor = stage_resolution/cache_resolution).
   -  https://github.com/NREL/PowerSimulations.jl/blob/88334e4031525675959993eb349b8b34276e3dce/src/core/simulation.jl#L480-L489
    - But with the new method to calculate initial conditions I brake the existing method for IntraStage chronology which expects cache to have only a single period.
    -  https://github.com/NREL/PowerSimulations.jl/blob/88334e4031525675959993eb349b8b34276e3dce/src/core/initial_conditions.jl#L38-L46


sourabhdalvi:
I'll add more comments tonight

jd-lara:
Just a clarification, this issue is only applicable for the cases where the chronology is IntraStage. In the InterStage chronology, each stage updates its own cache and can follow the current implementation for single stage.

jd-lara:
This PR is stale and the approach to solve this issue will use a different approach than the one proposed here.

PR #400, opened 2020-03-26
Jd/branch models bug fixes

### LINKED ISSUE #384: Branch flow/rate constraints repetitions
In many data sets, the value of the rate limit for branches parsed as monitored branches is, in reality, a flow constraint. Currently, we add both, this behavior can be improved to reduce the number of constraints created.

### LINKED ISSUE #394: Standard PTDF Formulation Duplicates Line Rate Constraints
Seems to be the result of calling `device_range` twice in the function `branch_rate_constraints!` which is declared in lines 116 to 151 of AC_branches.jl on the master branch.

jd-lara:
closes #384 and closes #394. It also organizes better the cases where bounds are added to the flow variables and when constraints are added. @jmaack24 this should make the work you guys wanted easier. By default `StaticLine` doesn't add the bounds on the flow variable.


src/devices_models/devices/AC_branches.jl
```diff
@@ -1,19 +1,21 @@
 #Generic Branch Models
 abstract type AbstractBranchFormulation <: AbstractDeviceFormulation end
+abstract type AbstractBoundedBranchFormulation <: AbstractBranchFormulation end
```
claytonpbarrows: I like this

src/devices_models/devices/AC_branches.jl
```diff
@@ -34,113 +36,55 @@ function flow_variables!(
...
-            DeviceRange(name, limit_values, services_ub, Vector{Symbol}()),
-        )
-    end
-    set_variable_bounds!(psi_container, constraint_data, FLOW_ACTIVE_POWER_FROM_TO, B)
-    set_variable_bounds!(psi_container, constraint_data, FLOW_ACTIVE_POWER_TO_FROM, B)
+    model::DeviceModel{B, <:AbstractBranchFormulation},
+    ::Type{<:PM.AbstractDCPModel},
+) where {B <: PSY.ACBranch}
+    constraint_data = _get_constraint_data(devices)
+    set_variable_bounds!(psi_container, constraint_data, FLOW_ACTIVE_POWER, B)
```
claytonpbarrows: We don't need the directional limits anymore?

src/devices_models/devices/AC_branches.jl
```diff
@@ -177,22 +121,28 @@ function branch_rate_constraints!(
...
     feedforward::Union{Nothing, AbstractAffectFeedForward},
-)
+) where {T <: PM.DCPPowerModel}
     flow_range_data = Vector{PSI.DeviceRange}(undef, length(devices))
     for (ix, d) in enumerate(devices)
-        minmax = (min = PSY.get_flowlimits(d).to_from, max = PSY.get_flowlimits(d).from_to)
+        if PSY.get_flowlimits(d).to_from != PSY.get_flowlimits(d).from_to
+            @info("Flow limits in Line $(PSY.get_name(d)) aren't equal. The minimum will be used in formulation $(T)")
+        end
+        limit = min(
```
claytonpbarrows: This could create some confusing behavior for the user. For example, if they adjust their data because of a binding flow limit, but only adjust the data in `get_flowlimits(d)`,  they may get a new binding flow limit from the `rate` value. This sort of auto-detection is a bit scary. Could we just stick to flow limits here? If not, we should consider adding an interface in a future PR like `set_line_limits` or something that sets rate and flow_limits consistently.

src/devices_models/devices/AC_branches.jl
jd-lara: The DCPModels have only one flow variable

src/devices_models/devices/AC_branches.jl
jd-lara: The thing here is that we are creating way too many bounds/constraints that have the same value. In some cases is useful to have both depending on the solution algorithm for instance OSQP takes fewer iterations to solve when bounds and constraints are combined. But for larger systems, we could end up doubling the branch flow limits and making it worse. 

Also, I don't know if it makes sense to impose a flow limit larger than the rate that would be a data error. I suggest we leave like this and revisit when resolving the PSIMA story about selective enforcing of flow limits.

PR #402, opened 2020-03-30
PSIMA-115 :Fixing bug with HydroFixed formulation

sourabhdalvi:


sourabhdalvi:
Not sure why codecov doesn't like this

src/devices_models/devices/hydro_generation.jl
```diff
@@ -560,12 +560,12 @@ function nodal_expression!(
     end
 
     for t in model_time_steps(psi_container)
-        for device_value in ts_data_active
+        for device in ts_data_active
             _add_to_expression!(
                 psi_container.expressions[:nodal_balance_active],
-                device_value[2],
+                device.bus_number,
```
jd-lara: should use the getter_functions here.

sourabhdalvi: sorry, I just looked at how nodal balance was implement in other files and did the same

PR #406, opened 2020-04-06
Add recorder events

daniel-thom:
This implements infrastructure for the simulation recorder as described in PSIMA-65.  It is tied to https://github.com/NREL-SIIP/InfrastructureSystems.jl/pull/71/.  The tests will fail until that is merged and tagged.

This now includes logging changes for Simulation objects.  `build!(simulation)` will create its own logger and take over from the global logger during `build!()` and `execute()`.  It will log to a file in the simulation directory, parallel to the existing `results`.  After each function it will relinquish control to the global logger.

Documentation can be viewed here:  https://github.com/NREL-SIIP/PowerSimulations.jl/blob/PSIMA-65/docs/src/man/simulation_recorder.md


jd-lara:
@daniel-thom can you please update the IS dependency ceiling to 0.6 to run the tests.

PR #407, opened 2020-04-07
avoiding construction of containers for services that dont exist

claytonpbarrows:


src/devices_models/device_constructors/common/constructor_validations.jl
```diff
@@ -8,3 +8,14 @@ function validate_available_devices(devices, device::Type{D}) where {D <: PSY.De
     return false
 
 end
+
+function validate_available_services(services, service::Type{S}) where {S <: PSY.Service}
```
claytonpbarrows, line 12: It's worth noting that this function follows the template [established above](https://github.com/NREL-SIIP/PowerSimulations.jl/pull/407/files#diff-107bf08edb15b5b38e9a20cf02dfb610R1) where the logic seems backwards to me. `validate_available_xxx()` returns `true` if there are no components for the requested model, otherwise it returns false. Shouldn't it be the other way around?

jd-lara: there is a convention in PSY, IS and PSI that the functions check_value throw errors and the functions validate_value return true for valid and false for invalid.

claytonpbarrows: Yeah, but this is the opposite of the validate_value behavior.

PR #411, opened 2020-04-09
Add logging documentation page

daniel-thom:


docs/src/man/logging.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Configuring the global logger
+To configure the global logger in a Jupyter Notebook or REPL you may configure
+your own logger with the Julia Logging standard library or use the convenience
+function provided by PowerSimulations.  This example will log messages of level
+`Logging.Error` to console and `Logging.Info` and higher to the file
+`power-simulations.log` in the current directory.
+
+```julia
+import PowerSimulations
```
claytonpbarrows, line 20: Do you also need to add an `import Logging`?

docs/src/man/logging.md
```diff
@@ -0,0 +1,53 @@
...
+)
+```
+
+### Configuring the simulation logger
+You can configure the logging level used by the simulation logger when you call
+`build!(simulation)`.  Here is an example that increases logging verbosity:
+
+```julia
+import PowerSimulations
+simulation = Simulation(...)
```
claytonpbarrows, line 34: here too

kdheepak:
Can we run blocks of code in docs as part of tests?

daniel-thom:
> Can we run blocks of code in docs as part of tests?

1. I tried to follow the [DoctTest instructions](https://juliadocs.github.io/Documenter.jl/stable/man/doctests/index.html) and got `IOError: stream is closed or unusable` while the code was tried to create a log event and write to a file and console.  While I agree that we should do this, I think I may have exhausted the amount of time it's worth.
2. The second example requires building a simulation.  I didn't actually create a full example with all inputs because I thought that was distracting.  So, the code wouldn't run, anyway.  If you disagree on this point, I can make a full example.

PR #413, opened 2020-04-21
Jd/bump dependencies

jd-lara:
I have allowed the Windows tests to fail given what it seems to be a Julia bug in the `maketmpdir()` function. 

cc. @daniel-thom


test/test_simulation_build.jl
```diff
@@ -364,7 +364,8 @@ try
     test_sequence_build(g_test_path)
 finally
     @info("removing test files")
-    rm(g_test_path, recursive = true)
+    GC.gc()
+    rm(g_test_path, force = true, recursive = true)
```
claytonpbarrows, line 368: Do we need the file path to have a meaningful name? If not, maybe the [mktempdir functio](https://docs.julialang.org/en/v1/base/file/#Base.Filesystem.mktempdir-Tuple{AbstractString}) would resolve the issue.

daniel-thom: mktempdir not working was the reason for this PR.  @jd-lara Do we need the gc() here?  I thought that only mktempdir was causing problems.

jd-lara: The gc call was supposed to help with mktempdir error but it didn't, we can remove really.

PR #414, opened 2020-04-23
PSIMA-138 PSIMA-139 PSIMA-140 improving operations problem interfaces

claytonpbarrows:


src/core/operations_problem.jl
```diff
@@ -305,8 +307,17 @@ function construct_device!(
     if haskey(op_problem.template.devices, name)
         throw(IS.ConflictingInputsError("Device with model name $(name) already exists in the Opertaion Model"))
     end
-    devices_ref = get_devices_ref(op_problem)
-    devices_ref[name] = device_model
+    if device_model.device_type <: PSY.Branch
```
jd-lara, line 310: we should split these into function and call them `set_model(::PSY.Type)`

src/operations_problems_templates.jl
```diff
@@ -123,10 +123,20 @@ ed_problem = EconomicDispatchProblem(system)
 - `devices::Dict{Symbol, DeviceModel}` : override default `DeviceModel` settings
 - `branches::Dict{Symbol, DeviceModel}` : override default `DeviceModel` settings
 - `services::Dict{Symbol, ServiceModel}` : override default `ServiceModel` settings
+- Key word arguments supported by `OperationsProblem`
 """
 
 function EconomicDispatchProblem(system::PSY.System; kwargs...)
-    template = template_economic_dispatch(; kwargs...)
+    kwargs = Dict(kwargs)
```
jd-lara, line 130: we have a list of these already in `definitions.jl`

src/operations_problems_templates.jl
```diff
@@ -147,10 +157,20 @@ uc_problem = UnitCommitmentProblem(system)
...
 - `branches::Dict{Symbol, DeviceModel}` : override default `DeviceModel` settings
 - `services::Dict{Symbol, ServiceModel}` : override default `ServiceModel` settings
+- Key word arguments supported by `OperationsProblem`
 """
 
 function UnitCommitmentProblem(system::PSY.System; kwargs...)
-    template = template_unit_commitment(; kwargs...)
+    kwargs = Dict(kwargs)
+    template_kwargs = Dict()
+    for kw in (:network, :devices, :branches, :services)
```
jd-lara, line 166: same here. it is better to make a list by saying `not in OPERATIONS_ACCEPTED_KWARGS`

src/core/operations_problem.jl
claytonpbarrows: I don't follow this suggestion. I could see a function with seveeral methods like: `get_component_ref(<:PSY.Branch)` that returns the field of the template so that the addition is made in the right place.

claytonpbarrows: So, just to be clear... you are suggesting something like:
```julia
function set_model(op_problem::OperationsProblem, name::Symbol, device_type::D, device_model::DeviceModel) where D <: PSY.Branch
    branches_ref = get_branches_ref(op_problem)
    branches_ref[name] = device_model
end

function set_model(op_problem::OperationsProblem, name::Symbol, device_type::D, device_model::DeviceModel) where D <: PSY.StaticInjection
    devices_ref = get_devices_ref(op_problem)
    devices_ref[name] = device_model
end

function set_model(op_problem::OperationsProblem, name::Symbol, device_type::D, device_model::DeviceModel) where D <: PSY.Service
    services_ref = get_services_ref(op_problem)
    services_ref[name] = device_model
end
```
@jd-lara is that correct?
I don't know if this is an improvement. Especially since this won't have a performance impact.

jd-lara: This will allow us to keep flexible the internal implementation of the template.

daniel-thom: nit:  Julia convention is to prioritize function parameters that are types.
```
function set_model(::PSY.StaticInjection, op_problem, name, device_model)
```

Also, since you are in operations_problem.jl I wouldn't hesitate to use 

```
op_problem.template.devices[name] = device_model
```

src/operations_problems_templates.jl
```diff
@@ -197,10 +223,16 @@ results = run_economic_dispatch(system; optimizer = optimizer)
...
 - `optimizer::JuMP optimizer` : a JuMP optimizer is a required key word
+- `savepath::AbstractString`  : Path to save results
+- Key word arguments supported by `OperationsProblem`
 """
 
 function run_economic_dispatch(sys::PSY.System; kwargs...)
+    solve_kwargs = Dict()
+    for kw in OPERATIONS_SOLVE_KWARGS
+        haskey(kwargs, kw) && (solve_kwargs[kw] = kwargs[kw])
+    end
```
daniel-thom, line 234: This pattern is an intersection of sets.  In case you want something more clever (and there's probably something even shorter).

```
solve_kwargs = Dict(x => kwargs[x] for x in intersect(keys(kwargs), OPERATIONS_SOLVE_KWARGS))
```

src/operations_problems_templates.jl
```diff
@@ -123,10 +123,20 @@ ed_problem = EconomicDispatchProblem(system)
...
 """
 
 function EconomicDispatchProblem(system::PSY.System; kwargs...)
-    template = template_economic_dispatch(; kwargs...)
+    kwargs = Dict(kwargs)
+    template_kwargs = Dict()
+    for (kw, arg) in kwargs
+        if !(kw in OPERATIONS_ACCEPTED_KWARGS)
+            template_kwargs[kw] = kwargs[kw]
+            pop!(kwargs, kw)
```
daniel-thom, line 135: Similar to my other comment but `setdiff`.

`pop!` returns the value so you could combine lines 134 and 135.

PR #416, opened 2020-04-23
Jd/fixed output

jd-lara:
I am making this PR now because it is getting really large and it still doesn't address the requirements but it standardizes more the fixed output. 

Also, it fixes several issues of unimplemented code features.


src/services_models/reserves.jl
```diff
@@ -99,6 +91,17 @@ function modify_device_model!(
...
 
+function include_service!(
+    constraint_data::DeviceTimeSeries,
+    services::Vector{<:PSY.Reserve},
+    SM::ServiceModel,
+)
+    range_data = constraint_data.range
+    isnothing(range_data) && return
+    include_service!(range_data, services, SM)
+    return
```
daniel-thom, line 102: nit: IMHO fewer returns is better

```
if !isnothing(range_data)
    include_service!(range_data, services, SM)
end
return
```

src/services_models/reserves.jl
```diff
@@ -143,3 +145,35 @@ function _device_services!(
...
+function add_device_services!(
+    constraint_data_in::RangeConstraintsData,
+    constraint_data_out::RangeConstraintsData,
+    device::D,
+    model::DeviceModel{D, <:AbstractStorageFormulation},
+) where {D <: PSY.Storage}
+    for service_model in get_services(model)
+        if PSY.has_service(device, service_model.service_type)
+            services =
+                [s for s in PSY.get_services(device) if isa(s, service_model.service_type)]
```
daniel-thom, line 158: You don't actually need a vector. You could use a generator instead.

Minor readability thing:  I find `s isa service_model.service_type` more readable.  Just my opinion.

src/services_models/reserves.jl
```diff
@@ -143,3 +145,35 @@ function _device_services!(
...
+            @assert !isempty(services)
+            if service_model.service_type <: PSY.Reserve{PSY.ReserveDown}
+                for (ix, service) in enumerate(services)
+                    push!(
+                        constraint_data_in.additional_terms_ub,
+                        constraint_name(PSY.get_name(service), service_model.service_type),
+                    )
+                end
+            end
+            if service_model.service_type <: PSY.Reserve{PSY.ReserveUp}
```
daniel-thom, line 168: elseif

src/services_models/reserves.jl
```diff
@@ -143,3 +145,35 @@ function _device_services!(
...
+    device::D,
+    model::DeviceModel{D, <:AbstractStorageFormulation},
+) where {D <: PSY.Storage}
+    for service_model in get_services(model)
+        if PSY.has_service(device, service_model.service_type)
+            services =
+                [s for s in PSY.get_services(device) if isa(s, service_model.service_type)]
+            @assert !isempty(services)
+            if service_model.service_type <: PSY.Reserve{PSY.ReserveDown}
+                for (ix, service) in enumerate(services)
```
daniel-thom, line 161: You aren't using `ix` here or in the next clause.  `for service in services`

src/devices_models/device_constructors/renewablegeneration_constructor.jl
```diff
@@ -63,17 +61,16 @@ end
 function construct_device!(
     psi_container::PSIContainer,
     sys::PSY.System,
-    model::DeviceModel{R, RenewableFixed},
-    system_formulation::Type{S};
-    kwargs...,
+    model::DeviceModel{R, FixedOutput},
+    ::Type{S},
```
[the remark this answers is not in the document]
jd-lara: This is sort of at odds with the idea that the first argument should be the one mutated. E.g. `push!()`

src/devices_models/devices/common/add_parameters.jl
```diff
@@ -1,34 +1,12 @@
...
     psi_container::PSIContainer,
     ts_data::Vector{DeviceTimeSeries},
     param_reference::UpdateRef,
     expression_name::Symbol,
     multiplier::Float64 = 1.0,
 )
-    if !model_has_parameters(psi_container)
-        throw(IS.DataFormatError("Operational Model doesn't have parameters enabled. Include the keyword use_parameters=true"))
-    end
+    @assert model_has_parameters(psi_container)
```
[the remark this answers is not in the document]
jd-lara: No, this is an internal function that is meant for developers to call. It has helped to catch bad uses of `add_parameters` when the implementation is incorrect.

src/utils.jl
```diff
@@ -271,3 +271,7 @@ end
 function remove_chars(s::String, char::String)
     return replace_chars(s::String, char::String, "")
 end
+
+function get_available_components(::Type{T}, sys::PSY.System) where {T <: PSY.Component}
+    return PSY.get_components(T, sys, x -> PSY.get_available(x))
+end
```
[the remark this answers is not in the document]
jd-lara: I will leave the function until we merge https://github.com/NREL-SIIP/PowerSystems.jl/pull/459

src/services_models/reserves.jl
jd-lara: we use isa() as a function throughout the code in PSI.

src/devices_models/device_constructors/renewablegeneration_constructor.jl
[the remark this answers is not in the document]
daniel-thom: I'm slightly wrong.  https://docs.julialang.org/en/v1/manual/style-guide/index.html#Write-functions-with-argument-ordering-similar-to-Julia-Base-1

src/core/psi_container.jl
```diff
@@ -163,7 +163,7 @@ function psi_container_init!(
     end
 
     if get_use_forecast_data(settings)
-        total_number_of_devices = length(PSY.get_components(PSY.Device, sys))
+        total_number_of_devices = length(get_available_components(PSY.Device, sys))
```
daniel-thom, line 166: Seems like you never want to do anything with components that are not available.  Should we have a flag in PowerSystems that prevents those components from being added?  Maybe this is not appropriate...just an idea to simplify the logic.

src/devices_models/devices/hydro_generation.jl
```diff
@@ -220,10 +174,16 @@ function activepower_constraints!(
     parameters = model_has_parameters(psi_container)
     use_forecast_data = model_uses_forecasts(psi_container)
 
-    ts_data_active, constraint_data =
-        _get_time_series(psi_container, devices, model, x -> PSY.get_activepowerlimits(x))
-
     if !parameters && !use_forecast_data
```
daniel-thom, line 177: Is it possible for parameters to be true and use_forecast_data to be false?  This code doesn't handle that case.

src/devices_models/devices/renewable_generation.jl
```diff
@@ -150,14 +94,16 @@ function activepower_constraints!(
...
     use_forecast_data = model_uses_forecasts(psi_container)
 
-    ts_data_active, _, constraint_data = _get_time_series(
-        psi_container,
-        devices,
-        model,
-        x -> (min = 0.0, max = PSY.get_activepower(x)),
-    )
-
     if !parameters && !use_forecast_data
```
daniel-thom, line 97: Same comment here

src/devices_models/devices/hydro_generation.jl
jd-lara: I don't think that case was ever handled.

src/core/psi_container.jl
jd-lara: I rather leave it as it is and have the users make modifications to their systems if they want to make a device unavailable explicitly.

PR #419, opened 2020-04-29
Fix pwl check

claytonpbarrows:
corrects the pwl check to verify convexity.


jd-lara:
@sourabhdalvi this doesn't resolve yet the implementation of the limits for the units under PWL costs.

claytonpbarrows:
I’m not convinced we should do what we discussed on the phone. If we do, we’ll either have to multiply by commitment status or we would be adding the P=0 cost when the unit is off. This way only fails when the 0th traunch is steeper than the others, and otherwise gets the right objective function. It will work in the other case, it just adds sos.

jd-lara:
> I’m not convinced we should do what we discussed on the phone. If we do, we’ll either have to multiply by commitment status or we would be adding the P=0 cost when the unit is off. 

It will imply that a Gen formulated as no-min is always on but has no minimum stable limit. 

> This way only fails when the 0th traunch is steeper than the others, and otherwise gets the right objective function. It will work in the other case, it just adds sos.

I don't think is accurate to say that if a unit formulated as no-min then the cost = 0 when output = 0. This case will then imply that _"almost"_ by default no min + pwl cost will imply an SOS cost function thus defeating the purpose of having simple linear model to get fast and dirty solutions.

PR #420, opened 2020-04-29
Jd/re export logging

jd-lara:
Avoid requiring the user to import Logging for the user of the logger.


daniel-thom:
I wouldn't have done this.  I don't see a problem making the user import a standard library module.

Suppose a user writes `import PowerSimulations` instead of `using PowerSimulations`.  Now you are suggesting they use `PowerSimulations.Info`.  I find that a bit odd.

If I were a user the first thing I would ask or try to find out is whether PSI levels were the same or different as Logging levels.

They can always use `using Logging`.

Just my 2 cents.

PR #424, opened 2020-05-04
Psima 150

Lilyhanig:
broadcast base power to parameters and variables that are not binary for CSV writing


claytonpbarrows:
Lgtm. It looks like you need to resolve some merge conflicts

PR #425, opened 2020-05-04
ORDC implementation

sourabhdalvi:
For this implementation, I had to make the following changes over the RangeReserve Formulation
- Adding `ORDC` PSY struct & new formulation `OperatingReserveDemandCurve`
- Instead of adding a parameter or forecast for requirement there is new variable for the RHS
   - Changed `add_variable`  to accept any subtype of `Component` as Services are not under `Device`
- This requirement variable has a PWL cost 
    - Changed  `add_to_cost`  to accept any subtype of `Component` as Services are not under `Device`
- Change `modify_device_model!` to accept any type of reserve formulation. This is probably not right but didn't want to repeat the as function for ORDC formulation.

These are just draft change

### DIFF
--- a/src/services_models/reserves.jl
+++ b/src/services_models/reserves.jl
@@ -124,7 +124,7 @@ end
 
 function modify_device_model!(
     devices_template::Dict{Symbol, DeviceModel},
-    service_model::ServiceModel{<:PSY.Reserve, RangeReserve},
+    service_model::ServiceModel{<:PSY.Reserve, <:AbstractReservesFormulation},
     contributing_devices::Vector{<:PSY.Device},
 )
     device_types = unique(typeof.(contributing_devices))
--- a/src/services_models/reserves.jl
+++ b/src/services_models/reserves.jl
@@ -1,5 +1,6 @@
 abstract type AbstractReservesFormulation <: AbstractServiceFormulation end
 struct RangeReserve <: AbstractReservesFormulation end
+struct OperatingReserveDemandCurve <: AbstractReservesFormulation
 ############################### Reserve Variables` #########################################
 """
 This function add the variables for reserves to the model
@@ -15,6 +16,20 @@ function activeservice_variables!(
         variable_name(PSY.get_name(service), SR),
         false;
         lb_value = d -> 0,
+    )
+    return
+end
+
+function activerequirement_variables!(
+    psi_container::PSIContainer,
+    devices::IS.FlattenIteratorWrapper{SR},
+) where {SR <: PSY.Reserve}
+    add_variable(
+        psi_container,
+        devices,
+        variable_name(SERVICE_REQUIREMENT, SR),
+        false;
+        lb_value = x -> 0.0,
     )
     return
 end
@@ -70,6 +85,40 @@ function service_requirement_constraint!(
             )
         end
     end
+    return
+end
+
+function service_requirement_constraint!(
+    psi_container::PSIContainer,
+    service::SR,
+    model::ServiceModel{SR, OperatingReserveDemandCurve},
+) where {SR <: PSY.Reserve}
+    parameters = model_has_parameters(psi_container)
+    use_forecast_data = model_uses_forecasts(psi_container)
+    initial_time = model_initial_time(psi_container)
+    @debug initial_time
+    time_steps = model_time_steps(psi_container)
+    name = PSY.get_name(service)
+    constraint = get_constraint(psi_container, constraint_name(REQUIREMENT, SR))
+    reserve_variable = get_variable(psi_container, variable_name(name, SR))
+    requirement_variable = get_variable(psi_container, variable_name(SERVICE_REQUIREMENT, SR))
+
+    for t in time_steps
+        constraint[name, t] = JuMP.@constraint(
+            psi_container.JuMPmodel,
+            sum(reserve_variable[:, t]) >= requirement_variable[name, t]
+        )
+    end
+
+    return
+end
+
+function cost_function(
+    psi_container::PSIContainer,
+    devices::IS.FlattenIteratorWrapper{SR},
+    ::Type{OperatingReserveDemandCurve},
+) where {SR <: PSY.Reserve}
+    add_to_cost(psi_container, devices, variable_name(SERVICE_REQUIREMENT, SR), :variable, -1.0)
     return
 end
 
--- a/src/services_models/reserves.jl
+++ b/src/services_models/reserves.jl
@@ -1,6 +1,6 @@
 abstract type AbstractReservesFormulation <: AbstractServiceFormulation end
 struct RangeReserve <: AbstractReservesFormulation end
-struct OperatingReserveDemandCurve <: AbstractReservesFormulation
+struct OperatingReserveDemandCurve <: AbstractReservesFormulation end
 ############################### Reserve Variables` #########################################
 """
 This function add the variables for reserves to the model
--- a/src/devices_models/devices/common/add_variable.jl
+++ b/src/devices_models/devices/common/add_variable.jl
@@ -46,7 +46,7 @@ function add_variable(
     expression_name::Union{Nothing, Symbol} = nothing,
     sign::Float64 = 1.0;
     kwargs...,
-) where {D <: Union{Vector{<:PSY.Device}, IS.FlattenIteratorWrapper{<:PSY.Device}}}
+) where {D <: Union{Vector{<:PSY.Component}, IS.FlattenIteratorWrapper{<:PSY.Component}}}
     time_steps = model_time_steps(psi_container)
     variable = add_var_container!(
         psi_container,
--- a/src/devices_models/devices/common/cost_functions.jl
+++ b/src/devices_models/devices/common/cost_functions.jl
@@ -390,7 +390,7 @@ function add_to_cost(
     var_name::Symbol,
     cost_symbol::Symbol,
     sign::Float64 = 1.0,
-) where {D <: IS.FlattenIteratorWrapper{<:PSY.Device}}
+) where {D <: IS.FlattenIteratorWrapper{<:PSY.Component}}
     resolution = model_resolution(psi_container)
     dt = Dates.value(Dates.Minute(resolution)) / 60
     variable = get_variable(psi_container, var_name)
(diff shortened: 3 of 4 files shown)


src/services_models/reserves.jl
```diff
@@ -19,6 +20,20 @@ function activeservice_variables!(
     return
 end
 
+function activerequirement_variables!(
+    psi_container::PSIContainer,
+    devices::IS.FlattenIteratorWrapper{SR},
+) where {SR <: PSY.Reserve}
```
jd-lara, line 26: This should be dispatched on the specific PSY reserve type that uses this.

sourabhdalvi: That make sense, will make that change

jd-lara:
you might want to wait until #423 is merged to continue

sourabhdalvi:
> per our conversation, I think renaming to `PWLReserveRequirement` or something along those lines make sense

I changed the name to StepwiseCostReserve, because that's the type of cost I'm implementing. Where the cost curve is a step function of the shortfall of reserves but I'm open to suggestions

jd-lara:
@sourabhdalvi what's the status of this?

test/test_data/data_5bus_pu.jl
```diff
@@ -598,20 +601,39 @@ reserve5(thermal_generators5) = [
...
 
 reserve5_il(interruptible_loads) = [
     VariableReserve{ReserveUp}("Reserve7", true, 30, 100),
     VariableReserve{ReserveDown}("Reserve8", true, 5, 50),
+    ReserveDemandCurve{ReserveUp}("ORDC3", true, 0.6, ORDC_cost),
+]
+
+function make_ordc_cost(cost::TwoPartCost)
+    var_cost = PSY.get_cost(PSY.get_variable(cost))
+    flatten_array = Array(collect(Iterators.flatten(var_cost))')
```
daniel-thom, line 626: I don't follow this.  `name` is an array but `flatten_array` is an array of arrays.  The reason isn't obvious to me.  Is it valid Julia code?  I see a single quote at the end that looks like a bug.

sourabhdalvi: What I'm wanted to do was convert a array of tuples 
```julia
 var_cost = PSY.get_cost(PSY.get_variable(ORDC_cost))
5-element Array{Tuple{Float64,Float64},1}:
 (9000.0, 0.0)
 (6000.0, 0.2)
 (500.0, 0.4)
 (10.0, 0.6)
 (0.0, 0.8)
```
to something like this 

```julia
Array(collect(Iterators.flatten(var_cost))')
1×10 Array{Float64,2}:
 9000.0  0.0  6000.0  0.2  500.0  0.4  10.0  0.6  0.0  0.8
```

sourabhdalvi: The single quote at the end is just to do a transpose

daniel-thom: What you have is fine.  I wasn't familiar with that Julia syntax.  Sorrry.

src/services_models/reserves.jl
```diff
@@ -81,9 +96,113 @@ function service_requirement_constraint!(
...
+function service_requirement_constraint!(
+    psi_container::PSIContainer,
+    service::SR,
+    model::ServiceModel{SR, StepwiseCostReserve},
+) where {SR <: PSY.Reserve}
+    parameters = model_has_parameters(psi_container)
+    use_forecast_data = model_uses_forecasts(psi_container)
+    initial_time = model_initial_time(psi_container)
+    @debug initial_time
+    time_steps = model_time_steps(psi_container)
```
daniel-thom, line 108: Nit:  These are unused. Do you need them?

test/test_utils/get_test_data.jl
```diff
@@ -224,9 +224,17 @@ function build_c_sys5_re(; kwargs...)
             reserve_re[2],
             [collect(get_components(RenewableDispatch, c_sys5_re))[end]],
         )
+        add_service!(c_sys5_re, reserve_re[3], get_components(RenewableDispatch, c_sys5_re))
         for t in 1:2, (ix, serv) in enumerate(get_components(VariableReserve, c_sys5_re))
             add_forecast!(c_sys5_re, serv, Deterministic("get_requirement", Reserve_ts[t]))
         end
+        for t in 1:2, (ix, serv) in enumerate(get_components(ReserveDemandCurve, c_sys5_re))
```
daniel-thom, line 231: Nit: `ix` isn't used in this new code, so you don't need `enumerate`.  I see that you probably copy/pasted existing code that does the same thing.

src/services_models/reserves.jl
```diff
@@ -81,9 +96,113 @@ function service_requirement_constraint!(
...
+        ts_vector = repeat(PSY.get_variable(PSY.get_op_cost(service)), time_steps[end])
+    end
+
+    resolution = model_resolution(psi_container)
+    dt = Dates.value(Dates.Minute(resolution)) / 60
+    variable = get_variable(psi_container, variable_name(SERVICE_REQUIREMENT, SR))
+    gen_cost = JuMP.GenericAffExpr{Float64, _variable_type(psi_container)}()
+    if !haskey(psi_container.variables, :PWL_RDC_cost_vars)
+        time_steps = model_time_steps(psi_container)
+        container = add_var_container!(
```
daniel-thom, line 155: Line 151 calls `get_variable` and line 153 goes directly to `psi_container.variables`.  This comment is not directed at this PR but we probably need to improve access to variables - @jd-lara

src/services_models/reserves.jl
```diff
@@ -81,9 +96,113 @@ function service_requirement_constraint!(
...
+            container[(PSY.get_name(service), t, ix)] = v
+        end
+        JuMP.add_to_expression!(gen_cost, c)
+    end
+
+    cost_expression = gen_cost * dt
+    T_ce = typeof(cost_expression)
+    T_cf = typeof(psi_container.cost_function)
+    if T_cf <: JuMP.GenericAffExpr && T_ce <: JuMP.GenericQuadExpr
+        psi_container.cost_function += cost_expression
```
daniel-thom, line 178: @jd-lara We should probably have a helper function to mutate psi_container.cost_function.

test/test_data/data_5bus_pu.jl
```diff
@@ -579,6 +579,9 @@ interruptible(nodes5) = [InterruptibleLoad(
     TwoPartCost(150.0, 2400.0),
 )]
 
+const ORDC_cost =
```
jd-lara, line 582: why are you making this a `const`?

test/test_utils/get_test_data.jl
jd-lara: instead of enumerate you can use `eachvalue`

src/services_models/reserves.jl
jd-lara: @sourabhdalvi I don't follow what you are trying to do here. It seems you copied the code from the optional PWL cost functions but in this case, I don't see why you need to do this instead of creating the variables upfront. In the PWL implementation, I used this pattern because storing those is optional. 

This part of the code needs revision along those lines.

src/services_models/reserves.jl
jd-lara: yes, we should move this to another function. This can be done later after merging, we should open a PSIMA issue for it.

src/services_models/reserves.jl
```diff
@@ -81,9 +96,113 @@ function service_requirement_constraint!(
...
+            "get_variable",
+            length(time_steps),
+        )))
+
+    else
+        ts_vector = repeat(PSY.get_variable(PSY.get_op_cost(service)), time_steps[end])
+    end
+
+    resolution = model_resolution(psi_container)
+    dt = Dates.value(Dates.Minute(resolution)) / 60
```
jd-lara, line 150: There is a conversion function in IS to do this conversion. Also. please don't divide by integers, there is a const in `definitions.jl` for `MINUTES_IN_HOUR`

src/services_models/reserves.jl
```diff
@@ -81,9 +96,113 @@ function service_requirement_constraint!(
...
+            [PSY.get_name(service)],
+            time_steps,
+            1:length(ts_vector[1]);
+            sparse = true,
+        )
+    else
+        container = get_variable(psi_container, :PWL_RDC_cost_vars)
+    end
+    for (t, var) in enumerate(variable[PSY.get_name(service), :])
+        c, pwlvars = _pwlgencost_sos(psi_container, var, ts_vector[t])
```
jd-lara, line 167: If you are using `_pwlgencost_sos` here and it isn't in the same file please rename the function by removing the `_`

src/services_models/reserves.jl
jd-lara, line 155: I think you can allocate the var container beforehand because there will be a separate container for each ORDC reserve.

src/services_models/reserves.jl
```diff
@@ -81,9 +96,113 @@ function service_requirement_constraint!(
...
+
+    else
+        ts_vector = repeat(PSY.get_variable(PSY.get_op_cost(service)), time_steps[end])
+    end
+
+    resolution = model_resolution(psi_container)
+    dt = Dates.value(Dates.Minute(resolution)) / 60
+    variable = get_variable(psi_container, variable_name(SERVICE_REQUIREMENT, SR))
+    gen_cost = JuMP.GenericAffExpr{Float64, _variable_type(psi_container)}()
+    if !haskey(psi_container.variables, :PWL_RDC_cost_vars)
```
jd-lara, line 153: The name should include the type and name and create a container per reserve.

src/services_models/reserves.jl
sourabhdalvi: I removed those lines other than this one
```julia
    initial_time = model_initial_time(psi_container)
    @debug initial_time
```

test/test_utils/get_test_data.jl
daniel-thom: or just `serv in get_components(ReserveDemandCurve, c_sys5_re)`

sourabhdalvi: Sorry, it was a copy-paste mistake

PR #428, opened 2020-05-07
PSIMA-145:Automating cache creation using initial conditions

sourabhdalvi:
That to make few changes to the existing methods
- Changed `get_stage_cache_definition`  to return an empty cache vector instead of nothing. (It is easier to populate empty array than a nothing)
- Added `_add_initial_condition_caches` that loops over the initial condition keys and checks if a cache is passed by the use, if not creates it.
- Changed `test_simulation_build` test to check for cache instead of error thrown

### DIFF
--- a/src/core/simulation.jl
+++ b/src/core/simulation.jl
@@ -122,46 +122,47 @@ function _add_initial_condition_caches(
     return
 end
 
-function _create_cache(ic_key::ICKey,
-    caches::Union{Nothing, Vector{<:AbstractCache}},
-)
+function _create_cache(ic_key::ICKey, caches::Union{Nothing, Vector{<:AbstractCache}})
     return
 end
 
-function _create_cache(ic_key::ICKey{Type{T}, TimeDurationON}, 
+function _create_cache(
+    ic_key::ICKey{TimeDurationON, Type{T}},
     caches::Union{Nothing, Vector{<:AbstractCache}},
-) where {T<: PSY.Device}
+) where {T <: PSY.Device}
 
     cache_keys = CacheKey.(caches)
     if isempty(cache_keys) || !in(CacheKey(TimeStatusChange, Type{T}), cache_keys)
         cache = TimeStatusChange(T, PSI.ON)
         push!(caches, cache)
     end
-    return 
+    return
 end
 
-function _create_cache(ic_key::ICKey{TimeDurationOFF, Type{T}}, 
+function _create_cache(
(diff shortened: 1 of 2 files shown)


src/core/simulation.jl
```diff
@@ -110,6 +110,61 @@ function unregister_recorders!(internal::SimulationInternal)
     end
 end
 
+function _add_initial_condition_caches(
+    sim::SimulationInternal,
+    stage::Stage,
+    caches::Union{Nothing, Vector{<:AbstractCache}},
```
jd-lara, line 116: Add a debug statement here is the caches are `nothing`

src/core/simulation.jl
```diff
@@ -110,6 +110,61 @@ function unregister_recorders!(internal::SimulationInternal)
...
+    return
+end
+
+function _create_cache(ic_key::ICKey, caches::Union{Nothing, Vector{<:AbstractCache}})
+    return
+end
+
+function _create_cache(
+    ic_key::ICKey{TimeDurationON, T},
+    caches::Union{Nothing, Vector{<:AbstractCache}},
```
jd-lara, line 131: why a nothing here?

src/core/simulation.jl
```diff
@@ -110,6 +110,61 @@ function unregister_recorders!(internal::SimulationInternal)
...
+        push!(caches, cache)
+    end
+    return
+end
+
+function _create_cache(
+    ic_key::ICKey{EnergyLevel, T},
+    caches::Union{Nothing, Vector{<:AbstractCache}},
+) where {T <: PSY.Device}
+
```
jd-lara, line 159: I don't see why the caches will take a nothing here

sourabhdalvi: I should actually remove that, but the reason I initial added it there was this function returned `Nothing`
https://github.com/NREL-SIIP/PowerSimulations.jl/blob/6ffdfa826540d66837af12af5b011d3c12e1ee77/src/core/simulation.jl#L242-L254

src/core/simulation.jl
sourabhdalvi: Sorry I forgot to remove that

src/core/simulation.jl
sourabhdalvi: Since the nothing will be removed that won't be needed

sourabhdalvi:
> I have some questions about the input constraints to the `_create_cache` functions

Hopefully the comment above clear thing up. Short story, I forgot to remove it.

PR #431, opened 2020-05-10
shortcutting result checks for non-optimization stage

claytonpbarrows:
avoids MILP check when stage doesn't have a JuMP model.

### DIFF
--- a/src/core/psi_container.jl
+++ b/src/core/psi_container.jl
@@ -490,6 +490,7 @@ function get_parameters_value(psi_container::PSIContainer)
 end
 
 function is_milp(container::PSIContainer)
+    isnothing(container.JuMPmodel) && return false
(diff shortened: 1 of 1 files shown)


jd-lara:
There should be a step somewhere before this call to address this where no psi_container is passed.

claytonpbarrows:
Like here? https://github.com/NREL-SIIP/PowerSimulations.jl/blob/d6720672de29c7280a362af8494afb23384d4625/src/core/simulation_results.jl#L74

jd-lara:
@claytonpbarrows, yes pretty much all the code in the for-loop needs to be made into another function parametrizing Stage and define custom make reference function for custom stages

jd-lara:
solved via #430

PR #433, opened 2020-05-12
Save PWL variables

jd-lara:


src/devices_models/devices/common/cost_functions.jl
```diff
@@ -177,7 +180,7 @@ function _pwlgencost_sos(
     gen_cost = JuMP.GenericAffExpr{Float64, _variable_type(psi_container)}()
     pwlvars = JuMP.@variable(
         psi_container.JuMPmodel,
-        [i = 1:length(cost_component)],
+        [i in 1:length(cost_component)],
```
daniel-thom, line 183: Does JuMP accept `1:length(cost_component)`?  If not, you could still simplify to `collect(1:length(cost_component))`.  Feel free to ignore.

jd-lara: no need to collect, the macro expands it when it gets evaluated. This is why it sometimes is capable of using generators as axes

PR #435, opened 2020-05-13
fixing pwl cost filter bug

claytonpbarrows:


src/devices_models/devices/common/cost_functions.jl
```diff
@@ -338,6 +335,9 @@ function ps_cost(
     dt::Float64,
     sign::Float64,
 )
+    cost_array = cost_component.cost
+    # If array is full of tuples with zeros return 0.0
+    all(iszero.(last.(cost_array))) && return 0.0
```
jd-lara, line 340: Change the return to ` AffExpr(0.0)` to keep the function type stable

PR #436, opened 2020-05-14
Refactor tests to build systems in function calls

daniel-thom:
This doesn't address all of PSIMA-147.  I'd like to take it in stages.  This removes most of the global variables and builds systems from function calls.

Note that it is likely slower than the previous version because it rebuilds systems many times.

### DIFF
--- a/test/test_services_constructor.jl
+++ b/test/test_services_constructor.jl
@@ -14,6 +14,7 @@
         branches,
         services_template,
     )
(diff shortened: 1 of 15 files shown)


test/test_services_constructor.jl
```diff
@@ -14,6 +14,7 @@
         branches,
         services_template,
     )
+    c_sys5_uc = build_c_sys5_uc(; add_reserves = true)
```
jd-lara, line 17: nice one

PR #439, opened 2020-05-19
Refactor time series range constraints

daniel-thom:


src/devices_models/devices/common/add_parameters.jl
```diff
@@ -1,27 +1,27 @@
...
+    constraints::Vector{DeviceTimeSeriesRangeConstraint},
     param_reference::UpdateRef,
     expression_name::Symbol,
     multiplier::Float64 = 1.0,
 )
     @assert model_has_parameters(psi_container)
     time_steps = model_time_steps(psi_container)
     ## TODO, replace multiplier
-    names = [r.name for r in ts_data]
+    names = [get_name(r) for r in constraints]
```
jd-lara, line 12: I think this one should work as a generator. It was an array before but it is an improvement.

src/devices_models/devices/common/timeseries_constraint.jl
```diff
@@ -1,31 +1,33 @@
...
     con_lb = add_cons_container!(psi_container, lb_name, names, time_steps)
 
-    for data in ts_data
-        if data.range.limits.min > -Inf && !isempty(data.range.additional_terms_lb)
-            con_lb[data.name, :].data .= JuMP.AffExpr(0.0)
+    for ts_range_constraint in range_constraints
+        rc_name = get_name(ts_range_constraint)
+        if ts_range_constraint.range_constraint.limits.min > -Inf &&
+           !isempty(ts_range_constraint.range_constraint.additional_terms_lb)
+            con_lb[rc_name, :].data .= JuMP.AffExpr(0.0)
```
jd-lara, line 17: Note to me: Check that this is broadcasting properly, we used to have a problem with this in JuMP since it assigned every entry of the array the same object.

src/devices_models/devices/common/timeseries_constraint.jl
```diff
@@ -225,52 +234,53 @@ Constructs lower bound for given variable using a parameter. The constraint is
...
+    range_constraints::Vector{DeviceTimeSeriesRangeConstraint},
     cons_name::Symbol,
     param_reference::UpdateRef,
     var_name::Symbol,
 )
     time_steps = model_time_steps(psi_container)
     variable = get_variable(psi_container, var_name)
     lb_name = middle_rename(cons_name, PSI_NAME_DELIMITER, "lb")
-    names = (d.name for d in ts_data)
+    names = [get_name(x) for x in range_constraints]
```
jd-lara, line 260: why revert to an array?

src/devices_models/devices/common/timeseries_constraint.jl
```diff
@@ -385,34 +396,35 @@ function device_timeseries_ub_bigM(
 
     varcts = get_variable(psi_container, var_name)
     varbin = get_variable(psi_container, binvar_name)
-    names = (d.name for d in ts_data)
+    names = [get_name(x) for x in range_constraints]
```
jd-lara, line 399: same here about the change from generator to array

src/devices_models/devices/common/timeseries_constraint.jl
daniel-thom: I thought I had a good reason.  Reverted.

src/devices_models/devices/hydro_generation.jl
```diff
@@ -147,17 +147,17 @@ function reactivepower_constraints!(
     system_formulation::Type{<:PM.AbstractPowerModel},
     feedforward::Union{Nothing, AbstractAffectFeedForward},
 ) where {H <: PSY.HydroGen, D <: AbstractHydroDispatchFormulation}
-    constraint_data = Vector{DeviceRange}(undef, length(devices))
+    constraints = Vector{DeviceRangeConstraint}(undef, length(devices))
```
jd-lara, line 150: I would have preferred to keep this as `constraint_data` since the constraint isn't implemented here is just the data. to add it to the model

src/devices_models/devices/renewable_generation.jl
```diff
@@ -42,7 +42,7 @@ function reactivepower_constraints!(
     system_formulation::Type{<:PM.AbstractPowerModel},
     feedforward::Union{Nothing, AbstractAffectFeedForward},
 ) where {R <: PSY.RenewableGen}
-    constraint_data = Vector{DeviceRange}(undef, length(devices))
+    constraints = Vector{DeviceRangeConstraint}(undef, length(devices))
```
jd-lara, line 45: same all over where this substitution happened

src/devices_models/devices/hydro_generation.jl
```diff
@@ -435,6 +436,27 @@ function initial_conditions!(
 end
 
 ########################## Addition of to the nodal balances ###############################
+
+#function NodalExpressionInputs(::Type{<:PSY.HydroGen}, ::Type{<:PM.AbstractPowerModel}, use_forecasts::Bool)
```
daniel-thom, line 440: What should we do about this section?  I left the existing code.

jd-lara: This code is a copy of the RenewableEnergy and it is non-sensical we can eliminate it

src/devices_models/devices/hydro_generation.jl
```diff
@@ -434,97 +435,18 @@ function initial_conditions!(
...
+function NodalExpressionInputs(
+    ::Type{<:PSY.HydroGen},
+    ::Type{<:PM.AbstractActivePowerModel},
+    use_forecasts::Bool,
+)
+    return NodalExpressionInputs(
+        "get_rating",
+        ACTIVE_POWER,
+        use_forecasts ? x -> PSY.get_rating(x) : x -> PSY.get_activepower(x),
+    )
```
daniel-thom, line 449: This change needs scrutiny.  Note that there is no implementation for reactive power.

jd-lara: Point taken, lets addressed it in a separate PR

PR #440, opened 2020-05-20
Jd/agc model

jd-lara:
This is WIP, the review is mostly for very bad errors.


src/services_models/services_constructor.jl
```diff
@@ -72,3 +72,32 @@ function construct_service!(
...
 end
+
+function construct_service!(
+    psi_container::PSIContainer,
+    services::IS.FlattenIteratorWrapper{PSY.AGC},
+    services_mapping::PSY.ServiceContributingDevicesMapping,
+    ::ServiceModel{PSY.AGC, T},
+    devices_template::Dict{Symbol, DeviceModel},
+) where {T <: AbstractAGCFormulation}
+    #Note: Frequency is a global variable
```
daniel-thom, line 83: I don't follow the comment.

src/services_models/services_constructor.jl
```diff
@@ -72,3 +72,32 @@ function construct_service!(
...
+    steady_state_frequency_variables!(psi_container)
+
+    for device_model in devices_template
+    end
```
daniel-thom, line 87: Is this supposed to do something?

src/services_models/services_constructor.jl
```diff
@@ -72,3 +72,32 @@ function construct_service!(
...
+            services_mapping[(
+                type = typeof(service),
+                name = PSY.get_name(service),
+            )].contributing_devices
+        #Variables
+        regulation_service_variables!(psi_container, service, contributing_devices)
+        # Constraints
+        smooth_ace_pid!(psi_container, service)
+        #steady_state_frequency_variables!(psi_container, service, model, )
+        #participation_assignment!(psi_container, service, model, contributing_devices)
```
daniel-thom, line 100: stale

src/services_models/services_constructor.jl
jd-lara: The Frequency of the system is the same all over not just for an area.

src/services_models/services_constructor.jl
jd-lara: a check goes here. Not implemented yet

src/services_models/services_constructor.jl
jd-lara: It is a broken code.

PR #451, opened 2020-06-14
Fix assignment of parameters during constraint creation

### LINKED ISSUE #450: Simulation initial conditions skip using ParameterJump
In the refactor of initial conditions the references to create ParameterRef's were lost. Currently, all initial conditions are kept from the beginning of the simulation.

jd-lara:
closes #450


src/core/update_initial_conditions.jl
```diff
@@ -82,18 +82,58 @@ function calculate_ic_quantity(
...
+    if !has_initial_conditions(ic_container, key)
+        @debug "Setting $(key.ic_type) initial conditions for the status of all devices $(T) based on system data"
+        ini_conds = Vector{InitialCondition}(undef, length_devices)
+        set_initial_conditions!(ic_container, key, ini_conds)
+        for (ix, dev) in enumerate(devices)
+            val_ = get_val_func(dev, key)
+            val = parameters ? PJ.add_parameter(psi_container.JuMPmodel, val_) : val_
+            ic = make_ic_func(ic_container, dev, val, cache)
+            ini_conds[ix] = ic
+            @debug key, ic, val, 0
```
daniel-thom, line 106: No commas in between the variable names.  Probably don't need the 0.  I have found it to be a good practice to always include a string in the log message that you can search for. 

Also, do we want a recorder event here?

src/core/update_initial_conditions.jl
```diff
@@ -82,18 +82,58 @@ function calculate_ic_quantity(
...
+    else
+        ini_conds = get_initial_conditions(ic_container, key)
+        ic_devices = Set((IS.get_uuid(ic.device) for ic in ini_conds))
+        for dev in devices
+            IS.get_uuid(dev) in ic_devices && continue
+            @debug "Setting $(key.ic_type) initial conditions for the status device $(PSY.get_name(dev)) based on system data"
+            val = get_val_func(dev, key)
+            ic = make_ic_func(ic_container, dev, val, cache)
+            push!(ini_conds, ic)
+            @debug key, ic, val, 0
```
daniel-thom, line 117: Same as above

src/core/update_initial_conditions.jl
jd-lara: This code runs during the simulation set-up. I would prefer that the recorder only runs to keep track of the simulation.

PR #460, opened 2020-06-19
Cb/status ic

claytonpbarrows:
- calculates status initial conditions based on the status field and "ON" variables rather than the "P" varaibles now that we have  a status field
 - So far, I've only tested on a single stage simulation. Before merging, we need to confirm that everything works in a simulation with multiple stages, including one without the "on" variables

### DIFF
--- a/src/core/update_initial_conditions.jl
+++ b/src/core/update_initial_conditions.jl
@@ -13,7 +13,7 @@ function calculate_ic_quantity(
     current_counter = time_cache[:count]
     last_status = time_cache[:status]
     var_status = isapprox(var_value, 0.0, atol = ABSOLUTE_TOLERANCE) ? 0.0 : 1.0
-    @debug last_status, var_status, abs(last_status - var_status)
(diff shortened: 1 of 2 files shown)


src/core/update_initial_conditions.jl
```diff
@@ -136,7 +130,7 @@ function status_init(
         psi_container,
         devices,
         ICKey(DeviceStatus, T),
-        _make_initial_condition_active_power,
+        _make_initial_condition_status,
         _get_status_value,
     )
```
claytonpbarrows, line 135: @jd-lara, I think we need a cache here. This works for a single stage simulation, but with a UC->ED, it fails because it can't find the ON variable in the ED stage when initializing the 2nd UC step.

jd-lara:
Unfortunately, this PR points to an issue when a unit is modeled in unit commitment and also has a Pmin = 0.0. Handling that corner case requires some additional code so it won't break the UC + ED simulation.

PR #462, opened 2020-06-22
Psima 89

Lilyhanig:
documentation for PowerGraphics to live in PowerSimulations

run

cd docs
] activate .
include("make.jl")

to see the documentation, go to index.html in the build folder


docs/make.jl
```diff
@@ -12,8 +34,17 @@ makedocs(
...
+            "Make Stack Plots" => "howto/3.1_make_stack_plots.md",
+            "Make Bar Plots" => "howto/3.2_make_bar_plots.md",
+            "Make Fuel Plots" => "howto/3.3_make_fuel_plots.md",
+            "Make Forecast Plots" => "howto/3.4_make_forecast_plots.md",
+            "Plot Fewer Variables" => "howto/3.5_plot_fewer_variables.md",
+            "Plot Multiple Results" => "howto/3.6_plot_multiple_results.md",
+        ],
         "Simulation Recorder" => "man/simulation_recorder.md",
-        "API" => Any["PowerSimulations" => "api/PowerSimulations.md"],
+        #  "API" => Any["PowerSimulations" => "api/PowerSimulations.md"],
```
jd-lara, line 47: Why is this commented out?

docs/Project.toml
```diff
@@ -1,7 +1,18 @@
 [deps]
+Cbc = "9961bab8-2fa3-5c5a-9d89-47fab24efd76"
```
jd-lara, line 2: Do we need all these dependencies?

docs/make.jl
Lilyhanig: it was throwing a warning, I can uncomment

docs/Project.toml
Lilyhanig: yeah because it actually runs code to make plots... I thought of making a fake results to plot with but it was extremely hacky and better to just run a 5 bus operations problem

PR #463, opened 2020-06-22
Refactor add_variables functions to use common logic

daniel-thom:


src/devices_models/devices/common/add_variable.jl
```diff
@@ -1,3 +1,134 @@
+struct AddVariableInputs
+    variable_names::Vector{String}
```
jd-lara, line 2: Normally we expect only a single variable. We should make a constructor that takes just one variable. Only in the case of Storage and commitment we have more than one variable at a time.

src/devices_models/devices/hydro_generation.jl
```diff
@@ -13,40 +13,26 @@ struct HydroCommitmentReservoirFlow <: AbstractHydroUnitCommitment end
...
-    add_variable(
-        psi_container,
-        devices,
-        variable_name(REACTIVE_POWER, H),
-        false,
-        :nodal_balance_reactive;
-        ub_value = d -> PSY.get_reactivepowerlimits(d).max,
-        lb_value = d -> PSY.get_reactivepowerlimits(d).min,
-        init_value = d -> PSY.get_reactivepower(d),
+function make_reactive_power_add_variable_inputs(::Type{<:PSY.HydroGen}, ::PSIContainer)
```
jd-lara, line 27: these are really long function names, I think we need something more descriptive and obvious.

daniel-thom: I'm open to suggestions.

claytonpbarrows: 1. `mkvarinp_reactive_power`
 2. `mk_reactive_power_var_inp`
 3. `reactive_power_variable_inputs`

I like the ones that would have a consistent prefix for the different applications (1 & 2). We could break from our style guid a bit and abbreviate these processes, so we could make something like `avi` for `the `add_variable_inputs` or `mavi`, giving us:

 4. `make_reactive_power_avi`
 5. `mavi_reactive_power`

daniel-thom: Keep in mind that we already have
```
function make_active_power_constraints_inputs
function make_reactive_power_constraints_inputs
function make_nodal_expression_inputs
```
We can change those if necessary.  I do think that we should follow the same pattern everywhere.

jd-lara: I suggest keeping just 

make_active_power_constraints
make_active_power_variables 

I Think a level 2 user doesn't need to know the whole inputs layer.

daniel-thom: Keep in mind that this function returns an instance of DeviceRangeConstraintInputs.  Here is the context of where it's used:

```
function activepower_constraints!(psi_container, devices, model)
    inputs = make_active_power_constraints_inputs(...)
    device_range_constraints!(psi_container, devices, model, feedforward, inputs)
```

Changing it to 
```
    inputs = make_active_power_constraints(...)
```
might confuse some readers.  Do you still have the same opinion?

claytonpbarrows: maybe just `make_reactive_power_variable_inputs`

src/devices_models/devices/common/add_variable.jl
```diff
@@ -1,3 +1,150 @@
...
+) where {T <: PSY.Component}
+    variable_names = inputs.variable_names
+    binary = inputs.binary
+    expression_name = inputs.expression_name
+    sign = inputs.sign
+    initial_value_func = inputs.initial_value_func
+    lb_value_func = inputs.lb_value_func
+    ub_value_func = inputs.ub_value_func
+
+    if !isnothing(inputs.devices_filter_func)
```
daniel-thom, line 129: Am I interpreting comments correctly such that I should do the following?

```
filter_func = nothing
if isnothing(inputs.devices_filter_func)
    if T <: PSY.Device
        filter_func = x -> PSY.get_available(x)
    end
else
    filter_func = inputs.filter_func
end
```

jd-lara: yes, that's reasonable

PR #464, opened 2020-06-23
Psima 89

Lilyhanig:
add PowerGraphics documentation

deploy docs commented out until PowerGraphics gets released so that PowerGraphics is up to date.


src/core/simulation_results.jl
```diff
@@ -0,0 +1,629 @@
...
+```julia
+sim_output = execute!(simulation)
+results = load_simulation_results(sim_output, "stage_name")
+```
+# Accepted Key Words
+"""
+function load_simulation_results(
+    sim_output::SimulationResultsReference,
+    stage_name::String;
+    kwargs...,
```
jd-lara: Which are the accepted kwargs

src/core/simulation_results.jl
```diff
@@ -0,0 +1,629 @@
...
+```
+# Accepted Key Words
+"""
+
+function load_simulation_results(
+    path::String,
+    stage_name::String,
+    step::Array,
+    variable::Array;
+    kwargs...,
```
jd-lara: same here?

PR #465, opened 2020-06-23
PSIMA-194 Separate slack variables

jd-lara:


src/core/definitions.jl
```diff
@@ -43,7 +43,8 @@ const UNSET_INI_TIME = Dates.DateTime(0)
 # Tolerance of comparisons
 # MIP gap tolerances in most solvers are set to 1e-4
 const ABSOLUTE_TOLERANCE = 1.0e-3
-const SLACK_COST = 1e6
+const SYSTEM_SLACK_COST = 1e6
```
claytonpbarrows, line 46: for consistency with the slack kwargs, suggest `BALANCE_SLACK_COST`

PR #468, opened 2020-06-24
Refactor add variables

daniel-thom:
This now covers all cases of `add_variables!` except for `flow_variables`.  Those currently use differently methodology and don't fit well into this paradigm.


src/devices_models/devices/common/add_variable.jl
```diff
@@ -1,3 +1,134 @@
...
+"""
+Add variables to the PSIContainer for a service.
+"""
+function add_variables!(
+    ::Type{T},
+    psi_container::PSIContainer,
+    service::U,
+    contributing_devices::Vector{V},
+) where {T <: VariableType, U <: PSY.Reserve, V <: PSY.Device}
+    inputs = make_variable_inputs(T, psi_container, service, V)
```
daniel-thom, line 78: There may be some disagreement here.  I am passing the type of the contributing devices, not the devices.  The filter function gets assigned in this function and applied in the next.

src/core/psi_container.jl
```diff
@@ -250,10 +250,16 @@ function decode_symbol(name::Symbol)
...
 
+# TODO: remove once all references are changed
 constraint_name(cons_type, device_type) = encode_symbol(device_type, cons_type)
 constraint_name(cons_type) = encode_symbol(cons_type)
+make_constraint_name(cons_type, device_type) = encode_symbol(device_type, cons_type)
+make_constraint_name(cons_type) = encode_symbol(cons_type)
+# TODO: remove once all references are changed
 variable_name(var_type, device_type) = encode_symbol(device_type, var_type)
 variable_name(var_type) = encode_symbol(var_type)
+make_variable_name(var_type, device_type) = encode_symbol(device_type, var_type)
```
daniel-thom, line 261: I could make global changes in a separate PR.  For now I made changes in files that I was refactoring as part of this PR.

src/devices_models/devices/common/add_variable.jl
jd-lara: This won't work generally since contributing devices aren't from a single ConcreteType. For instance, reserves can be provided by ThermalStandard and HydroDispatch.

src/core/psi_container.jl
jd-lara: We can include these changes in the same PR since they are a general refactor of the variable creation/handling

src/devices_models/devices/common/add_variable.jl
daniel-thom: The type of each contributing device isn't actually used until `add_variable` is called at the end of the sequence, so I don't think it matters.  It is only used for parameterization, so we could remove it altogether.  Unless I'm missing something...

src/core/psi_container.jl
daniel-thom: OK.  I'll wait until we reach agreement on the changes and do this at the end.

src/devices_models/devices/common/add_variable.jl
jd-lara: The issue is that we don't know what V will be. We wouldn't in principle even need to parametrize here.

daniel-thom: Then I'll remove it.

src/core/psi_container.jl
jd-lara: Let's do this in another PR

src/devices_models/devices/common/add_variable.jl
```diff
@@ -1,3 +1,134 @@
+struct AddVariableInputs
```
jd-lara, line 1: How about we rename this VariableInputSpecification? Just an idea to improve naming.

src/devices_models/devices/electric_loads.jl
```diff
@@ -5,46 +5,42 @@ struct InterruptiblePowerLoad <: AbstractControllablePowerLoadFormulation end
...
-    add_variable(
-        psi_container,
-        devices,
-        variable_name(ACTIVE_POWER, L),
-        false,
-        :nodal_balance_active,
-        -1.0;
-        ub_value = x -> PSY.get_maxactivepower(x),
-        lb_value = x -> 0.0,
+function make_variable_inputs(
```
jd-lara, line 8: and we can make this `make_variable_specification`

src/services_models/reserves.jl
```diff
@@ -5,33 +5,30 @@ struct StepwiseCostReserve <: AbstractReservesFormulation end
...
+    service::T,
+) where {T <: PSY.Reserve}
+    inputs = AddVariableInputs(;
+        variable_name = make_variable_name(PSY.get_name(service), T),
+        binary = false,
+        lb_value_func = x -> 0,
+        devices_filter_func = x -> PSY.get_available(x),
     )
-    return
+    return inputs
```
jd-lara, line 19: do we want the return at the top to be consistent other functions?

src/services_models/services_constructor.jl
```diff
@@ -67,7 +67,7 @@ function construct_service!(
         end
 
         #Variables
-        activeservice_variables!(psi_container, service, contributing_devices)
+        add_variables!(ServiceVariable, psi_container, service, contributing_devices)
```
jd-lara, line 70: we need a more specific name here since not all the service variables are the same name. This will ve a reserve variable.

src/services_models/services_constructor.jl
```diff
@@ -99,7 +98,7 @@ function construct_service!(
                 name = PSY.get_name(service),
             )].contributing_devices
         #Variables
-        activeservice_variables!(psi_container, service, contributing_devices)
+        add_variables!(ServiceVariable, psi_container, service, contributing_devices)
```
jd-lara, line 101: This is also a reserve variable

src/services_models/services_constructor.jl
```diff
@@ -129,7 +128,7 @@ function construct_service!(
             #    throw(IS.ConflictingInputsError("All area most have an AGC service assigned in order to model the System's Frequency regulation"))
         end
     end
-    area_mismatch_variables!(psi_container, areas)
+    add_variables!(AreaMismatchVariable, PSY.Area, psi_container, areas)
```
jd-lara, line 131: This is where this model kind of breaks a little bit since AreaMismatchVariable is very specific and unique to AGC models. For now we can leave it, but we need to keep in mind how to avoid proliferation of variable types.

daniel-thom: Would another of the variable types work?  ServiceVariable + PSY.Area may be sufficient for dipatch.

src/services_models/reserves.jl
daniel-thom: Yes.  I'll change it.

src/devices_models/devices/common/add_variable.jl
daniel-thom: Would you do the same for making constraints and nodal expressions?

jd-lara: Yes

daniel-thom: We would have
```
struct NodalExpressionInputSpecification
function make_nodal_expression_specification
struct DeviceRangeConstraintInputSpecification
function make_device_range_constraint_specification
```
I prefer `inputs` over `specification` but don't have a problem with it.  If @claytonpbarrows is onboard with you, I'll make the changes.

claytonpbarrows: Not putting a lot of thought into this I sort of prefer "inputs", but don't have a strong opinion.

claytonpbarrows:
```
struct NodalExpressionSpecification
function make_nodal_expression_specification
struct DeviceRangeConstraintSpecification
function make_device_range_constraint_specification
```

or 

```
struct NodalExpressionInputInputs
function make_nodal_expression_inputs
struct DeviceRangeConstraintInputs
function make_device_range_constraint_inputs
```

claytonpbarrows: Per conversation on the phone and strong input from @jd-lara, go with `Spec`

test/runtests.jl
```diff
@@ -23,8 +23,8 @@ Aqua.test_unbound_args(PowerSimulations)
 Aqua.test_undefined_exports(PowerSimulations)
 #Aqua.test_ambiguities(PowerSimulations)
 
-import PowerSystems.UtilsData: TestData
-download(TestData; branch = "master")
+#import PowerSystems.UtilsData: TestData
+#download(TestData; branch = "master")
```
daniel-thom, line 27: @jd-lara Do we need this?

jd-lara: I am not sure, I don't think so

jd-lara:
@daniel-thom is this ready for review?

daniel-thom:
> @daniel-thom is this ready for review?

Yes.  The building of constraints (including naming) is inconsistent with this.  But we can quickly follow up with a new release next week to address that.

jd-lara:
@daniel-thom ok, I will merge after the tests pass.

PR #469, opened 2020-06-24
Adding PM constraints to PSI container constraints dict

sourabhdalvi:


sourabhdalvi:
> I need to see the additional overhead this brings to the construction process before merging. It the overhead is significant we have to make it an optional thing.

Makes sense, let me know if you want me to test that

jd-lara:
> > I need to see the additional overhead this brings to the construction process before merging. It the overhead is significant we have to make it an optional thing.
> 
> Makes sense, let me know if you want me to test that

The previous message implied that this PR needs the test.

sourabhdalvi:
> > > I need to see the additional overhead this brings to the construction process before merging. It the overhead is significant we have to make it an optional thing.
> > 
> > 
> > Makes sense, let me know if you want me to test that
> 
> The previous message implied that this PR needs the test.

Benchmark Operations Model with addition of network constraint from PM
```bash
BenchmarkTools.Trial: 
  memory estimate:  51.01 MiB
  allocs estimate:  791833
  --------------
  minimum time:     191.121 ms (23.15% GC)
  median time:      205.578 ms (25.15% GC)
  mean time:        217.884 ms (26.89% GC)
  maximum time:     292.239 ms (39.26% GC)
  --------------
  samples:          23
  evals/sample:     1
  ```
  
 Benchmark Operations Model without addition of network constraints from PM
 ```bash
 BenchmarkTools.Trial: 
  memory estimate:  50.99 MiB
  allocs estimate:  791415
  --------------
  minimum time:     183.665 ms (23.36% GC)
  median time:      191.559 ms (25.46% GC)
  mean time:        198.709 ms (27.42% GC)
  maximum time:     240.219 ms (37.69% GC)
  --------------
  samples:          26
  evals/sample:     1
  ```

File: test/test_network_constructors.jl [hunk 1]
```diff
@@ -118,7 +118,7 @@ end
     c_sys14_dc = build_system("c_sys14_dc")
     systems = [c_sys5, c_sys14, c_sys14_dc]
     objfuncs = [GAEVF, GQEVF, GQEVF]
-    constraint_names = [:RateLimit_ub__Line, :RateLimit_lb__Line]
+    constraint_names = [:RateLimit_ub__Line, :RateLimit_lb__Line, :nodal_balance_active]
```
jd-lara, line 121: This is also the name of the expression which is not strictly the same thing.

sourabhdalvi: What name do you suggest? I was just following the existing naming practices in PSI

jd-lara:
```
:Bus_active_power_balance
:Bus_reactive_power_balance
```

sourabhdalvi:
> @sourabhdalvi we can merge when the formatter gets fixed

I'm not sure why formatter failed, its changing `src/services_models/agc.jl`
 file which I haven't changed and it was passing before. Also when I run it locally it doesn't change anything

test/test_network_constructors.jl
```diff
@@ -166,7 +166,12 @@ end
     c_sys14_dc = build_system("c_sys14_dc")
     systems = [c_sys5, c_sys14, c_sys14_dc]
     objfuncs = [GAEVF, GQEVF, GQEVF]
-    constraint_names = [:RateLimitFT__Line, :RateLimitTF__Line]
+    constraint_names = [
+        :RateLimitFT__Line,
+        :RateLimitTF__Line,
+        :Bus_active_power_balance,
+        :Bus_reactive_power_balance,
```
daniel-thom, line 173: Why is free-form text being used here instead of the const strings defined in src/core/definitions.jl?  This practice will make it very hard to debug problems because we won't be able to find where specific items are used.

jd-lara: @claytonpbarrows is there a reason for this?

daniel-thom: Also, constraint names should use the format `name__DeviceType`

jd-lara: agreed.

claytonpbarrows: I think this was never updated from the original PM interface. I'll try to address.

daniel-thom: You might as well call the function `make_constraint_name`, which is new in `dev0-6`.

sourabhdalvi: I'll create another PR to address that

PR #470, opened 2020-06-30
Dev0 6

Lilyhanig:
- cleaned up the show function for IS.Results for cases where the results are verrrry large


jd-lara:
@Lilyhanig please correct the formating

PR #472, opened 2020-07-08
Bug fix in PWL cost implementation

sourabhdalvi:


src/devices_models/devices/common/cost_functions.jl
```diff
@@ -376,7 +376,11 @@ function ps_cost(
     gen_cost = JuMP.GenericAffExpr{Float64, _variable_type(psi_container)}()
     for (t, var) in enumerate(variable)
         if !isnothing(bin)
-            c, pwl_vars = _pwl_cost(psi_container, var, cost_array, bin[t])
+            if typeof(bin) <: ParameterJuMP.ParameterRef
```
daniel-thom, line 379: `if bin isa ParameterJuMP.ParameterRef` is more idiomatic.

sourabhdalvi: I'll make that change, thanks

src/devices_models/devices/common/cost_functions.jl
```diff
@@ -182,7 +182,7 @@ function _pwlgencost_sos(
     psi_container::PSIContainer,
     variable::JV,
     cost_component::Vector{NTuple{2, Float64}},
-    status::Union{Nothing, JV} = nothing,
+    status::Union{Nothing, JuMP.AbstractVariableRef} = nothing,
```
jd-lara, line 185: I don't get why you made this change.

src/devices_models/devices/common/cost_functions.jl
```diff
@@ -376,7 +376,11 @@ function ps_cost(
     gen_cost = JuMP.GenericAffExpr{Float64, _variable_type(psi_container)}()
     for (t, var) in enumerate(variable)
         if !isnothing(bin)
-            c, pwl_vars = _pwl_cost(psi_container, var, cost_array, bin[t])
+            if bin isa ParameterJuMP.ParameterRef
```
jd-lara, line 379: Please add detailed comments in the code about the logic here, I think we will have to do some major refactoring here and knowing the original intention will be a good thing to have.

jd-lara:
@sourabhdalvi I am going to merge since the comments are not critical and dev0-6 needs to be fixed. Please address them later.

PR #474, opened 2020-07-09
Fixing typos in the MultiStart Formulation

sourabhdalvi:


test/test_device_thermal_generation_constructors.jl
```diff
@@ -361,7 +361,7 @@ end
             use_parameters = p,
         )
         construct_device!(op_problem, :Thermal, model)
-        moi_tests(op_problem, p, 528, 0, 192, 108, 192, true)
+        moi_tests(op_problem, p, 528, 0, 238, 60, 192, true)
```
jd-lara, line 364: Why this change? Can you comment on what cause this please?

sourabhdalvi: since I added an `!` to the `make_active_power_constraints_inputs` it was building the wrong constraint

PR #476, opened 2020-07-13
Missing continue in time constraint get_data

sourabhdalvi:


test/test_device_thermal_generation_constructors.jl
```diff
@@ -361,7 +361,7 @@ end
             use_parameters = p,
         )
         construct_device!(op_problem, :Thermal, model)
-        moi_tests(op_problem, p, 528, 0, 192, 108, 192, true)
+        moi_tests(op_problem, p, 528, 0, 188, 108, 192, true)
```
jd-lara, line 364: Can you comment on why this change, please?

jd-lara:
@sourabhdalvi it seems a test is still failing.

test/test_device_thermal_generation_constructors.jl
sourabhdalvi: When parameters are not used, the pglib constraint function creates spares constraint  and when the continue is added back I guess its skipping one device so drop in the constraint count

PR #483, opened 2020-07-17
Refactor range constraints

daniel-thom:
This PR refactors the way in which range constraints are constructed.  Other constraint types are unchanged.


jd-lara:
@daniel-thom overall looks good to me. I might need to get used to some of the changes in the definitions of the input but this is major improvement.

PR #484, opened 2020-07-20
[WIP] Add supplemental reserve handling.

andrewrosemberg:
Adds supplemental reserve handling.

This PR is related to the PowerSystems.jl [PR](https://github.com/NREL-SIIP/PowerSystems.jl/pull/547).


jd-lara:
x-ref https://github.com/NREL-SIIP/PowerSystems.jl/pull/547

jd-lara:
@andrewrosemberg I checked the PR but it seems that you are creating a separate reserve product. Would the same be possible by just creating another reserve category in PSY and then formulating it as a range reserve?

andrewrosemberg:
@jd-lara, sorry, I'm still getting used to the structure here, so correct me if I misunderstood your suggestion. 

> @andrewrosemberg I checked the PR but it seems that you are creating a separate reserve product.

Do you mean that because of the fact that I call both `device_semicontinuousrange` and `device_range` two requirement constraints will be created? I thought this would only duplicate the range constraint (maximum and minimum dispatch) since it only changes the output of `construct_device!` and not the `construct_service!`.

> Would the same be possible by just creating another reserve category in PSY and then formulating it as a range reserve?

I don't see how since any new reserve is considered in the `device_semicontinuousrange`. Just to remind, this supplemental reserve can be contributed both by committed and non-committed generators, so considering it on `device_semicontinuousrange` (which limits generation with respect to the commitment binary variable) would not allow non-committed generators to contribute. 

Therefore, when we have supplemental reserves, we need both:
* The generation limit constraint which takes into account the binary variables (for the dispatch generation + normal reserves): AFAIK `device_range`.
*  The generation limit constraint which ignores the binary variables (to limit our new reserve): AFAIK `device_semicontinuousrange`.

src/devices_models/devices/thermal_generation.jl
```diff
@@ -150,6 +150,19 @@ function activepower_constraints!(
...
+        range_data = DeviceRange(name, limits)
+        add_device_services!(range_data, d, model; add_supplemental=true)
+        constraint_data[ix] = range_data
+    end
+    device_range(
+        psi_container,
+        constraint_data,
+        constraint_name(ACTIVE_RANGE, T),
+        variable_name(ACTIVE_POWER, T),
+    )
```
nickrobinson251, line 165: For my edification, can you explain what this function call is doing?

Given its name suggest it's _not_ a mutating function, and we don't return the result, it looks as if it does nothing... but the same can be said of the exisiting call to `device_semicontinuousrange` above.

Looking at the code for `device_range` https://github.com/NREL-SIIP/PowerSimulations.jl/blob/d949d73564b1a91fed5d9291d28fab122e9f5cf9/src/devices_models/devices/common/range_constraint.jl#L43-L44
it seems it actually _is_ a mutating function (which would explain its. usage here) - is that right? If so, why is it not named `device_range!` with the trailing `!` to indicate it mutates the `PSIContainer`?

raphaelsaavedra: You're right -- see https://github.com/NREL-SIIP/PowerSimulations.jl/issues/485 for clarification. The naming is being changed in the next release to avoid confusion.

andrewrosemberg:
We are trying to have both:
```
        con_ub[data.name, t] = JuMP.@constraint(
            psi_container.JuMPmodel,
            expression_ub <= data.limits.max * varbin[data.name, t]
        )
         con_lb[data.name, t] = JuMP.@constraint(
            psi_container.JuMPmodel,
            expression_lb >= data.limits.min * varbin[data.name, t]
        )
```
where the `expression_x`would contain generation and normal reserves (without supplemental reserves);
and
```
       con_ub[data.name, t] = JuMP.@constraint(
            psi_container.JuMPmodel,
            expression_ub <= data.limits.max
        )
       con_lb[data.name, t] = JuMP.@constraint(
            psi_container.JuMPmodel,
            expression_lb >= data.limits.min
        )
```
where the `expression_x`would contain generation and all reserves (with supplemental reserves).
That's why I thought I would need to call both `device_semicontinuousrange ` and `device_range `.
However, This might not be the best way to implement this. 

Perhaps one alternative way would be to have two pairs of `additional_terms_ub`and `additional_terms_lb`, where only one pair would contain the supplemental reserve term. So we would only need to modify the `device_semicontinuousrange`.
What are your thoughts?

src/devices_models/devices/thermal_generation.jl
nickrobinson251: Oh you're ahead of the game - thank you!

andrewrosemberg:
@jd-lara, I refactored these changes accordingly to what we discussed. Please let me know if this was what you expected!

nickrobinson251:
What's the plan for tests?

andrewrosemberg:
> What's the plan for tests?

1. Add tests to the ones in `test_services_constructor.jl`
2. Add similar tests to the ones in `test_device_thermal_generation_constructors.jl`

Not sure what's the best practice here, but that's what I'm looking into now.

src/devices_models/device_constructors/supplemental_thermalgeneration_constructor.jl
```diff
@@ -0,0 +1,224 @@
+"""
+This function creates the model for a full themal dispatch formulation depending on combination of devices, device_formulation and system_formulation
```
raphaelsaavedra, line 2:
```suggestion
This function creates the model for a full thermal dispatch formulation depending on combination of devices, device_formulation and system_formulation
```

src/devices_models/device_constructors/supplemental_thermalgeneration_constructor.jl
```diff
@@ -0,0 +1,224 @@
...
+        Type{S}
+    )
+    devices = get_available_components(T, sys)
+    offline_activepower_constraints!(psi_container, devices, model, S, get_feedforward(model))
+    
+    return
+end
+
+"""
+This function creates the model for a full themal dispatch formulation depending on combination of devices, device_formulation and system_formulation
```
raphaelsaavedra, line 23:
```suggestion
This function creates the model for a full thermal dispatch formulation depending on combination of devices, device_formulation and system_formulation
```

src/devices_models/device_constructors/supplemental_thermalgeneration_constructor.jl
```diff
@@ -0,0 +1,224 @@
...
+        DeviceModel{T, ThermalStandardUnitCommitment}(),
+        Type{S}
+    )
+    devices = get_available_components(T, sys)
+    offline_activepower_constraints!(psi_container, devices, model, S, get_feedforward(model))
+
+    return
+end
+"""
+This function creates the model for a full themal dispatch formulation depending on combination of devices, device_formulation and system_formulation
```
raphaelsaavedra, line 43:
```suggestion
This function creates the model for a full thermal dispatch formulation depending on combination of devices, device_formulation and system_formulation
```

src/devices_models/device_constructors/supplemental_thermalgeneration_constructor.jl
```diff
@@ -0,0 +1,224 @@
...
+        Type{S}
+    )
+    devices = get_available_components(T, sys)
+    offline_activepower_constraints!(psi_container, devices, model, S, get_feedforward(model))
+
+    return
+end
+
+"""
+This function creates the model for a full themal dispatch formulation depending on combination of devices, device_formulation and system_formulation
```
raphaelsaavedra, line 64:
```suggestion
This function creates the model for a full thermal dispatch formulation depending on combination of devices, device_formulation and system_formulation
```

src/devices_models/device_constructors/supplemental_thermalgeneration_constructor.jl
```diff
@@ -0,0 +1,224 @@
...
```
raphaelsaavedra, line 106:
```suggestion
This function creates the model for a full thermal dispatch formulation depending on combination of devices, device_formulation and system_formulation
```

src/devices_models/device_constructors/supplemental_thermalgeneration_constructor.jl
```diff
@@ -0,0 +1,224 @@
...
```
raphaelsaavedra, line 127:
```suggestion
This function creates the model for a full thermal dispatch formulation depending on combination of devices, device_formulation and system_formulation
```

src/devices_models/device_constructors/supplemental_thermalgeneration_constructor.jl
```diff
@@ -0,0 +1,224 @@
...
```
raphaelsaavedra, line 152:
```suggestion
This function creates the model for a full thermal dispatch formulation depending on combination of devices, device_formulation and system_formulation
```

src/devices_models/device_constructors/supplemental_thermalgeneration_constructor.jl
```diff
@@ -0,0 +1,224 @@
...
```
raphaelsaavedra, line 202:
```suggestion
This function creates the model for a full thermal dispatch formulation depending on combination of devices, device_formulation and system_formulation
```

andrewrosemberg:
Closing it in favor of a future non-spinning PR

PR #486, opened 2020-07-21
Cb/forecast multipliers

claytonpbarrows:


src/devices_models/devices/hydro_generation.jl
```diff
@@ -350,7 +350,7 @@ function energy_balance_constraint!(
     for (ix, d) in enumerate(devices)
         ts_vector = get_time_series(psi_container, d, forecast_label)
         constraint_info =
-            DeviceTimeSeriesConstraintInfo(d, x -> PSY.get_rating(x), ts_vector)
+            DeviceTimeSeriesConstraintInfo(d, x -> PSY.get_max_active_power(x), ts_vector)
```
jd-lara, line 353: Isn't here a `get_inflow` functions. Also, add a TODO about fixing Hydro

jd-lara: I will fix in in dev0-6

PR #487, opened 2020-07-22
Add StaticGroupReserve Handling

andrewrosemberg:
Adds the `StaticGroupReserve` handling related to reserve requirements that group multiple reserve types (as discussed in this [issue](https://github.com/NREL-SIIP/PowerSimulations.jl/issues/479)).

x-ref [NREL-SIIP/PowerSystems.jl#554](https://github.com/NREL-SIIP/PowerSystems.jl/pull/554)


src/services_models/StaticGroupReserve.jl
```diff
@@ -0,0 +1,94 @@
...
+#     return
+# end
+
+function include_service!(
+    constraint_data::DeviceRange,
+    services,
+    ::ServiceModel{SR, <:AbstractReservesFormulation},
+) where {SR <: PSY.StaticGroupReserve}
+    return
+end
```
jd-lara, line 94: Nit: missing end of file

andrewrosemberg: I think I have solved it, but I'm not sure haha never saw this error.

andrewrosemberg:
> Every service in PSI is expressed as a pair ServiceModel{ModelType, ModelFormulation} clearly, in this case, this is an obvious "formulation" but we need to add those too.

I thought of using `RangeReserve`, but I guess this might not be the best solution.

andrewrosemberg:
> This looks pretty straight forward. I will suggest adding more docstrings to communicate the formulation better please. Every service in PSI is expressed as a pair `ServiceModel{ModelType, ModelFormulation}` clearly, in this case, this is an obvious "formulation" but we need to add those too.

Created a formulation called `GroupReserve` which will be able to cover Static and eventually variable group reserve. @jd-lara, is this what you expected? 

I will start focusing on documenting the code soon.

jd-lara:
the PR looks ok to me. But we should merge https://github.com/NREL-SIIP/PowerSystems.jl/pull/554 first so that the tests pass and also the PR's needs to be rebased. 

@claytonpbarrows we need to agree on the issue of service aggregations. Any thoughts?

src/services_models/StaticGroupReserve.jl
```diff
@@ -0,0 +1,71 @@
...
+############################### Reserve Variables` #########################################
+"""
+This function checks if the variables for reserves were created
+"""
+function check_activeservice_variables(
+    psi_container::PSIContainer,
+    contributing_services::Vector{<:PSY.Service},
+)
+    for service in contributing_services
+        # Should pop and error if no such variable exists
```
raphaelsaavedra, line 11:
```suggestion
        # Should pop an error if no such variable exists
```

src/services_models/StaticGroupReserve.jl
```diff
@@ -0,0 +1,71 @@
...
+        reserve_variable = get_variable(psi_container, variable_name(PSY.get_name(service), typeof(service)))
+    end
+    return
+end
+
+################################## Reserve Requirement Constraint ##########################
+"""
+This function creates teh requirement constraint that will be attained by the apropriate services
```
raphaelsaavedra, line 19:
```suggestion
This function creates the requirement constraint that will be attained by the apropriate services
```

src/services_models/services_constructor.jl
```diff
@@ -72,3 +72,44 @@ function construct_service!(
     end
     return
 end
+
+```
+Constructs a service for for StaticGroupReserve.
```
raphaelsaavedra, line 77:
```suggestion
Constructs a service for StaticGroupReserve.
```

src/services_models/StaticGroupReserve.jl
claytonpbarrows, line 19: nit: `teh` typo

src/services_models/services_constructor.jl
```diff
@@ -72,3 +72,44 @@ function construct_service!(
     end
     return
 end
+
+```
```
raphaelsaavedra, line 76: This should be `"""` rather than ``` right? (same for the line below)

andrewrosemberg:
@jd-lara Rebase also completed here!

andrewrosemberg:
Just did the rebase for the tag 0.6.2. 
Will wait for [this](https://github.com/NREL-SIIP/PowerSystems.jl/pull/571) PR to merge and get the new tag before rebasing again.

andrewrosemberg:
Tests are "ready", waiting for the bump in the PSY version.

andrewrosemberg:
Removed the WIP, ready for review

src/services_models/services_constructor.jl
```diff
@@ -20,7 +20,20 @@ function construct_services!(
 )
     isempty(services_template) && return
     incompatible_device_types = get_incompatible_devices(devices_template)
-    for service_model in values(services_template)
+    # group service needs to be constructed last
```
andrewrosemberg, line 23: I admit this could be much nicer and more efficient, but I'm not sure how. Open for suggestions.

test/test_services_constructor.jl
```diff
@@ -172,3 +172,42 @@ end
     # These values might change as the AGC model is refined
     moi_tests(agc_problem, false, 720, 0, 480, 0, 384, false)
 end
+
+@testset "Test GroupReserve from Thermal Dispatch" begin
```
andrewrosemberg, line 176: Not sure if this is the intended way for these tests, please let me know and I will correct it.

andrewrosemberg:
Open for suggestions of more tests.

src/services_models/GroupReserve.jl
```diff
@@ -0,0 +1,88 @@
...
+    time_steps = model_time_steps(psi_container)
+    name = PSY.get_name(service)
+    constraint = get_constraint(psi_container, make_constraint_name(REQUIREMENT, SR))
+    use_slacks = get_services_slack_variables(psi_container.settings)
+    reserve_variables = [
+        get_variable(psi_container, PSY.get_name(r), typeof(r))
+        for r in contributing_services
+    ]
+
+    if use_forecast_data && :forecast ∈ fieldnames(typeof(service))
```
nickrobinson251, line 41: may i ask about the use of `:forecast ∈ fieldnames(...)` in this PR? What is the role it is playing?
is it being used as a proxy for some train like "has forecasts"? I wonder if it might just be better to have a function for that? 

Also, AFAICT `fieldnames` is not used in this way anywhere else in PSim, and usually field names are not considered public API (so given `Service` comes from PSys should perhaps not be relied upon here?)

andrewrosemberg: Good point! just changed to has_forecasts.

src/services_models/services_constructor.jl
nickrobinson251: an option would be to "leave" the group service to last. Is there only ever one service with `formulation == GroupReserve`? If so, something like this is an alternative:

```julia
    function _build!(service_model)
        @debug "Building $(service_model.service_type) with $(service_model.formulation) formulation"
        services = service_model.service_type[]
        if validate_services!(
            service_model.service_type,
            services,
            incompatible_device_types,
            sys,
        )
            construct_service!(
                psi_container,
                services,
                sys,
                service_model,
                devices_template,
                incompatible_device_types,
            )
        end
    end    
    groupservice = nothing
    for (key, service_model) in service_models
        if service_model === GroupReserve  # group service needs to be constructed last
            groupservice = key
            continue
        end
        build!(service_model)
    end
    groupservice === nothing || _build!(service_models[groupservice])

Comment (andrewrosemberg) [reply src/services_models/services_constructor.jl:23]: There can be more than one service but currently not more than one template. I like this alternative! what do you guys think? 
From what I see, there are some `_build!` functions in PSI, Nick is this similarity intentional?

Comment (nickrobinson251) [reply src/services_models/services_constructor.jl:23]: cool :) 

> From what I see, there are some _build! functions in PSI, Nick is this similarity intentional?

No, i did not know about these functions. It seems that [one is superficially similar](https://github.com/NREL-SIIP/PowerSimulations.jl/blob/16e1b453ed21e714db58065e03601af4405b2e54/src/core/operations_problem.jl#L393) but does much more. `_build!` was just the first name that came to mind - feel free to go with something else.

Also, to be clear, the proposal is an internal function (actually a closure) _inside_ `construct_services!` - closing over e.g. `psi_container` among other things. It literally just fulls the contents of the `for`-loop out, to avoid duplicating the code. 

so the `construct_services!` function would be structured  like
```julia
funtion construct_services!(args...)
    ...
    function _build!(service_model)
       ...
    end
    for ....
        _build!(...)
    end
    _build!(...)
end
```

src/services_models/GroupReserve.jl
```diff
@@ -0,0 +1,88 @@
...
+                resource_expression =
+                    sum(
+                        sum(reserve_variable[:, t])
+                        for reserve_variable in reserve_variables
+                    ) + slack_vars[t]
+            else
+                resource_expression = sum(
+                    sum(reserve_variable[:, t]) for reserve_variable in reserve_variables
+                )
+            end
```
nickrobinson251, line 71:
```suggestion
            param[name, t] = PJ.add_parameter(psi_container.JuMPmodel, ts_vector[t])
            resource_expression = sum(
                sum(reserve_variable[:, t]) for reserve_variable in reserve_variables
            )
            if use_slacks
                resource_expression += slack_vars[t]
            end     
```
also you could use `mapreduce` rather than `sum(sum(`, but i'm not sure it's clearer:  `mapreduce(v -> sum(v[:, t]), +, reserve_variables)`

src/services_models/services_constructor.jl
```diff
@@ -169,3 +182,50 @@ function construct_service!(
...
+    names = (PSY.get_name(s) for s in services)
+
+    if model_has_parameters(psi_container)
+        container = add_param_container!(
+            psi_container,
+            UpdateRef{SR}("service_requirement", "get_requirement"),
+            names,
+            time_steps,
+        )
+        get_parameter_array(container)
```
nickrobinson251, line 207: what does this line do? it looks like it returns an array that's not assigned/used anywhere?

src/services_models/GroupReserve.jl
```diff
@@ -0,0 +1,88 @@
...
+This function checks if the variables for reserves were created
+"""
+function check_activeservice_variables(
+    psi_container::PSIContainer,
+    contributing_services::Vector{<:PSY.Service},
+)
+    for service in contributing_services
+        # Should pop an error if no such variable exists
+        reserve_variable =
+            get_variable(psi_container, PSY.get_name(service), typeof(service))
```
nickrobinson251, line 13: No need to assign this to a variable if it's not used
```suggestion
        get_variable(psi_container, PSY.get_name(service), typeof(service))
```

Will this throw an informative exception if the variables don't exist?

test/test_services_constructor.jl
nickrobinson251: can you explain the tests to me?
Am i right in saying that all the code here is setup, then all the actual tests are in [`moi_tests`](https://github.com/NREL-SIIP/PowerSimulations.jl/blob/3816aae989a3c4da286b85fd79e081f7bb369f02/test/test_utils/model_checks.jl#L4)?
i'm not yet familiar with the tests in this package. AFAICT this matches the tests in this file. 👍  just wonder if there are any additional tests, e.g. unit tests for some of the functions added here, that can/should also be added?

src/services_models/services_constructor.jl
jd-lara: Normally we don't use _build functions or device constructors. We have reserved those for building the simulations/operations structs. We try to keep the constructors pretty verbose to resemble code from AML's and make it easier power systems modelers to "read" the models.

test/test_services_constructor.jl
jd-lara: MOI tests are meant to do the testing that makes sure the correct number of constraints and variables exist in the container. It also checks that the model uses parameters adequately.

src/services_models/GroupReserve.jl
jd-lara: @andrew please check the number of sum operations in JuMP we know that having too many sum operations instead of `add_to_expression` leads to performances issues.

src/services_models/GroupReserve.jl
jd-lara: @daniel-thom did we implement a cleaner getter function that doesn't require calling `get_name` inside?

src/services_models/services_constructor.jl
```diff
@@ -169,3 +182,50 @@ function construct_service!(
...
+function construct_service!(
+    psi_container::PSIContainer,
+    services::Vector{SR},
+    ::PSY.System,
+    model::ServiceModel{SR, GroupReserve},
+    ::Dict{Symbol, DeviceModel},
+    ::Vector{<:DataType},
+) where {SR <: PSY.StaticReserveGroup}
+    time_steps = model_time_steps(psi_container)
+    names = (PSY.get_name(s) for s in services)
```
daniel-thom, line 198: You have following the same pattern as the existing code.  However, I think this should be
```
names = [PSY.get_name(s) for s in services]
```
This becomes an axis for a JuMP.DenseAxisArray and I think it would be much more performant to give it an array of strings instead of a generator expression (and make the array more readable).  I actually raised this internally yesterday.

daniel-thom: @jd-lara Actually, performance doesn't really matter.  I looked at the source and JuMP builds a dictionary out of whatever you pass in for the axes.  But they do store the axes.

I think it's better to change our code to use arrays for two reasons:
1. Readability:  show an array of names and not the entire generator expression
2. I don't think we want the DenseAxisArrays to hold references to our PSY components for their entire lifetime.

src/services_models/GroupReserve.jl
andrewrosemberg: @daniel-thom I believe it is `ActiveServiceVariable`? just changed to it. Please let me know if that's correct.

andrewrosemberg: I ended up not using get_variable in favor of a better error.

src/services_models/services_constructor.jl
andrewrosemberg: Not sure haha I copied from the `construct_service` for `RangeReserve`. @daniel-thom, can you help us here?

src/devices_models/device_constructors/common/constructor_validations.jl
```diff
@@ -32,3 +32,26 @@ function validate_services!(
...
+
+    for s in services_
+        contributing_services = PSY.get_contributing_services(s)
+        if isempty(contributing_services)
+            @warn("The contributing services for group service $(PSY.get_name(service)) is empty, consider removing the group service from the system")
+        else
+            push!(services, s)
+        end
+    end
+    return true
```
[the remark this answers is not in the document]
andrewrosemberg: I agree with you, but that creates a weird behavior for when there are only a few have nonempty `contributing_services `, and the ones with empty `contributing_services` still get created. Therefore, I would add in addition to your suggestion a check in `construct_service` itself. What do you think?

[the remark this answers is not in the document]
andrewrosemberg: Should we do the same for the `validate_services` of a normal service?

src/services_models/services_constructor.jl
andrewrosemberg: I changed it for `StaticReserveGroup` :)

src/services_models/services_constructor.jl
nickrobinson251: sorry, i'm not sure i understand the comment. to clarify, i'm not suggesting we use any different functionality than is already used, i'm just proposing a reorgnaising on some code inside this function

src/services_models/GroupReserve.jl
andrewrosemberg: my tests gave a 0 `operator_counter ` for both ways (`sum` and `add_to_expression`) but I changed it to `add_to_expression` anyway. Is it suppose to be zero? I'm confused about what this means.

src/services_models/services_constructor.jl
andrewrosemberg: Just implemented the proposed change. @jd-lara please let me know if that's ok.

src/services_models/GroupReserve.jl
```diff
@@ -0,0 +1,92 @@
...
+This function checks if the variables for reserves were created
+"""
+function check_activeservice_variables(
+    psi_container::PSIContainer,
+    contributing_services::Vector{<:PSY.Service},
+)
+    for service in contributing_services
+        name = PSY.get_name(service)
+        var =
+            get(psi_container.variables, make_variable_name(name, typeof(service)), nothing)
```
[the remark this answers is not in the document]
andrewrosemberg: I changed the error to mention that it is a "contributing service" to help the user see where it came from. Do you guys think that this is a good practice? Do you have a suggestion on how to use an accessor function here?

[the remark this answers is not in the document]
daniel-thom: That could be helpful, but the stacktrace for the exception will include this function/line, so we can figure it out.  I understand that that is not as obvious as a simple error message like this.  If you prefer having your own log message then you could catch the exception, log the error, and then rethrow.

Here is the accessor function.
```
get_variable(psi_container, name, typeof(service))
```

@jd-lara We need something documented in the code about how to create and retrieve variables.  I get confused about how some names are based on the component type and some include component names.  It will take a while to make the entire code base consistent, but at least new code can follow the guidance.

src/services_models/GroupReserve.jl
jd-lara: ok, if the operator counter is 0 we are ok. The difference is in JuMP the use of AffineExpression can be destructive and re-allocate memory depending on how it is done.

src/devices_models/device_constructors/common/constructor_validations.jl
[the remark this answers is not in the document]
jd-lara: @andrewrosemberg Validate services should take care of this.

src/services_models/GroupReserve.jl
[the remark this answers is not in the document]
jd-lara: @daniel-thom correct we have some legacy code about the creation and retrieval of variables that needs consistency. @andrewrosemberg, the accessor Dan mentioned is the correct one to use in this context.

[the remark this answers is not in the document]
andrewrosemberg: fair enough, yeah the stack trace already does that.  Just changed back to `get_variable`!

andrewrosemberg:
I believe all comments were addressed. Please let me know if something else needs revising.

src/services_models/services_constructor.jl
```diff
@@ -39,6 +40,17 @@ function construct_services!(
...
+    groupservice = nothing
+
+    for (key, service_model) in services_template
+        if service_model.formulation === GroupReserve  # group service needs to be constructed last
+            groupservice = key
+            continue
+        end
+        _construct_valid_services!(service_model)
+    end
+    groupservice === nothing || _construct_valid_services!(services_template[groupservice])
```
daniel-thom, line 53: Is this the standard Julian way of writing

```
if !isnothing(groupservice)
    _construct_valid_services!(services_template[groupservice])
end
```
?

andrewrosemberg: I believe so. @nickrobinson251 ?

nickrobinson251: yeah, short circuit "and" `&&` and "or" `||` are used this way a lot in Julia. 
But it is exactly equivalent to ```julia
if groupservice !== nothing
    _construct_valid_services!(services_template[groupservice])
end
```

Similar to how `a ? x : y` is used instead of `if a; x; else y; end`

However `a === nothing` is sometimes recommended over `isnothing(a)` for [some reason to do with the compiler].

daniel-thom: I can get used to it if it's standard practice.

PR #489, opened 2020-07-24
Create simplified get_variable method

daniel-thom:


src/core/constraints.jl
```diff
@@ -52,18 +52,6 @@ const NODAL_BALANCE_REACTIVE = "nodal_balance_reactive"
 
 abstract type ConstraintType end
 
-function make_constraint_name(::Type{T}) where {T <: ConstraintType}
```
daniel-thom, line 55: Dheepak convinced me that I shouldn't use this pattern any more.  I still think it is a good pattern that makes up for a useful feature that Julia lacks.  However, we should use `Julian` patterns.

PR #490, opened 2020-07-27
ramp constraint refactor

### LINKED ISSUE #458: Ramp down infeasibility in simulations
In a UC + ED simulation, if the UC turns off the generator the initial condition of the ED keeps the result from the previous ED run. As a result if the ramp from min to off violates the 5-minute ramp limit, this leads to an infeasible result. 

The Ramp Initial condition for variable P needs to consider variable ON from the previous step.

jd-lara:
closes #458 

@daniel-thom this PR required some sub-optimal code in the calculation of the initial conditions to fix #458. This code will have to be refactored once the data store is implemented.


src/core/update_initial_conditions.jl
```diff
@@ -152,22 +182,8 @@ function output_init(
...
-    psi_container::PSIContainer,
-    devices::IS.FlattenIteratorWrapper{PSY.ThermalMultiStart},
-)
-    _make_initial_conditions!(
-        psi_container,
-        devices,
-        ICKey(DevicePower, PSY.ThermalMultiStart),
-        _make_initial_condition_active_power,
         _get_active_power_output_above_min_value,
+        #TimeStatusChange,
```
daniel-thom, line 186: Delete?

src/core/update_initial_conditions.jl
```diff
@@ -95,33 +138,35 @@ function _make_initial_conditions!(
...
             IS.get_uuid(dev) in ic_devices && continue
-            @debug "Setting $(key.ic_type) initial conditions for the status device $(PSY.get_name(dev)) based on system data"
-            val = get_val_func(dev, key)
+            @debug "Setting $(key.ic_type) initial conditions device $(PSY.get_name(dev)) based on system data"
+            val_ = get_val_func(dev, key)
+            val = parameters ? PJ.add_parameter(psi_container.JuMPmodel, val_) : val_
             ic = make_ic_func(ic_container, dev, val, cache)
             push!(ini_conds, ic)
-            @debug "set initial condition" key ic val
+            @debug "set initial condition" key ic val_
```
daniel-thom, line 161: You probably want this to be `val`, right?

src/core/update_initial_conditions.jl
```diff
@@ -41,39 +43,80 @@ function calculate_ic_quantity(
...
+            current_status = isapprox(series[current], 1.0; atol = ABSOLUTE_TOLERANCE)
+            previous_status = isapprox(series[current - 1], 1.0; atol = ABSOLUTE_TOLERANCE)
+            status_change_to_on = current_status && !previous_status
+            status_change_to_off = !current_status && previous_status
+            status_remains_on = current_status && previous_status
+            status_remains_off = !current_status && !previous_status
+        else
+            status_remains_on = true
+        end
+        time_cache[:current] += 1
```
daniel-thom, line 88: This will cause an `UndefVarError` exception if `min_power == 0`.

src/core/update_initial_conditions.jl
```diff
@@ -340,8 +382,13 @@ function _get_active_power_output_value(device, key)
 end
 
 function _get_active_power_output_above_min_value(device, key)
-    return PSY.get_status(device) ?
-           PSY.get_active_power(device) - PSY.get_active_power_limits(device).min : 0.0
+    if !PSY.get_status(device)
+        return 0.0
+    end
+    power_above_min = PSY.get_active_power(device) - PSY.get_active_power_limits(device).min
+    PSY.get_name(device)
```
daniel-thom, line 389: Stale line

src/core/update_initial_conditions.jl
jd-lara: I want val_ because if the setting is to have parameters this will produce a `param_noname` entry instead of the value

PR #492, opened 2020-07-29
Psima 89 take2

Lilyhanig:
- all sorts of power graphics documentation


docs/Project.toml
```diff
@@ -1,7 +1,19 @@
 [deps]
+CSV = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
+Cbc = "9961bab8-2fa3-5c5a-9d89-47fab24efd76"
+DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 DocumenterTools = "35a29f4d-8980-5a13-9543-d66fff28ecb8"
+GLPK = "60bf3e95-4087-53dc-ae20-288a0d20c6a6"
+InfrastructureSystems = "2cd47ed4-ca9b-11e9-27f2-ab636a7671f1"
```
jd-lara, line 8: Do you need to add plots if PowerGraphics is already a dependency?

docs/Project.toml
```diff
@@ -1,7 +1,19 @@
...
+Literate = "98b081ad-f1c9-55d3-8b20-4c87d4299306"
+ORCA = "47be7bcc-f1a6-5447-8b36-7eeeff7534fd"
+PlotlyJS = "f0f68f2c-4968-5e81-91da-67840de0976a"
```
jd-lara, line 11: Same for PlotlyJS?

docs/src/PowerGraphics/3.0_set_up_plots.jl
```diff
@@ -0,0 +1,32 @@
+# # Getting Started with Power Graphics
+
+# This uses the plotting package for SIIP [PowerGraphics](http://github.com/nrel-siip/PowerGraphics.jl.git)
+
+# Start by creating a results object using [PowerSimulations](http://github.com/nrel-siip/PowerSimulations.jl.git)
+
+# Set up Power Graphics:
+# ```julia
+# import PowerGraphics
```
jd-lara, line 9: using PowerGraphics. Why do you call plots?

docs/Project.toml
Lilyhanig: probably not...

jd-lara:
@Lilyhanig try to change this https://github.com/Lilyhanig/PowerSimulations.jl/blob/31a667d3f362676ef9aabde524efbd5f32b74110/.github/workflows/docs.yml#L16 

to 1.4

PR #494, opened 2020-08-01
Empty contributing devices case

### LINKED ISSUE #493: Simulation fails when services' contributing devices is empty
If a system in the model has empty contributing devices, the simulation fails.

jd-lara:
closes #493

### DIFF
--- a/src/services_models/services_constructor.jl
+++ b/src/services_models/services_constructor.jl
@@ -30,8 +30,12 @@ function construct_service!(
 ) where {SR <: PSY.Reserve}
     services_mapping = PSY.get_contributing_device_mapping(sys)
     time_steps = model_time_steps(psi_container)
-    names = (PSY.get_name(s) for s in services)
-
+    names = [
+        key.name
+        for
+        (key, val) in services_mapping if
+        key.type == SR && !isempty(val.contributing_devices)
+    ]
     incompatible_device_types = Vector{DataType}()
     for model in values(devices_template)
         formulation = get_formulation(model)
@@ -62,15 +66,15 @@ function construct_service!(
 
     for service in services
         contributing_devices =
-            services_mapping[(
-                type = typeof(service),
-                name = PSY.get_name(service),
-            )].contributing_devices
+            services_mapping[(type = SR, name = PSY.get_name(service))].contributing_devices
         if !isempty(incompatible_device_types)
             contributing_devices =
                 [d for d in contributing_devices if typeof(d) ∉ incompatible_device_types]
         end
-
+        if isempty(contributing_devices)
+            @warn("The contributing devices for service $(PSY.get_name(service)) is empty, cinsider removing the service from the system")
+            continue
+        end
         #Variables
         add_variables!(ActiveServiceVariable, psi_container, service, contributing_devices)
         # Constraints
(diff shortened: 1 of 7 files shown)


src/services_models/services_constructor.jl
```diff
@@ -30,8 +30,12 @@ function construct_service!(
 ) where {SR <: PSY.Reserve}
     services_mapping = PSY.get_contributing_device_mapping(sys)
     time_steps = model_time_steps(psi_container)
-    names = (PSY.get_name(s) for s in services)
-
+    names = [
```
daniel-thom, line 33: I understand why these changes fix the problem.  However, this is a non-trivial amount of logic that I suspect other users will have to replicate and possibly forget or mess up when they make similar functions.  Is that true?  If so, would it be better to filter out these services in the place where this function is called?

jd-lara: That's why I suggested to not create the entry in the output of `PSY.get_contributing_device_mapping`

src/services_models/services_constructor.jl
```diff
@@ -23,15 +23,14 @@ end
...
+    services::Vector{SR},
     sys::PSY.System,
     model::ServiceModel{SR, RangeReserve},
     devices_template::Dict{Symbol, DeviceModel},
 ) where {SR <: PSY.Reserve}
     services_mapping = PSY.get_contributing_device_mapping(sys)
     time_steps = model_time_steps(psi_container)
-    names = (PSY.get_name(s) for s in services)
-
+    names = [PSY.get_name(s) for s in services]
```
daniel-thom, line 33: @jd-lara Does the content of `names` need to be consistent with the services you add variables and constraints for?  You are calling add_param_container and add_cons_container with names that could later be skipped.

jd-lara: No, the validate services already made those checks and only adds the services that are valid.

src/services_models/services_constructor.jl
```diff
@@ -62,15 +61,15 @@ function construct_service!(
...
-            )].contributing_devices
+            services_mapping[(type = SR, name = PSY.get_name(service))].contributing_devices
         if !isempty(incompatible_device_types)
             contributing_devices =
                 [d for d in contributing_devices if typeof(d) ∉ incompatible_device_types]
         end
-
+        if isempty(contributing_devices)
+            @warn("The contributing devices for service $(PSY.get_name(service)) is empty, cinsider removing the service from the system")
+            continue
```
daniel-thom, line 71: This service name was added above in the `names` variable.  I was asking if this inconsistency is a problem.  The validate function did not check this condition.

Also: s/cinsider/consider/

src/services_models/services_constructor.jl
```diff
@@ -7,8 +7,8 @@ function construct_services!(
     isempty(services_template) && return
     for service_model in values(services_template)
         @debug "Building $(service_model.service_type) with $(service_model.formulation) formulation"
-        services = PSY.get_components(service_model.service_type, sys)
-        if validate_available_services(service_model.service_type, services)
+        services = service_model.service_type[]
+        if validate_available_services(service_model.service_type, services, sys)
```
daniel-thom, line 11: Consider renaming this to `validate_available_services!` so that it's more obvious that you're mutating `services`.

PR #496, opened 2020-08-01
address inconsistencies in services

jd-lara:
addresses the comments in #494

### DIFF
--- a/src/devices_models/device_constructors/common/constructor_validations.jl
+++ b/src/devices_models/device_constructors/common/constructor_validations.jl
@@ -6,7 +6,7 @@ function validate_available_devices(device::Type{D}, devices) where {D <: PSY.De
     return true
 end
 
-function validate_available_services(
+function validate_services!(
     service::Type{S},
     services::Vector{S},
     sys::PSY.System,
@@ -24,7 +24,9 @@ function validate_available_services(
             name = PSY.get_name(s),
         )].contributing_devices)
             push!(services, s)
+            continue
         end
+        @warn("The contributing devices for service $(PSY.get_name(service)) is empty, consider removing the service from the system")
     end
     return true
 end
--- a/src/devices_models/device_constructors/common/constructor_validations.jl
+++ b/src/devices_models/device_constructors/common/constructor_validations.jl
@@ -24,11 +24,11 @@ function validate_services!(
             services_mapping[(type = S, name = PSY.get_name(s))].contributing_devices
         contributing_devices =
             [d for d in contributing_devices_ if typeof(d) ∉ incompatible_device_types]
-        if !isempty(contributing_devices)
+        if isempty(contributing_devices)
+             @warn("The contributing devices for service $(PSY.get_name(service)) is empty, consider removing the service from the system")
+        else
             push!(services, s)
         end
-        continue
-        @warn("The contributing devices for service $(PSY.get_name(service)) is empty, consider removing the service from the system")
     end
     return true
 end
--- a/src/devices_models/device_constructors/common/constructor_validations.jl
+++ b/src/devices_models/device_constructors/common/constructor_validations.jl
@@ -9,6 +9,7 @@ end
 function validate_services!(
     service::Type{S},
     services::Vector{S},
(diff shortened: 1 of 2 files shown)


src/services_models/services_constructor.jl
```diff
@@ -66,10 +66,8 @@ function construct_service!(
             contributing_devices =
                 [d for d in contributing_devices if typeof(d) ∉ incompatible_device_types]
         end
-        if isempty(contributing_devices)
-            @warn("The contributing devices for service $(PSY.get_name(service)) is empty, cinsider removing the service from the system")
-            continue
-        end
+        #Services without contributing devices should have been filtered out in the validation
```
daniel-thom, line 69: You might be missing my point.  The statement just above this one performs another filter on contributing_devices.  The code is now correct, but your comment is inconsistent with that.

jd-lara: I don't follow, if contributing devices for the service is empty the filtering didn't work. The above code in lines 63 - 66 all it does is find the contributing devices for the particular service.

daniel-thom: It seems possible that `[d for d in contributing_devices if typeof(d) ∉ incompatible_device_types]` could return an empty array and then cause your assert to trip.  Maybe that it not possible based on conditions unbeknownst to me; that's just how the code reads.

src/devices_models/device_constructors/common/constructor_validations.jl
```diff
@@ -6,9 +6,10 @@ function validate_available_devices(device::Type{D}, devices) where {D <: PSY.De
     return true
 end
 
-function validate_available_services(
+function validate_services!(
     service::Type{S},
     services::Vector{S},
+    incompatible_device_types::Vector{<:DataType},
```
daniel-thom, line 12: One of these would be more idiomatic.
```
    incompatible_device_types::Vector,
```
```
 incompatible_device_types::Vector{T} where T
```

src/devices_models/device_constructors/common/constructor_validations.jl
```diff
@@ -19,12 +20,15 @@ function validate_available_services(
...
+        contributing_devices_ =
+            services_mapping[(type = S, name = PSY.get_name(s))].contributing_devices
+        contributing_devices =
+            [d for d in contributing_devices_ if typeof(d) ∉ incompatible_device_types]
+        if !isempty(contributing_devices)
             push!(services, s)
         end
+        continue
+        @warn("The contributing devices for service $(PSY.get_name(service)) is empty, consider removing the service from the system")
     end
```
daniel-thom, line 32: I think you want something like this:

```suggestion
        if isempty(contributing_devices)
             @warn("The contributing devices for service $(PSY.get_name(service)) is empty, consider removing the service from the system")
        else
            push!(services, s)
        end
    end
```

src/services_models/services_constructor.jl
```diff
@@ -106,6 +117,10 @@ function construct_service!(
                 type = typeof(service),
                 name = PSY.get_name(service),
             )].contributing_devices
+        if !isempty(incompatible_device_types)
+            contributing_devices =
+                [d for d in contributing_devices if typeof(d) ∉ incompatible_device_types]
+        end
```
daniel-thom, line 123: Can this code block be removed?  You shouldn't need to pass incompatible_device_types to this function since you already handled it.

jd-lara: No, because there might be more devices than the incompatible ones in the contributing devices.

src/devices_models/device_constructors/common/constructor_validations.jl
jd-lara: But we don't use the T variable.

src/services_models/services_constructor.jl
daniel-thom: Seems like we're missing each other. In the `construct_services` function you already filtered out the incompatible devices.  This check looks redundant.  Same for the other `construct_service` function.

You should probably assert that contributing_devices is not empty, just like you did in the other one.

PR #498, opened 2020-08-05
Update Hydro Models

rodrigomha:
Implement and expand the three models available for Hydro to include Commitment. The models are:
- [x] `RunOfRiver`. Already implemented for both Dispatch and Commit.
- [x] `ReservoirFlow` will be changed to ReservoirBudget. This model has a budget of energy that must be considered by the end of the timeseries. Inflow is not used and must be considered by the user in the data of the budget. 
- [x] `ReservoirStorage` is the usual model with SoC type constraints. In addition a limitation of storage is considered for each time step. 

Label for forecast are important. `"get_inflow"` will be maintained for the inflow. I'm thinking on `"get_hydro_budget"` for the forecast of energy budget, and `"get_hydro_storage"` maybe for the limitations on storage SoC?


src/devices_models/devices/hydro_generation.jl
```diff
@@ -526,24 +576,26 @@ function device_energy_limit_param_ub(
...
         name = get_name(constraint_info)
+        resolution = model_resolution(psi_container)
+        inv_dt = 1.0 / (Dates.value(Dates.Second(resolution)) / SECONDS_IN_HOUR)
         forecast = constraint_info.timeseries
-        multiplier = constraint_info.multiplier
+        multiplier = constraint_info.multiplier * inv_dt
         constraint[name] = JuMP.@constraint(
             psi_container.JuMPmodel,
-            sum([variable[name, t] for t in time_steps]) <= multiplier * sum(forecast)
+            sum([variable[name, t] for t in time_steps]) <= multiplier * forecast[end]
```
claytonpbarrows, line 598: I don't understand this change. What is significance of the end of the energy_budget forecast? Since we can't have mixed resolution forecasts, we had previously implemented it with the assumption that the energy budget was distributed evenly across the resolution of the forecast. e.g. if you have daily energy budget data, and an hourly `System` you would make an energy_budget forecast that had (daily energy budget)/24 in each hour. In this case, you would have to sum the RHS of the constraint. I think your case would require repeating the daily energy budget 24x since you are just using one value.

rodrigomha: It is up to us here. After discussing with @jd-lara, the idea was to modify the time series to be the same value the entire time series (i.e. each entry will be the budget), and the value that matters is the last one.

However, this idea of having the budget divided by 24 is also an option. It is up to us to decide how we want the data to be provided.
This also will appear as a similar constraint for the energy capacity in the ReservoirStorage, on which we want that the energy at the end of the time series to be greater than the proportion of the storage capacity. Again we may want to decide how we are going to provide the time series. As long as we stick with one we can decide use the similar approach for both. I was also thinking on generalizing the labels here, and instead of `hydro_budget` and `energy_bound, we can go with `storage_energy_limit` or something similar that has different interpretation depending on the model.

jd-lara: I am not sure if adding many paramenters and their multipliers is the way to go with the budget. That formulation is pretty messy and it might even need time resolution conversions. We can discuss more about the implications of each.

rodrigomha:
@jd-lara, @claytonpbarrows 

There is an issue with the initial condition of the storage. The way is currently set-up uses a parameter in the struct `HydroEnergyReservoir`. Not sure if this is a good idea. In a simulation this will require to update the parameter itself in the struct. Is this a typical idea when setting simulations and feedforward updates between operation problems?

Basically, the main issue with Hydro models is that the current implementation of only allowing forecasts as input data is a bottleneck for the modeling itself. In practice, we need one parameter for the energy budget, one parameter for the initial condition and one parameter for required storage value at the end of the operational problem. Using a forecast for something that is basically one parameter is an issue.

src/devices_models/devices/hydro_generation.jl
claytonpbarrows: Using the last of repeated values means that whatever forecast is given for the budget has to be carefully calculated for the specific problem horizon. By dividing a budget into average values per time period, the op_problem can sum over whatever horizon it represents to get a valid energy budget.

claytonpbarrows:
> @jd-lara, @claytonpbarrows
> 
> There is an issue with the initial condition of the storage. The way is currently set-up uses a parameter in the struct `HydroEnergyReservoir`. Not sure if this is a good idea. In a simulation this will require to update the parameter itself in the struct. Is this a typical idea when setting simulations and feedforward updates between operation problems?
> 
> Basically, the main issue with Hydro models is that the current implementation of only allowing forecasts as input data is a bottleneck for the modeling itself. In practice, we need one parameter for the energy budget, one parameter for the initial condition and one parameter for required storage value at the end of the operational problem. Using a forecast for something that is basically one parameter is an issue.

We should use the stored energy field value for the initial condition of the first problem in a simulation. After that we should use feedforward to set the stored energy without having to manipulate the struct value.

@sourabhdalvi might have ideas on how to proceed with the required end stored energy

rodrigomha:
> > @jd-lara, @claytonpbarrows
> > There is an issue with the initial condition of the storage. The way is currently set-up uses a parameter in the struct `HydroEnergyReservoir`. Not sure if this is a good idea. In a simulation this will require to update the parameter itself in the struct. Is this a typical idea when setting simulations and feedforward updates between operation problems?
> > Basically, the main issue with Hydro models is that the current implementation of only allowing forecasts as input data is a bottleneck for the modeling itself. In practice, we need one parameter for the energy budget, one parameter for the initial condition and one parameter for required storage value at the end of the operational problem. Using a forecast for something that is basically one parameter is an issue.
> 
> We should use the stored energy field value for the initial condition of the first problem in a simulation. After that we should use feedforward to set the stored energy without having to manipulate the struct value.
> 
> @sourabhdalvi might have ideas on how to proceed with the required end stored energy

Is in this case then better to add an additional field in the HydroReservoir struct for the budget? Could be again as a ratio of the total energy capacity or as energy in `per-unit * hour`. This idea makes more sense than adding a forecast that is using the sum of all values or the last one. We can use that value for both the energy budget formulation (`DeltaT * (sum of all P) <= budget)` and for the energy storage (SoC) at the end of the period (`SoC[end] >= budget`).

In such case, the update of the initial storage capacity and the budget (or final storage capacity) can be treated in the same way.

rodrigomha:
I revert back to the sum of budget for now for the HydroBudget model. Commitment models for RunOfRiver and ReservoirBudget are working fine by themselves in an OperationsProblem. However, the feedforward to pass both binary variables and an IntegralLimitFF has yet to be done. 

To do: CommitmentReservoirStorage. FF for other models

rodrigomha:
> This looks really nice. Thanks!
> My one comment (and this really belongs in PSY) is that the `get_hydro_budget` forecast label is inconsistent with all other forecast labels in that `hydro_budget` isn't a field in `HydroEnergyReservoir`. There is nothing from presenting this from working, it's just a bit different than our usual approach. I suppose that's why @sourabhdalvi was using the `get_storage_capacity` label before.

I see. Yes, this is kind an issue here. In addition, some discussion was made that in the future we are interested in providing a terminal state for the storage. Maybe we can change our budget to be initial state minus terminal state? Assuming that terminal state has to be smaller than the initial one. That is (sort of) more consistent instead of using the storage capacity.

PR #501, opened 2020-08-07
fix incorrect recorder assignment

jd-lara:


odow:
Hahahahaha oh man. What a bug. No wonder you were bashing your head against this.

PR #503, opened 2020-08-18
Constraint is wrong if efficiency is not 1

tamas-borbath:
Hi,

There seems to be a small bug in this formulation. Eff_in and Eff_out should probably behave the same way. 
Great project, it makes my life a lot easier these days!

Thanks!

### DIFF
--- a/src/devices_models/devices/common/energy_balance_constraint.jl
+++ b/src/devices_models/devices/common/energy_balance_constraint.jl
@@ -53,7 +53,7 @@ function energy_balance(
         # Create the PGAE outside of the constraint definition
         balance =
             initial_conditions[ix].value + varin[name, 1] * eff_in * fraction_of_hour -
-            (varout[name, 1]) * fraction_of_hour / eff_out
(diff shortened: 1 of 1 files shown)


src/devices_models/devices/common/energy_balance_constraint.jl
```diff
@@ -67,7 +67,7 @@ function energy_balance(
             psi_container.JuMPmodel,
             varenergy[name, t] ==
             varenergy[name, t - 1] + varin[name, t] * eff_in * fraction_of_hour -
-            (varout[name, t]) * fraction_of_hour / eff_out
+            (varout[name, t]) * fraction_of_hour * eff_out
```
jd-lara, line 70: if we use this model, the `eff_out` values need to be larger than 1.0. If this is the case. It also needs to address the calculation of the these factors elsewhere in the code.

tamas-borbath: You are correct. This should stay the way it was. 
In fact, the bug I tried to solve comes from the function call in storage.jl.

PR #507, opened 2020-08-19
Cb/nomin cost fix

claytonpbarrows:


jd-lara:
The PR looks fine but the implementation of the `_pwl_check` is incorrect and the tests are failing.

claytonpbarrows:
@jd-lara This is done

PR #508, opened 2020-08-20
patch get name

### LINKED ISSUE #506: Fix get_name function
`PSI.get_name` and `PSY.get_name`  conflict when imported because `PSI` creates its own function instead of extending `IS.get_name`. Was this intentional? Code implementations of mentioned function:
https://github.com/NREL-SIIP/PowerSimulations.jl/blob/e620fa1829e23f7f8b2ea993bc10705eda89dc2a/src/core/simulation.jl#L290
https://github.com/NREL-SIIP/PowerSimulations.jl/blob/e620fa1829e23f7f8b2ea993bc10705eda89dc2a/src/devices_models/devices/common/constraints_structs.jl#L27
https://github.com/NREL-SIIP/PowerSimulations.jl/blob/e620fa1829e23f7f8b2ea993bc10705eda89dc2a/src/devices_models/devices/common/constraints_structs.jl#L44
https://github.com/NREL-SIIP/PowerSimulations.jl/blob/e620fa1829e23f7f8b2ea993bc10705eda89dc2a/src/devices_models/devices/common/constraints_structs.jl#L137

jd-lara:
closes #506

### DIFF
--- a/src/core/simulation.jl
+++ b/src/core/simulation.jl
@@ -213,7 +213,7 @@ end
                 )
 
 """
-mutable struct Simulation <: IS.InfrastructureSystemsType
+mutable struct Simulation
     steps::Int
     stages::Dict{String, Stage{<:AbstractOperationsProblem}}
     initial_time::Union{Nothing, Dates.DateTime}
--- a/src/core/simulation.jl
+++ b/src/core/simulation.jl
@@ -213,7 +213,7 @@ end
                 )
 
 """
-mutable struct Simulation
+mutable struct Simulation <: IS.InfrastructureSystemsType
     steps::Int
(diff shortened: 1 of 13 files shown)


src/core/simulation.jl
```diff
@@ -213,7 +213,7 @@ end
                 )
 
 """
-mutable struct Simulation
+mutable struct Simulation <: IS.InfrastructureSystemsType
```
daniel-thom, line 216: I'm not sure this change is appropriate.  InfrastructureSystemsType is typically a thing you store in a system.  An undesirable side effect is that called `methodswith(Simulation, supertypes=true)` is going to report methods that do not exist for this type.

jd-lara: Ok, I will remove it.

PR #509, opened 2020-08-20
Updates for FixedOutput model

rodrigomha:
The following change follows the same logic of FixedOutput for Renewables.

However, this idea is not correct. Basically, we are using the same forecast for active power for reactive power.

The solution here is essentially make Q = 0 all times, or use a constant power factor, that depending on P, will update Q. Any ideas here on how to move forward?

### DIFF
--- a/src/devices_models/devices/hydro_generation.jl
+++ b/src/devices_models/devices/hydro_generation.jl
@@ -368,6 +368,20 @@ end
 
 function NodalExpressionSpec(
     ::Type{T},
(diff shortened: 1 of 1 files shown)


jd-lara:
@rodrigomha and @claytonpbarrows let's merge this PR to address the shortcoming of the current approach in the hydro works

jd-lara (approved):
@claytonpbarrows we should merge this into psy 1.0 branch.

PR #511, opened 2020-08-20
[WIP] Make new data store

jd-lara:


jd-lara:
@daniel-thom should we close this?

daniel-thom:
Yes, this is obsolete and we can delete it. It was just a POC.

PR #513, opened 2020-08-20
Update System constructor calls, apply coding standards

daniel-thom:
1. Updates PSY.System constructor calls per https://github.com/NREL-SIIP/PowerSystems.jl/pull/593.
2. Applies Julia coding standards to cases where function argument order was incorrect or where a mutating function did not end with `!`.
3. Suppresses an expected error log message in a test.

The diff is large, but there is no real logic change.

### DIFF
--- a/src/devices_models/device_constructors/hydrogeneration_constructor.jl
+++ b/src/devices_models/device_constructors/hydrogeneration_constructor.jl
@@ -38,23 +38,23 @@ function construct_device!(
     end
 
     #Variables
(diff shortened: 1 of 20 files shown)


jd-lara:
The fail in documentation is upstream from PowerGraphics

jd-lara:
@daniel-thom the docs are failing because the code in https://github.com/NREL-SIIP/PowerSimulations.jl/blob/master/docs/src/get_test_data.jl needs updating too for PSY 0.28

Issue #514, opened 2020-08-26
Improved Support for some Day-Ahead energy - services co-optimization 


jd-lara:
The current design of devices and services doesn't support a straight forward way to make energy - services co-optimization problem. Several steps are required to support this in a straight forward way. I am opening this issue to keep track of the progress to get this going. 

- [x] New Cost Function structure  x-ref: https://github.com/NREL-SIIP/PowerSystems.jl/pull/596 
- [x] Refactor cost functions implementation x-ref: https://github.com/NREL-SIIP/PowerSimulations.jl/pull/516
- [x] Implement a system constructor finalizer call. 

cc. @andrewrosemberg @raphaelsaavedra

andrewrosemberg:
should we remove the old wrong solution from PSY? https://github.com/NREL-SIIP/PowerSystems.jl/blob/e6bcc29f88d25ad5d7d61d4b893525651c6ee88e/src/models/generated/StaticReserve.jl#L35
I can make a PR for that.

jd-lara:
@andrewrosemberg and @raphaelsaavedra, we have now support for market bids as requested originally. @sourabhdalvi has implemented examples of this. The last bullet point about the construction process will take a bit more to get done since we need to make changes to the problem build process.

andrewrosemberg:
@jd-lara and @sourabhdalvi thanks for the effort and work done! This will help us a lot. Could you like the examples here?

PR #516, opened 2020-08-27
refactor cost functions

jd-lara:
This addresses long-standing issues on the cost functions. It also implements the new cost function for the joint energy - services market bid. 

It also removes some hacks to handle PWL functions cases. 

x-ref #514 

cc. @andrewrosemberg @raphaelsaavedra this is WIP for feedback. At this stage the tests don't pass but you can see the path I am taking. 

@claytonpbarrows please take a look.

### DIFF
--- a/src/core/definitions.jl
+++ b/src/core/definitions.jl
@@ -61,7 +61,6 @@ const MAX_START_STAGES = 3
 const OBJECTIVE_FUNCTION_POSITIVE = 1.0
 const OBJECTIVE_FUNCTION_NEGATIVE = -1.0
 
-
 # Interface limitations
 const OPERATIONS_ACCEPTED_KWARGS = [
     :horizon,
--- a/src/core/definitions.jl
+++ b/src/core/definitions.jl
@@ -37,6 +37,12 @@ const DenseAxisArrayContainer = Dict{Symbol, JuMP.Containers.DenseAxisArray}
     EMPTY = 0
 end
 
+@enum SOS_STATUS_VARIABLE begin
+    NO_VARIABLE = 1
+    PARAMETER = 2
+    VARIABLE = 3
+end
+
 # Settings constants
 const UNSET_HORIZON = 0
 const UNSET_INI_TIME = Dates.DateTime(0)
@@ -52,6 +58,9 @@ const SECONDS_IN_MINUTE = 60.0
 const MINUTES_IN_HOUR = 60.0
 const SECONDS_IN_HOUR = 3600.0
 const MAX_START_STAGES = 3
+const OBJECTIVE_FUNCTION_POSITIVE = 1.0
+const OBJECTIVE_FUNCTION_NEGATIVE = -1.0
+
 
 # Interface limitations
 const OPERATIONS_ACCEPTED_KWARGS = [
--- a/src/devices_models/device_constructors/hydrogeneration_constructor.jl
+++ b/src/devices_models/device_constructors/hydrogeneration_constructor.jl
@@ -63,7 +63,7 @@ function construct_device!(
     feedforward!(psi_container, devices, model, get_feedforward(model))
 
     #Cost Function
-    cost_function(psi_container, devices, D, S)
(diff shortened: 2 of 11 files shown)


jd-lara:
The device's cost functions are ready. The PWL cost function for the ORDC reserve still needs fixing.

andrewrosemberg:
great work José!
I'm looking into your changes and I have a couple of questions:

1. Couldn't find if there is a restriction on negative prices. Is this the case? because renewables and some other subsidized generators (or ones that just really want to be dispatched) sometimes submit **negative prices** and we want to make sure we can cover those.

2. A. Does this cover also **negative MW** ? Some generators might be paid to consume energy so their cost curves might have negative parts (see the weird hydro [here](https://nrel-siip.slack.com/files/U018T00UX45/F01A913JWDQ/special_generators.pdf))  
2. B. On that matter, do you see any problems in allowing generators to have negative `Pmin`? or should we create a new type?

raphaelsaavedra:
>Couldn't find if there is a restriction on negative prices. Is this the case? because renewables and some other subsidized generators (or ones that just really want to be dispatched) sometimes submit negative prices and we want to make sure we can cover those.

Note that PSY and PSI model the cost curve, not the offer curve. While these generators do submit offer curves, the resulting cost curve will always be non-negative.

andrewrosemberg:
> > Couldn't find if there is a restriction on negative prices. Is this the case? because renewables and some other subsidized generators (or ones that just really want to be dispatched) sometimes submit negative prices and we want to make sure we can cover those.
> 
> Note that PSI models the cost curve, not the offer curve. While these generators do submit offer curves, the resulting cost curve will always be non-negative.

Could you give us an example?

raphaelsaavedra:
>Note that PSI models the cost curve, not the offer curve. While these generators do submit offer curves, the resulting cost curve will always be non-negative.

Sorry, I misspoke here. The cost curve can be negative in some points.

>Could you give us an example?

Just consider any offer curve that is a stepwise function starting from negative values. When you integrate that offer curve to get to the cost curve, the result will be a piecewise linear function where the first segment has negative slope (and eventually can have positive slope).

andrewrosemberg:
understood, so in fact, it can be negative. Either @raphaelsaavedra it would be better for you to describe here what are you trying to implement instead of internally.

raphaelsaavedra:
Hi @jd-lara, here's a brief summary of some "edge cases" for generators that we'd like to make sure that PowerSimulations is able to handle.

* Units that submit negative offer prices: from what we've seen, these are usually renewables, but there are some other cases too. Basically units that don't mind being price takers. As mentioned above, this gets translated to a cost curve that can have negative values (but is still convex).
* Units that inform a negative Pmin: these are the so-called DRR type II units, that submit offers together with generators but are basically loads that can reduce their load for a price. This gets translated to an offer curve that is entirely for negative MW values (i.e., Pmin < Pmax < 0). Additionally, there are some rare cases where Pmin < 0 < Pmax (we're not exactly sure what these are, we suspect asynchronous resources).

andrewrosemberg:
Hey, there is another cost to address that might be very easy given the new structure. We need a service-specific pice-wise linear cost. 
Currently, service slacks are unique. Could we perhaps pass them optionally in the problem definition? Remembering that we need piecewise instead of a single slope.

jd-lara:
> Hey, there is another "cost" (actually penalty) that might be very easy to implement given the new structure. We need a service-specific pice-wise-linear slack penalty. 
> Currently, service slacks penalties are a single unique linear value for all services ([ref](https://github.com/NREL-SIIP/PowerSimulations.jl/blob/16e1b453ed21e714db58065e03601af4405b2e54/src/core/definitions.jl#L48)). Could we perhaps pass slack pwl penalties optionally in the problem definition? Remembering that we need piecewise instead of a single slope.

This is already implemented for a slightly different reserve that has a PWL penalty for shortage. Currently that's the one with some issues due to time Variant costs.

PR #517, opened 2020-08-27
Change reserves formulations to abstract

andrewrosemberg:
Changes reserves formulations (RangeReserve and StepwiseCostReserve) to abstract.
This pattern will help type 2 users reuser most of the structure existing and won't imply any breaking changes or big refactoring.


andrewrosemberg:
dspatch_abstract

jd-lara:
we have clarified how to achieve the same behavior without the abstraction.

PR #519, opened 2020-08-28
Change instances of Int64 to Int

raphaelsaavedra:
Is there any reason for the occurrences of `Int64` in the code? Having them as `Int` would be good. We have a small test running which is failing for 32-bit runners and this might be the issue.


jd-lara:
Thanks for the fix, this is some legacy code.

PR #528, opened 2020-10-14
ph merge

jd-lara:
@claytonpbarrows this merges your changes for Pump Hydro into the update psy branch. I fixed the conflicts but we still need to fix some tests for the hydro.


src/core/simulation.jl
```diff
@@ -452,8 +449,10 @@ function _check_steps(
     for (stage_number, stage_name) in sim.sequence.order
         stage = sim.stages[stage_name]
         execution_counts = get_executions(stage)
-        @assert length(findall(x -> x == stage_number, sim.sequence.execution_order)) ==
-                execution_counts
+        transitions =
+            sim.sequence.execution_order[vcat(1, diff(sim.sequence.execution_order)) .== 1]
```
jd-lara, line 453: @claytonpbarrows what's the purpose of this logic?

src/core/simulation_results.jl
```diff
@@ -67,8 +67,9 @@ references = make_references(sim, "2019-10-03T09-18-00-test")
 ```
 """
 function make_references(sim::Simulation, date_run::String)
-    sim.internal.date_ref[1] = sim.initial_time
-    sim.internal.date_ref[2] = sim.initial_time
+    for stage_number in keys(sim.internal.date_ref)
```
jd-lara, line 70: nice catch

src/core/simulation.jl
claytonpbarrows: 1. find all the times that you execute a subsequent stage ( find 1->2 transitions, not 2->1)
 2. the total times that a stage is executed in the entire simulation / the number of simulation steps should be equal to the expected number of stage executions per step

The previous logic was failing when we ran a simulation that ran multiple executions of some stages.

src/core/simulation.jl
```diff
@@ -581,7 +580,10 @@ function _build!(sim::Simulation)
             throw(IS.ConflictingInputsError("Stage $(stage_name) not found in the stages definitions"))
         end
         stage_interval = get_stage_interval(sim, stage_name)
-        stage.internal.executions = Int(get_step_resolution(sim.sequence) / stage_interval)
+        step_resolution =
```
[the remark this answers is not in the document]
claytonpbarrows: this was only working if you had a simulation that was perfectly sequential in it's execution order. For example, if `sim.sequence.execution_order` was `[1,2,2,3,3,3,3,1,2,2,3,3,3,3]` it would work, but if it was `[1,2,3,3,2,3,3,1,2,3,3,2,3,3]` it was failing.

src/core/simulation.jl
jd-lara: We need to add more tests for all potential sequences to make sure that everything works as expected.

src/core/simulation.jl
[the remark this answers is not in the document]
jd-lara: We could probably make this test into a more streamlined format.

PR #529, opened 2020-10-22
Update psy1 0

claytonpbarrows:


src/services_models/GroupReserve.jl
```diff
@@ -36,7 +36,8 @@ function service_requirement_constraint!(
         for r in contributing_services
     ]
 
-    ts_vector = get_time_series(psi_container, service, "requirement")
+    ts_vectors = [get_time_series(psi_container, s, "requirement") for s in contributing_services]
+    ts_vector = sum(hcat(ts_vectors...), dims = 2)
```
claytonpbarrows, line 40: @jd-lara I'm not sure this is how we want to do this, but it works.

jd-lara: we should add a note to refactor this later. The priority is to restore capabilities using PSY-1.0

jd-lara (approved):
I will merge this PR to start the transitions. As discussed to enable the cross package testing we need to maintain master as the dev branch.

PR #532, opened 2020-10-23
Fix TimeSeries for GroupReserve

andrewrosemberg:
Fixes TimeSeries handling for `GroupReserve`. The requirement in a `GroupReserve` is supposed to be an extra constraint and shouldn't be affected by the requirements of the contributing services.

EX: `GroupReserve` is composed of Regulation and Spinning.

The market operator sets the requirement for Regulation to be 100 MWh (contributing generators should reserve together this amount, let's call their summed contribution R_reg) and Spinning to be 50 MWh (let's call their summed contribution R_spin). For security, the operator creates a new requirement of 200 MW (a `GroupReserve`) for the sum of Regulation and Spinning reserves. The constraints should be:
 1) R_reg >= 100
 2) R_spin >= 50
 3) R_reg + R_spin >= 200

### DIFF
--- a/src/services_models/GroupReserve.jl
+++ b/src/services_models/GroupReserve.jl
@@ -24,7 +24,6 @@ function service_requirement_constraint!(
     contributing_services::Vector{<:PSY.Service},
 ) where {SR <: PSY.StaticReserveGroup}
     parameters = model_has_parameters(psi_container)
-    use_forecast_data = model_uses_forecasts(psi_container)
     initial_time = model_initial_time(psi_container)
     @debug initial_time
     time_steps = model_time_steps(psi_container)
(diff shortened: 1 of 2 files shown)


src/services_models/GroupReserve.jl
```diff
@@ -36,10 +36,7 @@ function service_requirement_constraint!(
         for r in contributing_services
     ]
 
-    # TODO: should we make a get_time_series method for the StaticReserveGroup that handles the following two lines?
-    ts_vectors =
-        [get_time_series(psi_container, s, "requirement") for s in contributing_services]
-    ts_vector = sum(hcat(ts_vectors...), dims = 2)
+    ts_vector = get_time_series(psi_container, service, "requirement")
```
claytonpbarrows, line 39: OK, this makes sense. I think you just need to add a "requirement" time series to the "init" group reserve [here](https://github.com/NREL-SIIP/PowerSimulations.jl/blob/master/test/test_services_constructor.jl#L207) to get the test to pass.

andrewrosemberg: So I wanted to make this generic for when we have a `VariableReserveGroup` just like the code for normal reserves.
However, currently, we only have `StaticReserveGroup` so there is no way to add a time-series to the PSY service.

@claytonpbarrows do you prefer I remove the time-series handling for now? I haven't figured it out how to make it so it defaults back to `ts_vector = ones(time_steps[end])` when there are no time-series added.

claytonpbarrows: Yeah, I think we should remove the time series handling since this method is specific to `StaticReserveGroup`. Then we can create another method to handle the time series for the `VariableReserveGroup`.

>I haven't figured it out how to make it so it defaults back to ts_vector = ones(time_steps[end]) when there are no time-series added.

The existing examples that default to the vector of ones are for when the entire OperationsProblem isn't using time series data. So, for now we don't support defaulting to a vector of ones when the requested time series data isn't available.

andrewrosemberg: time-series handling removed!

andrewrosemberg:
tests passing 👍

claytonpbarrows (approved):
LGTM. Thanks for catching this!

PR #533, opened 2020-10-23
Updating cost functions to allow time series cost data

sourabhdalvi:


src/core/definitions.jl
```diff
@@ -65,6 +65,7 @@ const SECONDS_IN_HOUR = 3600.0
 const MAX_START_STAGES = 3
 const OBJECTIVE_FUNCTION_POSITIVE = 1.0
 const OBJECTIVE_FUNCTION_NEGATIVE = -1.0
+const DEFAULT_RESERVE_COST = 1.0e-4
```
jd-lara, line 68: Add a comment about the use of this constant. e.g. This value is used to avid degeneracy of the solutions.

src/devices_models/devices/common/cost_functions.jl
```diff
@@ -49,6 +49,17 @@ function AddCostSpec(
     error("AddCostSpec is not implemented for $T / $U")
 end
 
+function add_service_variables!(
+    spec::AddCostSpec,
+    services::Vector{T},
+) where {T <: PSY.Service}
+    for service in services
+        name = PSY.get_name(service)
+        spec.addtional_linear_terms[name] = make_variable_name(name, typeof(service))
```
jd-lara, line 58: We should do this with a setter function just in case the implementation changes later.

src/devices_models/devices/common/cost_functions.jl
```diff
@@ -149,15 +162,29 @@ function linear_gen_cost!(
...
     add_to_cost_expression!(psi_container, gen_cost * dt)
     return
 end
 
+function linear_gen_cost!(
+    psi_container::PSIContainer,
+    var_name::Symbol,
+    component_name::String,
+    linear_term::Float64,
+    time_period::UnitRange{Int},
```
jd-lara, line 180: time_period as a range?

src/devices_models/devices/common/cost_functions.jl
```diff
@@ -186,32 +213,32 @@ function pwl_gencost_sos!(
...
         var_key = make_variable_name(OnVariable, spec.component_type)
-        bin = get_variable(psi_container, var_key)[component_name, :]
+        bin = get_variable(psi_container, var_key)[component_name, time_period]
         @debug("Using Piecewise Linear cost function with variable $(var_key)")
     else
         @assert false
     end
 
-    for (t, var) in enumerate(variable)
+    for var in variable
```
jd-lara, line 241: You don't need to use a for loop here since the variable will be a single value for the component and time_period

src/devices_models/devices/common/cost_functions.jl
```diff
@@ -386,10 +425,12 @@ function add_to_cost!(
...
     cost_data::PSY.MultiStartCost,
-    component_name::String,
+    component::PSY.Component,
 )
+    component_name = PSY.get_name(component)
     resolution = model_resolution(psi_container)
     dt = Dates.value(Dates.Second(resolution)) / SECONDS_IN_HOUR
+    time_steps = model_time_steps(psi_container)
 
     if !isnothing(spec.fixed_cost) && spec.has_status_variable
```
jd-lara, line 435: I think is better to loop over the time axis here and add each linear cost one at a time instead of passing time to the function.

src/devices_models/devices/common/cost_functions.jl
```diff
@@ -443,51 +490,96 @@ function add_to_cost!(
...
+    component_name = PSY.get_name(component)
     @debug "Market Bid" component_name
     resolution = model_resolution(psi_container)
     dt = Dates.value(Dates.Second(resolution)) / SECONDS_IN_HOUR
-    variable_cost = PSY.get_variable(cost_data)
-    variable_cost!(psi_container, spec, component_name, variable_cost)
+    time_steps = model_time_steps(psi_container)
+    initial_time = model_initial_time(psi_container)
+    variable_cost_forecast = get_time_series(psi_container, component, "variable_cost")
+    variable_cost_forecast = map(PSY.VariableCost, variable_cost_forecast)
```
jd-lara, line 502: Add a todo there to say we need to eliminate this step.

src/devices_models/devices/common/cost_functions.jl
```diff
@@ -443,51 +490,96 @@ function add_to_cost!(
...
+                psi_container,
+                make_variable_name(StopVariable, spec.component_type),
+                component_name,
+                spec.shut_down_cost(cost_data) * spec.multiplier,
+                t,
+            )
+        end
     end
 
     # Here goes the services bids
```
jd-lara, line 551: remove comment

src/devices_models/devices/common/cost_functions.jl
```diff
@@ -443,51 +490,96 @@ function add_to_cost!(
...
+                linear_gen_cost!(
+                    psi_container,
+                    spec.addtional_linear_terms[PSY.get_name(service)],
+                    component_name,
+                    forecast_data,
+                    t,
+                )
+            end
+        else
+            error("Current version only supports linear cost bid for services, please change the forecast data for $(PSY.get_name(service))")
```
jd-lara, line 567: I think that for these cases we should create a `ModelingError` type in IS. @daniel-thom what do you think?

jd-lara (changes requested):
Overall looks good but we need to move the time loop outside instead of passing it around. We might need to move some of the get_variable code outside too.

src/devices_models/devices/common/cost_functions.jl
```diff
@@ -10,7 +10,7 @@ struct AddCostSpec
     shut_down_cost::Union{Nothing, Function}
     fixed_cost::Union{Nothing, Function}
     # Not Currently in use. For future extensions
```
daniel-thom, line 12: Seems like this comment can be removed.

src/devices_models/devices/common/cost_functions.jl
daniel-thom, line 58: It looks like this uses the fallback `make_variable_name` method where no VariableType is specified.  Should there be a new method, perhaps using ActiveServiceVariable?

src/devices_models/devices/common/cost_functions.jl
```diff
@@ -49,6 +49,17 @@ function AddCostSpec(
     error("AddCostSpec is not implemented for $T / $U")
 end
 
+function add_service_variables!(
+    spec::AddCostSpec,
+    services::Vector{T},
```
daniel-thom, line 54: Given that the function is called add_service_variables you may not need T for dispatch.  Making services untyped would allow easier use of the function because users could pass any iterable.

src/devices_models/devices/common/cost_functions.jl
```diff
@@ -59,10 +70,12 @@ function cost_function!(
     ::Type{<:PM.AbstractPowerModel},
     feedforward::Union{Nothing, AbstractAffectFeedForward} = nothing,
 ) where {T <: PSY.Component, U <: AbstractDeviceFormulation}
-    spec = AddCostSpec(T, U, psi_container)
-    @debug T, spec
     for d in devices
-        add_to_cost!(psi_container, spec, PSY.get_operation_cost(d), PSY.get_name(d))
+        spec = AddCostSpec(T, U, psi_container)
```
daniel-thom, line 74: Does `spec` need to be created in every loop iteration?  Or can one instance be shared?

src/devices_models/devices/common/cost_functions.jl
```diff
@@ -443,51 +490,96 @@ function add_to_cost!(
...
+                t,
+            )
+        end
     end
 
     # Here goes the services bids
+    ancillary_services = PSY.get_ancillary_services(cost_data)
+    for service in ancillary_services
+        forecast_data = get_time_series(psi_container, component, PSY.get_name(service))
+        forecast_data = map(PSY.VariableCost, forecast_data)
```
daniel-thom, line 555: Doesn't your new PSY code obviate the need to do this?

src/devices_models/devices/common/cost_functions.jl
daniel-thom: `error` is called many times in PSI.  This fits with the current usage.  I think that it does result in a clean, direct message to the user telling them how to correct the problem.

test/test_services_constructor.jl
```diff
@@ -195,7 +195,9 @@ end
     services = get_components(Service, c_sys5_uc)
     contributing_services = Vector{Service}()
     for service in services
-        push!(contributing_services, service)
+        if !(typeof(service) <: PSY.ReserveDemandCurve)
```
daniel-thom, line 198: Is there a way to express this behavior in a function?  If this is only code that will be used in this test then it's fine.  If not, this type of code doesn't grow well when new types are added.

test/test_utils/get_test_data.jl
```diff
@@ -215,13 +215,18 @@ function build_c_sys5_re(; kwargs...)
...
         end
-        #for t in 1:2, serv in get_components(ReserveDemandCurve, c_sys5_re)
-        #    add_time_series!(
-        #        c_sys5_re,
-        #        serv,
-        #        PiecewiseFunction("variable", 5, ORDC_cost_ts[t]),
-        #    )
-        #end
+        for (ix, serv) in enumerate(get_components(ReserveDemandCurve, c_sys5_re))
+            forecast_data = SortedDict{Dates.DateTime, TimeArray}()
```
daniel-thom, line 219: Whenever possible I think that we should not use the TimeArray constructor for Deterministic.  It's less efficient and we don't want people to follow the example when the alternative is just as easy.  You could add `TimeSeries.values()` to line 222.

test/test_services_constructor.jl
sourabhdalvi: this is only for test, I don't expect it to be used anywhere else

src/devices_models/devices/common/cost_functions.jl
sourabhdalvi: Jose's said he plans on adding caching to the `get_time_series` function and then we can update the usage here, originally I had used the methods from PSY

src/devices_models/devices/common/cost_functions.jl
sourabhdalvi: One instance can't be shared as different devices can have different services, so different terms in the `addtional_linear_terms` field

src/devices_models/devices/common/cost_functions.jl
sourabhdalvi: For this variable there isn't a pre-defined name (like the const names) it used the name of the service itself to build for the variable name in reserves.jl, so the new method would need use to pass the service along  with the variable struct, not sure if that's the best way to do this ?
https://github.com/NREL-SIIP/PowerSimulations.jl/blob/37e14409d543ba57e80632b53a1e6f2b8d8397b9/src/services_models/reserves.jl#L8-L14

jd-lara (approved):

PR #536, opened 2020-10-28
Add test for StaticReserve as RangeReserve

andrewrosemberg:
Adds test for StaticReserve as RangeReserve.


andrewrosemberg:
@jd-lara added the new dispatch of `service_requirement_constraint` for `StaticReserve`.

jd-lara (approved):
Looks good to me, I will merge as soon as tests pass.

PR #541, opened 2020-11-05
don't export logging

jd-lara:


src/PowerSimulations.jl
```diff
@@ -149,7 +149,6 @@ export get_con_index
 export get_var_index
 export get_result_variable
 export get_variable_names
-export configure_logging
```
daniel-thom, line 152: What is the reason for the change? I thought this was desired by all. Were you getting conflicts with the IS version?

jd-lara:
Its conflicting with PSY exports of the same. I think we can refer to PSY for setting the logger.

daniel-thom:
> 
> 
> Its conflicting with PSY exports of the same. I think we can refer to PSY for setting the logger.

We added PSI.configure_logging because you thought that the IS/PSY version was too complicated. The PSI version accepts only `console_level`, `file_level`, and `filename`. I was OK with only using the IS/PSY version but found some value in your argument. Your opinion today implies that we should delete the PSI version altogether. Right?

jd-lara:
> > Its conflicting with PSY exports of the same. I think we can refer to PSY for setting the logger.
> 
> We added PSI.configure_logging because you thought that the IS/PSY version was too complicated. The PSI version accepts only `console_level`, `file_level`, and `filename`. I was OK with only using the IS/PSY version but found some value in your argument. Your opinion today implies that we should delete the PSI version altogether. Right?

you are correct, I should not export it in PSY and have the users call PSY.configure_logging if they want to change the setting.

.github/workflows/pr_testing.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     runs-on: ${{ matrix.os }}
```
jd-lara, line 10: @daniel-thom can you please do these changes to the same file in PSY whenever you do your PR.

daniel-thom (approved):

PR #542, opened 2020-11-12
Fixing PTDF balance constraint

sourabhdalvi:


src/network_models/ptdf_model.jl
```diff
@@ -8,8 +8,7 @@ function ptdf_networkflow(
     time_steps = model_time_steps(psi_container)
     network_flow =
         add_cons_container!(psi_container, :network_flow, PTDF.axes[1], time_steps)
-    nodal_balance =
-        add_cons_container!(psi_container, :nodal_balance, PTDF.axes[2], time_steps)
+    nodal_balance = add_cons_container!(psi_container, :nodal_balance, time_steps)
```
jd-lara, line 11:
```suggestion
    nodal_balance = add_cons_container!(psi_container, :system_balance, time_steps)
```

src/network_models/ptdf_model.jl
```diff
@@ -62,13 +61,10 @@ function ptdf_networkflow(
...
         end
 
-        for b in buses
-            number = PSY.get_number(b)
-            nodal_balance[number, t] = JuMP.@constraint(
-                psi_container.JuMPmodel,
-                nodal_balance_expressions[number, t] == 0
-            )
-        end
+        nodal_balance[t] = JuMP.@constraint(
```
jd-lara, line 64:
```suggestion
        system_balance[t] = JuMP.@constraint(
```

test/test_network_constructors.jl
```diff
@@ -103,6 +104,8 @@ end
         psi_constraint_test(ps_model, constraint_names)
         psi_checkobjfun_test(ps_model, objfuncs[ix])
         psi_checksolve_test(ps_model, [MOI.ALMOST_OPTIMAL, MOI.OPTIMAL])
+        # TODO: add values from Bθ formulation to compare
```
jd-lara, line 107: Please add this test before merging

test/test_network_constructors.jl
```diff
@@ -162,6 +165,46 @@ end
     end
 end
 
+@testset "Test Locational Marginal Prices between DC lossless with PowerModels vs StandardPTDFModel" begin
+    network = [DCPPowerModel, StandardPTDFModel]
+    sys = build_system("c_sys5")
+    dual_constraint = [
+        [PSI.make_constraint_name(PSI.NODAL_BALANCE_ACTIVE, PSY.Bus)],
```
jd-lara, line 172: It's better not to expose internal methods here. We should add a better interface to request the duals than passing the symbols but that is another issue.

test/test_network_constructors.jl
```diff
@@ -162,6 +165,46 @@ end
...
+        ps_model = OperationsProblem(
+            TestOpProblem,
+            net,
+            sys;
+            optimizer = OSQP_optimizer,
+            use_parameters = p,
+            PTDF = ptdf,
+            constraint_duals = dual_constraint[ix],
+        )
+        construct_device!(ps_model, :Thermal, thermal_model)
```
jd-lara, line 189: Why is this test calling `construct_device` here? Please use a template in the test.

jd-lara (changes requested):
The code in the tests needs some refactoring to make it cleaner.

test/test_network_constructors.jl
sourabhdalvi: this is similar to what you have in other tests in that file, so I just followed that

test/test_network_constructors.jl
sourabhdalvi: I can replace it with just symbols

test/test_network_constructors.jl
sourabhdalvi: I added that test but just forgot to remove that TODO

test/test_network_constructors.jl
jd-lara: Some of the tests are old before the implementation of Templates, we need to refactor the testing to remove the individual construct device pattern

test/test_network_constructors.jl
jd-lara: Yes, that's better. This PR reflected some pending work with the interfaces

jd-lara (approved):
@sourabhdalvi can you change name of the constraint here please to make them consistent? Also, use the same syntax here calling ``add_cons_container!(psi_container, :system_balance, time_steps)`` instead of assign? 

https://github.com/NREL-SIIP/PowerSimulations.jl/blob/2b8689d3ea98722aefa5f7810a492488271a5fd6/src/network_models/copperplate_model.jl#L6

jd-lara (approved):

PR #556, opened 2020-11-30
handling time series of int values

claytonpbarrows:
This resolves an issue that I encountered where PSI was expecting a vector of floats but instead gets a vector of ints. This makes the conversion when needed.


daniel-thom (approved):

jd-lara (changes requested):
I see why we want to do this, but probably the best approach would be to add a method to DeviceTimeSerieConstraintInfo that allow any `Vector{Number}` and makes the conversion.

jd-lara (approved):

PR #560, opened 2020-12-02
Change method dispatch

kdheepak:
- HydroGen
- ElectricLoads
- RenewableGen
- Storage
- HydroPumpedStorage
- FlowActivePowerVariable


src/devices_models/devices/storage.jl
```diff
@@ -1,66 +1,43 @@
...
+############## ReactivePowerVariable, Storage ####################
+
+get_variable_binary(::ReactivePowerVariable, ::Type{<:PSY.Storage}) = false
+
+get_variable_expression_name(::ReactivePowerVariable, ::Type{<:PSY.Storage}) = :nodal_balance_reactive
+
+############## EnergyVariable, Storage ####################
+
+get_variable_binary(::EnergyVariable, ::Type{<:PSY.Storage}) = false
+# TODO: why isn't this defined
```
daniel-thom, line 32: Do these need to be implemented before merging?

src/PowerSimulations.jl
```diff
@@ -1,6 +1,8 @@
 isdefined(Base, :__precompile__) && __precompile__()
 module PowerSimulations
 
+using InteractiveUtils
```
jd-lara, line 4: why using here and not Import? is this a left over from other stuff?

src/network_models/network_constructor.jl
```diff
@@ -54,7 +54,7 @@ function construct_network!(
             btype,
             Vector([[b for b in dc_branches if typeof(b) == btype]]),
         )
-        flow_variables!(psi_container, StandardPTDFModel, typed_dc_branches)
+        add_variables!(psi_container, StandardPTDFModel(), typed_dc_branches)
```
daniel-thom, line 57: Why did you change this to an instance instead of a type?

src/core/psi_container.jl
```diff
@@ -17,7 +17,7 @@ mutable struct PSIContainer
         settings::PSISettings,
         jump_model::Union{Nothing, JuMP.AbstractModel},
     )
-        #PSY.check_forecast_consistency(sys)
+        # PSY.check_forecast_consistency(sys)
```
jd-lara, line 20: We can remove this line, this is no longer a PSY function.

daniel-thom (commented):
Overall this is exactly like what you proposed and looks good. Just a few minor comments.

src/devices_models/devices/DC_branches.jl
```diff
@@ -14,9 +14,9 @@ flow_variables!(
     devices::IS.FlattenIteratorWrapper{<:PSY.DCBranch},
 ) = nothing
 
-function flow_variables!(
+function add_variables!(
     psi_container::PSIContainer,
-    system_formulation::Type{StandardPTDFModel},
+    system_formulation::StandardPTDFModel,
```
jd-lara, line 19: The system formulation variable is not used here. It can be removed.

src/devices_models/devices/common/add_variable.jl
```diff
@@ -52,7 +6,11 @@ function add_variables!(
     ::Type{T},
     devices::Union{Vector{U}, IS.FlattenIteratorWrapper{U}},
 ) where {T <: VariableType, U <: PSY.Component}
-    _add_variables!(psi_container, devices, AddVariableSpec(T, U, psi_container))
+    add_variable!(
+        psi_container,
+        T(),
```
jd-lara, line 11: Is it better to evaluate here instead of passing the type? I don't understand the reason for doing that instead of passing `T`. The code style in PSI has used T as type so far.

src/devices_models/devices/common/add_variable.jl
```diff
@@ -62,47 +20,13 @@ function add_variables!(
...
-    else
-        filter_func = spec.devices_filter_func
-    end
-    if !isnothing(filter_func)
-        devices = filter!(filter_func, collect(devices))
-    end
-
     add_variable!(
         psi_container,
+        T(),
```
jd-lara, line 27: same here

src/devices_models/devices/common/range_constraint.jl
```diff
@@ -150,7 +150,7 @@ function device_semicontinuousrange!(
     return
 end
 
-#This function looks suspicious and repetitive. Needs verification
+# This function looks suspicious and repetitive. Needs verification
```
jd-lara, line 153: Is this comment still valid?

src/devices_models/devices/electric_loads.jl
```diff
@@ -4,44 +4,31 @@ struct StaticPowerLoad <: AbstractLoadFormulation end
...
-    return AddVariableSpec(;
-        variable_name = make_variable_name(T, U),
-        binary = false,
-        expression_name = :nodal_balance_reactive,
-        sign = -1.0,
-        lb_value_func = x -> 0.0,
-        ub_value_func = x -> PSY.get_max_reactive_power(x),
-    )
-end
+get_variable_sign(_, ::Type{<:PSY.ElectricLoad}) = -1.0
```
jd-lara, line 9: we have been using empty spaces for this. If we are moving into this style is important to document it and make it consistent.

src/devices_models/devices/electric_loads.jl
```diff
@@ -4,44 +4,31 @@ struct StaticPowerLoad <: AbstractLoadFormulation end
...
-    ::Type{U},
-    ::PSIContainer,
-) where {T <: OnVariable, U <: PSY.ElectricLoad}
-    return AddVariableSpec(; variable_name = make_variable_name(T, U), binary = true)
-end
+########################### ActivePowerVariable, ElectricLoad ####################################
+
+get_variable_binary(::ActivePowerVariable, ::Type{<:PSY.ElectricLoad}) = false
+
+get_variable_expression_name(::ActivePowerVariable, ::Type{<:PSY.ElectricLoad}) = :nodal_balance_active
```
jd-lara, line 15: This part could be problematic in the future as a variable could be added to different expressions. I think we talked about it before. For now is ok as nothing is breaking but this could require making It a function of the formulation too.

src/network_models/network_constructor.jl
jd-lara, line 57: Same comment as above.

jd-lara (commented):
Overall looks great, I have some questions about the changes made. Mostly there are some patterns that aren't consistent with the rest of the code base that I think we should make consistent or avoid.

src/core/psi_container.jl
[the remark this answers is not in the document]
kdheepak: Resolved in c359a46a.

src/devices_models/devices/common/add_variable.jl
```diff
@@ -222,3 +205,44 @@ function set_variable_bounds!(
...
+    else
+        initial_value = nothing
+    end
+
+    add_variable!(
+        psi_container,
+        OnVariable(),
+        devices,
+    )
+    varstatus = get_variable(psi_container, OnVariable, PSY.ThermalMultiStart)
```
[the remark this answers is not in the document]
kdheepak: Resolved in c359a46a.

src/devices_models/devices/storage.jl
kdheepak: I discussed this with @jd-lara and he said it wasn't required. I forgot to remove the comments.

Resolved in c359a46a.

src/PowerSimulations.jl
kdheepak: This is a leftover from a debugging session. Resolved in f877cf4d.

src/core/psi_container.jl
kdheepak: Resolved in c359a46.

src/devices_models/devices/common/add_variable.jl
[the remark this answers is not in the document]
kdheepak: @jd-lara has made similar comments, so I'm addressing them all here.

In this PR we are doing the following:

```julia
get_variable_lower_bound(::ActivePowerVariable, d::PSY.ThermalGen, _) = PSY.get_active_power_limits(d).min
get_variable_upper_bound(::ActivePowerVariable, d::PSY.ThermalGen, _) = PSY.get_active_power_limits(d).max
```

If we passed just the type, these methods would become this instead:

```julia
get_variable_lower_bound(::Type{<:ActivePowerVariable}, d::PSY.ThermalGen, _) = PSY.get_active_power_limits(d).min
get_variable_upper_bound(::Type{<:ActivePowerVariable}, d::PSY.ThermalGen, _) = PSY.get_active_power_limits(d).max
```

I think this is more complicated to read and understand for the average user.

Also, previously, we were always passing the types and the types alone when creating a "VariableSpec" (i.e. we were using `::Type{<:ThermalGen}` and returning anonymous functions that accepted an arugment to `d::PSY.ThermalGen`). 

Now however, we are using an instance of the object and hence have to define `d::PSY.ThermalGen`. Defining the other argument as `ActivePowerVariable` felt more natural to me.

I think this change to using instances in cases where the compiler can make optimizations makes the code easier to read, easier to write, and has no cost to performance.

Happy to discuss this further in a call.

src/devices_models/devices/common/add_variable.jl
```diff
@@ -141,25 +65,79 @@ If binary = true:
...
+    for t in time_steps, d in devices
+        name = PSY.get_name(d)
+        variable[name, t] = JuMP.@variable(
+            psi_container.JuMPmodel,
+            base_name = "$(var_name)_{$(name), $(t)}",
+            binary = binary
+        )
+
+        ub = get_variable_upper_bound(variable_type, d, psi_container.settings)
+        !isnothing(ub) && JuMP.set_upper_bound(variable[name, t], ub)
```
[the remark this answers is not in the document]
kdheepak: Makes sense. Would you like me to make these changes in this PR? Current status of the codebase is as follows:

```plaintext
$ rg isnothing src --count-matches | wc -l
23

$ rg '=== nothing' src --count-matches | wc -l
2

$ rg '!== nothing' src --count-matches | wc -l
0
```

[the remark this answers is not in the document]
daniel-thom: I personally started writing `===` instead of `isnothing` but didn't sweep through the codebases and change all occurrences. I also don't know if Jose came to the same conclusion.

src/devices_models/devices/common/add_variable.jl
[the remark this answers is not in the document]
daniel-thom: I was mostly just asking why you made the change. Given that rationale, why didn't you change `get_variable_binary`, `get_variable_expression_name`, and `get_variable_sign` to accept instances of PSY.Component? Instances are available at the call sites.

As Jose said, PSI does heavily use a paradigm of passing the type to functions where no instance would ever be used.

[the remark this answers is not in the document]
kdheepak: > Given that rationale, why didn't you change get_variable_binary, get_variable_expression_name, and get_variable_sign to accept instances of PSY.Component? Instances are available at the call sites.

That would be clearer. If @jd-lara is on board, I'll make that change.

src/devices_models/devices/DC_branches.jl
kdheepak: Resolved in 40e2de21.

src/devices_models/devices/common/range_constraint.jl
kdheepak: Probably not as originally intended but when I refactor range constraints I'll remove this comment. This will be resolved in a future PR.

src/devices_models/devices/electric_loads.jl
kdheepak: Can you clarify what you've been using empty spaces for?

src/devices_models/devices/electric_loads.jl
kdheepak: We can extend it to support future use cases.

src/devices_models/devices/DC_branches.jl
kdheepak: Resolved in 461c2777 instead.

src/devices_models/devices/common/add_variable.jl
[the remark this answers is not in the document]
jd-lara: I prefer we keep the current paradigm. Most of the collaborators and developers are already used to this. It is also consistent with the interfaces in PSY. For instance, get_component() uses a type as input. Most of our current users are already used to passing types and not instances.

src/devices_models/devices/common/add_variable.jl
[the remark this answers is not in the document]
jd-lara: making the change in this PR is appropriate.

src/devices_models/devices/electric_loads.jl
jd-lara: when the argument isn't used in the specific function for instance 

```julia
function foo(val::Int, other_val::Int)
return val + other_val
end

function foo(val::Int, ::Any)
return val
end
```

src/devices_models/devices/DC_branches.jl
```diff
@@ -127,7 +127,7 @@ function branch_rate_constraints!(
     psi_container::PSIContainer,
     devices::IS.FlattenIteratorWrapper{B},
     model::DeviceModel{B, <:AbstractDCLineFormulation},
-    system_formulation::Union{
+    ::Union{
```
jd-lara, line 130: This union is redundant (I know you didn't introduce it). 

since 

```julia 
PM.AbstractActivePowerModel <: PM.AbstractPowerModel
```

src/devices_models/devices/hydro_generation.jl
```diff
@@ -11,164 +11,95 @@ struct HydroCommitmentRunOfRiver <: AbstractHydroUnitCommitment end
...
-        expression_name = :nodal_balance_reactive,
-        initial_value_func = x -> PSY.get_reactive_power(x),
-        lb_value_func = x -> PSY.get_reactive_power_limits(x).min,
-        ub_value_func = x -> PSY.get_reactive_power_limits(x).max,
-    )
-end
+get_variable_initial_value(pv::ActivePowerVariable, d::PSY.HydroGen, settings) =
+    get_variable_initial_value(pv, d, get_warm_start(settings) ? WarmStartVariable() : ColdStartVariable())
+get_variable_initial_value(::ActivePowerVariable, d::PSY.HydroGen, ::WarmStartVariable) = PSY.get_active_power(d)
+get_variable_initial_value(::ActivePowerVariable, d::PSY.HydroGen, ::ColdStartVariable) = nothing
```
jd-lara, line 22: we don't have coldstart for Hydro. This is only for thermals.

src/devices_models/devices/hydro_generation.jl
```diff
@@ -11,164 +11,95 @@ struct HydroCommitmentRunOfRiver <: AbstractHydroUnitCommitment end
...
-        variable_name = make_variable_name(T, U),
-        binary = false,
-        expression_name = :nodal_balance_reactive,
-        initial_value_func = x -> PSY.get_reactive_power(x),
-        lb_value_func = x -> PSY.get_reactive_power_limits(x).min,
-        ub_value_func = x -> PSY.get_reactive_power_limits(x).max,
-    )
-end
+get_variable_initial_value(pv::ActivePowerVariable, d::PSY.HydroGen, settings) =
+    get_variable_initial_value(pv, d, get_warm_start(settings) ? WarmStartVariable() : ColdStartVariable())
```
jd-lara, line 20: same here

src/devices_models/devices/hydro_generation.jl
```diff
@@ -11,164 +11,95 @@ struct HydroCommitmentRunOfRiver <: AbstractHydroUnitCommitment end
...
-) where {T <: EnergyVariable, U <: PSY.HydroGen}
-    return AddVariableSpec(;
-        variable_name = make_variable_name(T, U),
-        binary = false,
-        initial_value_func = x -> PSY.get_initial_storage(x),
-        lb_value_func = x -> 0.0,
-        ub_value_func = x -> PSY.get_storage_capacity(x),
-    )
-end
+get_variable_lower_bound(::ActivePowerVariable, d::PSY.HydroGen, _) = PSY.get_active_power_limits(d).min
```
jd-lara, line 24: Can you make another function here that dispatches on the formulation `HydroDispatchRunOfRiver` to return 0.0. This is related to #563.

src/devices_models/devices/common/add_variable.jl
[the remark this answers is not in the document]
kdheepak: Resolved in https://github.com/NREL-SIIP/PowerSimulations.jl/pull/560/commits/7fb5eda67b7873f612ac8be80e7f358bea57c2ca.

src/devices_models/devices/common/add_variable.jl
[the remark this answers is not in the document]
kdheepak: After having discussions over the phone and chat, we decided that we should make the type / instance changes in this case. The current code is backward compatible and will not affect users. We also decided that we would approach future type / instance changes on a case by case basis.

Some references:

https://docs.julialang.org/en/v1/manual/style-guide/#Avoid-confusion-about-whether-something-is-an-instance-or-a-type

https://docs.julialang.org/en/v1/manual/style-guide/#Don't-use-unnecessary-static-parameters

src/devices_models/devices/electric_loads.jl
kdheepak: I believe we discussed that using `_` is less noisy that using `::Any`. Can someone refresh my memory?

src/devices_models/devices/hydro_generation.jl
kdheepak: Resolved in https://github.com/NREL-SIIP/PowerSimulations.jl/pull/560/commits/20ab12eff7f5adf899419a65ff2d2587a4cb9a95

src/devices_models/devices/DC_branches.jl
kdheepak: Resolved in https://github.com/NREL-SIIP/PowerSimulations.jl/pull/560/commits/20ab12eff7f5adf899419a65ff2d2587a4cb9a95

src/devices_models/devices/hydro_generation.jl
kdheepak: Resolved in https://github.com/NREL-SIIP/PowerSimulations.jl/pull/560/commits/20ab12eff7f5adf899419a65ff2d2587a4cb9a95.

src/devices_models/devices/hydro_generation.jl
kdheepak: Resolved in https://github.com/NREL-SIIP/PowerSimulations.jl/pull/560/commits/28d99027f2bd2e4140806e0cdb92ebfed64f50cf.

src/devices_models/devices/electric_loads.jl
daniel-thom: I'm not sure that we reached consensus. I have no idea what Julia people would say. Here are my fairly weak opinions. I would go along with anything.
- In general passing arguments to functions that don't get used can be indicative of a sub-optimal design. In the cases here I think it makes sense. Hopefully, we don't have this pattern in many places.
- If it is a one-line function then `_` is less noisy.
- If it is a multi-line function then including the name of the variable at least informs the reader of what is being passed (to match the interface).

```
foo(val::Int, _) = val

function foo(val::Int, other_val)
    return val
end
```

jd-lara: Just make sure that all the code base is consistent. I prefer not to use _ and pass the expected type.

daniel-thom (approved):
Looks great. Very nice improvement.

src/devices_models/devices/electric_loads.jl
kdheepak: > If it is a one-line function then _ is less noisy.

I agree with this. Let me made this change for just one line functions.

kdheepak: Okay, it is only implemented for one line functions.

jd-lara (approved):
There are some areas where this code will have some conflicts with the simulation store code. Mostly over white space changes.

src/devices_models/devices/electric_loads.jl
kdheepak: > In general passing arguments to functions that don't get used can be indicative of a sub-optimal design.

These methods are specialized. Here's two examples:

Call site:

```julia
lb = get_variable_upper_bound(variable_type, d, psi_container.settings)
...
init = get_variable_initial_value(variable_type, d, psi_container.settings)
```

For `get_variable_initial_value` the function is defined like this:

```julia
get_variable_initial_value(pv::ActivePowerVariable, d::PSY.ThermalMultiStart, settings) =
    get_variable_initial_value(pv, d, get_warm_start(settings) ? WarmStartVariable() : ColdStartVariable())
get_variable_initial_value(::ActivePowerVariable, d::PSY.ThermalMultiStart, ::WarmStartVariable) = PSY.get_active_power(d)
get_variable_initial_value(::ActivePowerVariable, d::PSY.ThermalMultiStart, ::ColdStartVariable) = nothing
```

For `get_variable_upper_bound` the function is defined like this:

```julia
get_variable_upper_bound(::ActivePowerVariable, d::PSY.ThermalMultiStart, _) = PSY.get_active_power_limits(d).max
```

My thinking here is that:

1) `settings` is untyped in `get_variable_initial_value` so it should be untyped in `get_variable_upper_bound`. 
2) `::WarmStartVariable` / `::ColdStartVariable` is used for dispatch in `get_variable_initial_value`. We leave out the variable name here because it doesn't matter, and I'm doing the same for `get_variable_upper_bound`.

src/devices_models/devices/hydro_generation.jl
kdheepak: Can you clarify exactly what needs to be added here?

jd-lara: add 

```julia
get_variable_lower_bound(::ActivePowerVariable, d::PSY.HydroGen, ::HydroDispatchRunOfRiver) = 0.0
```

kdheepak: Resolved in https://github.com/NREL-SIIP/PowerSimulations.jl/pull/560/commits/7bf83ef6f1dc9396039e3358d0a44824f10690a5

src/devices_models/devices/electric_loads.jl
kdheepak: I have one last point that I have to make on this. I think it is not clear what the name of this variable should be. At the call site, we might be passing in a formulation, an instance of a formulation, the settings object etc. That is another reason I think using an underscore makes the most sense.

kdheepak: I think we can form a consensus on the use of `_` in this fashion in a future PR. I'll go ahead and merge this for now.

src/devices_models/devices/renewable_generation.jl
```diff
@@ -1,34 +1,26 @@
...
-        binary = false,
-        expression_name = :nodal_balance_reactive,
-    )
-end
+get_variable_binary(::ActivePowerVariable, ::Type{<:PSY.RenewableGen}) = false
+
+get_variable_expression_name(::ActivePowerVariable, ::Type{<:PSY.RenewableGen}) = :nodal_balance_active
+
+get_variable_lower_bound(::ActivePowerVariable, d::PSY.RenewableGen, _) = 0.0
+get_variable_upper_bound(::ActivePowerVariable, d::PSY.RenewableGen, _) = PSY.get_active_power(d)
```
jd-lara, line 15:
```suggestion
get_variable_upper_bound(::ActivePowerVariable, d::PSY.RenewableGen, _) = PSY.get_max_active_power(d)
```

PR #566, opened 2020-12-09
Cost Function re-factor

sourabhdalvi:
Now the unit conversion is done in the cost function and the cost data is expected as follows
- variable_cost::Float64(units = $/MW, )
- variable_cost::Tuple(units: a = $/MW**2, b = $/MW )
- variable_cost::Tuple(units: powerpoint = MW, cost = $)
- fixed::Float64 (units = $)
- no_load::Float64 (units = $)
- start_up:: Float64  (units = $)
- shut_down:: Float64  (units = $)

Required x-ref https://github.com/NREL-SIIP/PowerSystems.jl/pull/679


jd-lara:
Can you please add a Documentation markdown page about this so we don't forget to explain properly.

sourabhdalvi:
> Can you please add a Documentation markdown page about this so we don't forget to explain properly.

Yes, I'll add this to the Docs

jd-lara (approved):

PR #567, opened 2020-12-10
sim store integration

jd-lara:
Preliminary PR with initial changes. 

Summary of changes up to this point. 

1. Removes all uses of Arrow files 
2. Add several getter functions to different places 
3. Add several fields to some of the store objects. 
4. Initial approach to the simulation results object
5. addresses some deprecations in the use of HDF5 
6. Makes some changes in the stages and simulations to have a cleaner execution process. 
7. implements graceful exit infeasibilities are present.


src/core/simulation_results.jl
```diff
@@ -1,616 +1,130 @@
...
-        for (ix, time) in enumerate(date_df.Date)
-            file_path = date_df[ix, :File_Path]
-            var = read_arrow_file(file_path)
-            variables[(variable[l])] = vcat(variables[(variable[l])], var[1:time_length, :])
-            if l == 1
-                time_stamp = vcat(time_stamp, _read_time(file_path, time_length))
-                check_file_integrity(dirname(file_path))
-            end
-        end
+function results_pre_process(store::HdfSimulationStore, stage_name::Symbol)
```
jd-lara, line 30: @daniel-thom I opened the PR to see if the use of the data store is correct here. The intention is to first get the available information about the system to only open queries to the store that are valid.

src/core/hdf_simulation_store.jl
```diff
@@ -458,19 +473,23 @@ function _get_columns_dataset(store::HdfSimulationStore, key)
     return getfield(store.datasets[key.stage], key.type)[key.name].columns
 end
 
-function _get_dataset(::Type{OptimizerStats}, store::HdfSimulationStore)
+function get_dataset(::Type{OptimizerStats}, store::HdfSimulationStore)
     return store.file[OPTIMIZER_DATASET_PATH]
 end
 
-function _get_dataset(store::HdfSimulationStore, key)
+function get_dataset(store::HdfSimulationStore, stage_name::Symbol)
```
jd-lara, line 480: @daniel-thom I am not sure if you might approve of this. I welcome suggestions about how to get the available symbols in a dataset.

src/core/simulation_results.jl
```diff
@@ -100,23 +100,21 @@ function SimulationResults(
                              )
 end
 
-function SimulationResults(sim::Simulation, stage_name::String; results_output_path = nothing)
-    simulation_store_path = joinpath(get_store_dir(sim), "data_store")
+function SimulationResults(sim::Simulation, stage_name::String)
```
jd-lara, line 103: @claytonpbarrows and @daniel-thom under this scheme now we read the results directly from the simulation object

jd-lara:
The simulation store changes reduced the simulation times significantly. Now the most amount of time is spent in the stage updates 

image

src/core/hdf_simulation_store.jl
```diff
@@ -31,6 +31,10 @@ mutable struct StageDatasets
     variables::Dict{Symbol, Dataset}
 end
 
+get_duals(stage_data_set::StageDatasets) = stage_data_set.duals
```
daniel-thom, line 34: nit: `dataset` is one word, so no underscore.

src/core/simulation.jl
```diff
@@ -444,12 +467,14 @@ function _check_steps(
...
         stage = sim.stages[stage_name]
         execution_counts = get_executions(stage)
-        transitions =
-            sim.sequence.execution_order[vcat(1, diff(sim.sequence.execution_order)) .== 1]
-        @assert length(findall(x -> x == stage_number, sim.sequence.execution_order)) /
+        transitions = get_sequence(sim).execution_order[vcat(
+            1,
+            diff(get_sequence(sim).execution_order),
+        ) .== 1]
+        @assert length(findall(x -> x == stage_number, get_sequence(sim).execution_order)) /
```
daniel-thom, line 477: Consider using `assert_op`. It will print the values if this trips.

src/core/simulation.jl
```diff
@@ -536,6 +559,7 @@ function build!(
     recorders = [],
     console_level = Logging.Error,
     file_level = Logging.Info,
+    serialize = true,
```
daniel-thom, line 562: Need to add a docstring.

src/core/simulation.jl
```diff
@@ -1066,8 +1102,19 @@ function _initialize_stage_storage!(sim::Simulation, store, cache_size_mib)
         num_rows = num_executions * get_steps(sim)
 
         # TODO DT: not sure this is correct
-        resolution = intervals[stage_name][1]
-        stage_params = SimulationStoreStageParams(num_executions, horizon, resolution)
+        interval = intervals[stage_name][1]
```
daniel-thom, line 1105: This resolves the comment on the previous line.

src/core/simulation_results.jl
daniel-thom: This should accept the abstract type `SimulationStore`. I think we should keep the interfaces to the store very generic. `read_result`, `write_result`, etc. `get_dataset` would be problematic if we changed to an implementation that does not have `datasets`.

src/core/simulation_results.jl
```diff
@@ -1,616 +1,133 @@
...
-        results_folder,
-        param_values,
-    )
+    stage_params = stages_params[stage_name]
+    stage_results_params["base_power"] = get_base_power(stage_params)
+    stage_results_params["system_uuid"] = get_system_uuid(stage_params)
+    stage_dataset = get_dataset(store, stage_name)
+    stage_results_params["available_variables"] = keys(get_variables(stage_dataset))
+    stage_results_params["available_params"] = keys(get_parameters(stage_dataset))
+    stage_results_params["available_duals"] = keys(get_duals(stage_dataset))
```
daniel-thom, line 44: We need a better interface for this. I expected that we _might_ need something like this and left this comment in the HDF file:
```
# TODO: Interfaces to add
# - list_stages(store)
# - list_names(store, type) for duals/parameters/variables by stage
# - get_step_resolution(store)
# - get_resolution(store, stage)
```
What you think about those names?

daniel-thom (commented):
First pass.

src/core/simulation_results.jl
jd-lara: What is type in `# - list_names(store, type) for duals/parameters/variables by stage` ? 

Otherwise, the names look good to me.

daniel-thom: It should be one of these constants: https://github.com/NREL-SIIP/PowerSimulations.jl/blob/2bbbc790c1a425c4fe792ef697e8be81e4bf8d75/src/core/simulation_store.jl#L1

jd-lara: I would prefer not to use the word type here since it can be confusing. I thought that these will be structs

src/core/simulation_results.jl
jd-lara: Do we need the additional interfaces for this change?

src/core/hdf_simulation_store.jl
daniel-thom: I don't think this should be called outside of this file. We can leave it for now and I'll refactor next week when I add the `list_names` and other interfaces.

src/core/simulation.jl
```diff
@@ -977,33 +1010,40 @@ function _execute!(sim::Simulation, store; cache_size_mib = 1024, kwargs...)
...
-                                stage,
-                                get_current_time(sim),
-                                raw_results_path,
-                                store,
-                            )
+                            settings = get_settings(stage)
+                            # TODO: Add Fallback when run_stage fails
+                            status = run_stage!(step, stage, get_current_time(sim), store)
+                            if status == SUCESSFUL_RUN
+                                flush(store)
```
daniel-thom, line 1033: I think that we should flush once the stage is finished. Not after every execution.

src/core/simulation.jl
```diff
@@ -1034,12 +1074,10 @@ function _execute!(sim::Simulation, store; cache_size_mib = 1024, kwargs...)
             end
         end
         flush(store)
```
daniel-thom, line 1076: Can be deleted.

src/core/simulation.jl
jd-lara: This is when the stage is done. If a stage runs 96 times it will only flush at the end and if the 87th run fails then it will not have those values.

src/core/simulation_results.jl
daniel-thom: The SimulationStore functions `read_result`, `write_result`, and `list_names` need to be passed arguments specifying dual/parameter/variable and the symbol name. Let me know your preference.

src/core/simulation_results.jl
```diff
@@ -1,616 +1,133 @@
...
-        throw(IS.HashMismatchError("The hash value in the written files does not match the read files, results may have been tampered."))
+    if execution === nothing
+        execution = maximum([parse(Int, f) for f in readdir(path) if occursin(r"^\d+$", f)])
+    end
+    execution_path = joinpath(path, string(execution))
+    if !isdir(execution_path)
+        error("Execution $execution not in the simulations results")
+    end
+    if !check_folder_integrity(execution_path)
+        @warn("The results folder $(execution_path) is not consistent with the default folder structure. This can lead to unwanted to errors or unwanted results")
```
daniel-thom, line 72:
```suggestion
        @warn("The results folder $(execution_path) is not consistent with the default folder structure. This can lead to errors or unwanted results")
```

src/core/hdf_simulation_store.jl
jd-lara: ok, yeah I wasn't sure the best approach here to get this but now it has been clarified.

src/core/simulation_results.jl
```diff
@@ -1,616 +1,133 @@
...
+        Dict{Symbol, DataFrames.DataFrame}(),
     )
-    return results
 end
+
+get_model_base_power(result::SimulationResults) = result.base_power
+IS.get_variables(result::SimulationResults) = result.variable_values
+#IS.get_total_cost(result::SimulationResults) = result.total_cost
+#IS.get_optimizer_log(results::SimulationResults) = results.optimizer_log
+#IS.get_timestamp(result::SimulationResults) = result.time_stamp
```
daniel-thom, line 131: Are these still needed?

jd-lara: Yes, I am re-implemeting now.

src/core/simulation_results.jl
jd-lara: value? 

```Julia
const CONTAINER_DUALs_VALUE = :duals
const CONTAINER_PARAMETERS_VALUE = :parameters
const CONTAINER_VARIABLES_VALUE = :variables
const CONTAINER_VALUE =
    Set((CONTAINER_DUALs_VALUE,CONTAINER_PARAMETERS_VALUE, CONTAINER_VARIABLES_VALUE))
```

daniel-thom: CONTAINER_DUALS, CONTAINER_PARAMETERS, CONTAINER_VARIABLES

jd-lara: Sounds good

jd-lara:
@daniel-thom This PR is passing the tests. There are still pending items but it should enable the next step of work to continue when merged

src/core/simulation.jl
```diff
@@ -977,33 +1010,39 @@ function _execute!(sim::Simulation, store; cache_size_mib = 1024, kwargs...)
...
-                            )
+                            settings = get_settings(stage)
+                            # TODO: Add Fallback when run_stage fails
+                            status = run_stage!(step, stage, get_current_time(sim), store)
+                            if status == SUCESSFUL_RUN
+                            elseif !get_allow_fails(settings) && (status != SUCESSFUL_RUN)
+                                set_simulation_status!(sim, FAILED_RUN)
+                                break
+                            elseif get_allow_fails(settings) && (status != SUCESSFUL_RUN)
+                                continue
```
daniel-thom, line 1037: Commentary: In the passing case you proceed through this block. In this case where there is an allowed failure you immediately continue the nearest for loop. A lot of logic gets skipped. If it is correct to do this then a comment is probably warranted. The logic may be unnecessarily complex.

jd-lara: Yeah, if the fail is allowed, the remaining steps should be skipped. We could probably refactor the function to simplify it. I think that should be done in the next iteration

jd-lara:
I will have to add more tests before merging into master, but the current ones cover the basic cases we need to continue development.

src/core/simulation.jl
```diff
@@ -977,33 +1010,39 @@ function _execute!(sim::Simulation, store; cache_size_mib = 1024, kwargs...)
...
+                            elseif status == RUNNING
+                                set_simulation_status!(sim, FAILED_RUN)
+                                error("Stage still in RUNNING status")
+                            else
+                                set_simulation_status!(sim, FAILED_RUN)
+                                error("Invalid stage status")
+                            end
+
                             sim.internal.run_count[step][stage_number] += 1
```
daniel-thom, line 1046: This line should be moved down to line 1064 with the other increments. Also, you are saying that it is acceptable to skip those increments if there is an allowable failure. I'm not saying it's wrong; it just seems odd to a reader that is not completely familiar with the inner workings of the code.

src/core/simulation_results.jl
```diff
@@ -1,517 +1,378 @@
...
-sim = Simulation("Test", 7, stages, "/Users/yourusername/Desktop/"; system_to_file = false)
-execute!(sim::Simulation; kwargs...)
-references = make_references(sim, "2019-10-03T09-18-00-test")
-```
-"""
-function make_references(sim::Simulation, date_run::String)
-    for stage_number in keys(sim.internal.date_ref)
-        sim.internal.date_ref[stage_number] = sim.initial_time
+""" Holds the results of the simulation for plotting or exporting"""
+function SimulationResults(
```
daniel-thom, line 69: Suggest moving all constructors immediately below the struct definition.

src/core/simulation_results.jl
```diff
@@ -1,517 +1,378 @@
...
+    end
+    return false
 end
 
-# internal function for differentiating variables from duals in file names
-function _concat_dual(duals::Vector{Symbol})
-    dual = []
-    for d in duals
-        dual = vcat(dual, "dual_" * String(d))
+function _results_pre_process(store::HdfSimulationStore, stage_name::Symbol)
```
daniel-thom, line 32: This should accept the abstract type `SimulationStore`.

src/core/simulation_results.jl
```diff
@@ -1,517 +1,378 @@
...
+mutable struct SimulationResults <: PSIResults
+    stage::String
+    base_power::Float64
+    execution_path::AbstractString
+    results_output_folder::String
+    existing_timestamps::StepRange{Dates.DateTime, Dates.Millisecond}
+    results_timestamps::Vector{Dates.DateTime}
+    system::Union{Nothing, PSY.System}
+    existing_variables::Array{Symbol}
+    variable_values::Dict{Symbol, SortedDict{Dates.DateTime, DataFrames.DataFrame}}
```
daniel-thom, line 13: Consider using type alias(es).

src/core/simulation_results.jl
```diff
@@ -1,517 +1,378 @@
...
+
+mutable struct SimulationResults <: PSIResults
+    stage::String
+    base_power::Float64
+    execution_path::AbstractString
+    results_output_folder::String
+    existing_timestamps::StepRange{Dates.DateTime, Dates.Millisecond}
+    results_timestamps::Vector{Dates.DateTime}
+    system::Union{Nothing, PSY.System}
+    existing_variables::Array{Symbol}
```
daniel-thom, line 12: Would it be possible to remove these `existing_` members by instead storing them inside the `_values` members? At initial construction you could add each field to the outer dict. The inner dict would remain empty until the user performs a read.

src/core/simulation_results.jl
```diff
@@ -1,517 +1,378 @@
...
-                time_stamp = vcat(time_stamp, _read_time(file_path, time_length))
-                check_file_integrity(dirname(file_path))
-            end
+    results_dict = get_results(res)
+    current_names = keys(results_dict)
+    additional_names = setdiff(names, current_names)
+    current_timestamps = res.results_timestamps
+
+    if !isempty(additional_timestamps) && !isempty(additional_names)
+        @info "Requested new names and times"
```
daniel-thom, line 240: Should the info messages in this function be debug?

src/core/simulation_results.jl
```diff
@@ -1,517 +1,378 @@
...
 """
-function load_simulation_results(
-    sim_output::SimulationResultsReference,
-    stage_name::String;
-    kwargs...,
+function get_variables_values!(
+    res::SimulationResults,
+    names::Vector{Symbol};
+    initial_time::Union{Nothing, Dates.DateTime} = nothing,
+    count::Union{Int, Nothing} = nothing,
```
daniel-thom, line 272: Concerns:
1. It looks like the default behavior is to load all data into memory, which will grow unbounded. That does not seem ideal.
2. Suppose the user calls this function with the first initial_time and a count of 10. Then calls it again with 11 - 20. I don't see where you purge the old values. Will the data in memory grow unbounded? The docstrings don't address whether count applies to previously-read data.

The reads from the store should be relatively fast, especially if we can keep the .h5 file handle open. Would it be better to try that out first? This is a bit of complexity that may not be required which could also cause memory issues. A user doing plots may not notice the slightly slower behavior.

src/core/simulation_results.jl
```diff
@@ -1,517 +1,378 @@
...
-        for (ix, time) in enumerate(date_df.Date)
-            file_path = date_df[ix, :File_Path]
-            var = read_arrow_file(file_path)
-            variables[(variable[l])] = vcat(variables[(variable[l])], var[1:time_length, :])
-            if l == 1
-                time_stamp = vcat(time_stamp, _read_time(file_path, time_length))
-                check_file_integrity(dirname(file_path))
-            end
-        end
+    if initial_time === nothing
```
daniel-thom, line 274: The logic in this function appears identical to the logic for duals and parameters. You could eliminate the redundancy with a helper function (and keep these explicitly-named functions).

test/test_simulation_results.jl
```diff
@@ -0,0 +1,69 @@
...
+            steps = 2,
+            stages = stages_definition,
+            stages_sequence = sequence_cache,
+            simulation_folder = file_path,
+        )
+        build_out = build!(sim)
+        @test build_out == PSI.BUILT
+        execute_out = execute!(sim)
+        @test execute_out == PSI.SUCCESSFUL_RUN
+        results_uc = SimulationResults(sim, "ED")
```
daniel-thom, line 56: Should there be some result reads here?

jd-lara: yes

src/core/simulation_results.jl
```diff
@@ -218,16 +216,17 @@ function _process_timestamps(
...
 
 function _get_variables_values(res::SimulationResults, names::Vector{Symbol}, timestamps)
-    _validate_names(res, get_existing_variables, names)
-    same_time_stamps = isempty(setdiff(res.results_timestamps, timestamps))
     existing_names = get_existing_variables(res)
-    same_names = isempty([n for n in names if n ∉ existing_names])
-    if !same_time_stamps && !same_names
-        @debug "reading variables from data store"
-        vals = _get_store_value(res, STORE_CONTAINER_VARIABLES, names, timestamps)
+    _validate_names(existing_names, names)
```
daniel-thom, line 220: You don't really need `get_existing_variables` to return a Vector. It can return an iterator. If anything, `validate_names` would be faster if you passed `existing_names` as a set.

src/core/simulation_results.jl
```diff
@@ -1,517 +1,420 @@
...
+    _store_result!(res.dual_values, _get_duals_values(res, duals, res.results_timestamps))
+    _store_result!(
+        res.parameter_values,
+        _get_variables_values(res, parameters, res.results_timestamps),
+    )
+    return nothing
+end
+
+function _clear_result_dict(dict)
+    for k in keys(dict)
```
daniel-thom, line 397:
```suggestion
    for v in values(dict)
        empty!(v)
```

src/core/simulation_results.jl
```diff
@@ -1,517 +1,420 @@
...
-# internal function to remove the overlapping results and only use the most recent
-function _read_time(file_path::String, time_length::Number)
-    time_file_path = joinpath(dirname(file_path), "time_stamp.arrow")
-    temp_time_stamp = read_arrow_file("$time_file_path")
-    time_stamp = temp_time_stamp[(1:time_length), :]
-    time_stamp = convert.(Dates.DateTime, time_stamp)
-    return time_stamp
+
+function _get_variables_values(res::SimulationResults, names::Vector{Symbol}, timestamps)
+    existing_names = get_existing_variables(res)
```
daniel-thom, line 219: For duals and parameters you check for empty names. Because the logic is duplicated in each function it looks like you forgot to do it here. But maybe it was intentional.

src/core/simulation_results.jl
```diff
@@ -1,517 +1,420 @@
...
-            file_path = step_df[ix, :File_Path]
-            var = read_arrow_file("$file_path")
-            variables[(variable[l])] = vcat(variables[(variable[l])], var[1:time_length, :])
-            if l == 1
-                time_stamp = vcat(time_stamp, _read_time(file_path, time_length))
-                check_file_integrity(dirname(file_path))
-            end
-        end
+function _get_duals_values(res::SimulationResults, names::Vector{Symbol}, timestamps)
+    isempty(names) && return res.dual_values
```
daniel-thom, line 255: Having the user pass an empty vector of names seems slightly awkward. Should there be another function that doesn't accept `names`?

src/core/simulation_results.jl
```diff
@@ -1,517 +1,420 @@
...
+    count::Int,
+    variables::Vector{Symbol} = Symbol[],
+    duals::Vector{Symbol} = Symbol[],
+    parameters::Vector{Symbol} = Symbol[],
+)
+    res.results_timestamps = _process_timestamps(res, initial_time, count)
+    _store_result!(
+        res.variable_values,
+        _get_variables_values(res, variables, res.results_timestamps),
+    )
```
daniel-thom, line 387: Would this work?
```suggestion
   merge!(
        res.variable_values,
        _get_variables_values(res, variables, res.results_timestamps),
    )
```

src/core/simulation_results.jl
jd-lara: want it to return a vector for the user to "see it" if needed. So they can check what's available when interfacing with the SimulationsResults

src/core/simulation_results.jl
jd-lara: I forgot and will add it too.

src/core/simulation_results.jl
jd-lara: this function is internal, so I can use the same code for load_simulation_results! and get_variable_results. The request could be empty.

src/core/simulation_results.jl
jd-lara: merge will iterate over all the keys in both. _store_result! just searches for the keys in the dict with the values we want to store. I can try it but it seemed better to iterate over the smallest set.

src/core/simulation_results.jl
daniel-thom: This function is internal but the user could call `get_duals_values(res, []Symbol)` and get cached results. You didn't document that, but it would work. You could formalize that interface by having another method that doesn't have the names parameter.

src/core/simulation_results.jl
daniel-thom: I don't think that it correct. I think that it will do exactly what you are doing.
```
merge!(d::AbstractDict, others::AbstractDict...)

Update collection with pairs from the other collections.
```

daniel-thom (approved):

PR #569, opened 2020-12-18
Adding ThermalMultiStart Simulation capability 

sourabhdalvi:


jd-lara (commented):
The PR looks good but probably will require some re-implementation after the many merges pending from me, Dan and Dheepak.

jd-lara:
@sourabhdalvi can you combine these changes with #583 and close this PR please? It will make it a cleaner update.

PR #570, opened 2020-12-18
Refactor store

daniel-thom:
This PR does the following:
- Removes some built-in validation of timestamps in the store. I am confident that the code is correct without them.
- Refactors the interface between SimulationResults and SimulationStore.
- Splits StageResults from SimulationResults.
- Uses `Base.empty!` instead of `clear_simulation_results!`. I think this makes more sense, especially with `length` and `isempty`. If you disagree, I'll change it back.

Something else that crossed my mind while making these changes. `SimulationStoreParams` and `SimulationStoreStageParams` are now passed through several layers of the codebase. I suggest that we consider building these into the base Simulation or SimulationSequence structs. They really have nothing to do with the store.

### DIFF
--- a/src/PowerSimulations.jl
+++ b/src/PowerSimulations.jl
@@ -152,7 +152,6 @@ export get_dual_values
 export get_stage_results
 export list_stages
 export load_simulation_results!
-export clear_simulation_results!
 
 ## Utils Exports
 export write_results
--- a/src/PowerSimulations.jl
+++ b/src/PowerSimulations.jl
@@ -12,6 +12,7 @@ export OperationsProblemTemplate
 export InitialCondition
 export SimulationSequence
 export SimulationResults
+export StageResults
 
 #Network Relevant Exports
 export StandardPTDFModel
@@ -146,6 +147,8 @@ export get_duals_values
 export get_variable_values
 export get_parameter_values
 export get_dual_values
+export get_stage_results
+export list_stages
 export load_simulation_results!
 
 ## Utils Exports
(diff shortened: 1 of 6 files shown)


daniel-thom:
@jd-lara Your recent changes caused three error prints to get logged. You can suppress error messages caused by negative tests with the `@test_logs` macro. Can you fix these?

jd-lara:
> @jd-lara Your recent changes caused three error prints to get logged. You can suppress error messages caused by negative tests with the `@test_logs` macro. Can you fix these?

Yes, I will add those in the next round of tests updates.

src/PowerSimulations.jl
```diff
@@ -148,8 +149,9 @@ export get_dual_values
 export get_variable_values
 export get_parameter_values
 export get_dual_values
+export get_stage_results
+export list_stages
 export load_simulation_results!
-export clear_simulation_results!
```
jd-lara, line 152: Why is this not exported anymore?

test/test_simulation_results.jl
```diff
@@ -139,7 +147,7 @@ function test_simulation_results(file_path::String)
             count = 25,
         )
 
-        clear_simulation_results!(results_ed)
+        empty!(results_ed)
```
jd-lara, line 150: is it preferable to use Base.empty! ?

daniel-thom: I don't think there is a correct answer. The Julia standard library uses `empty` rather than `clear` for this sort of thing, so it should probably be `empty_simulation_results`, `empty_results`, or just `empty`. We probably are not consistent on this point.

We are storing containers of results, and so `empty!(results)` felt natural to me. I will go with whatever you prefer.

jd-lara: Base.empty! is ok to me. I didn't realize this was the change when I saw the other change.

jd-lara (approved):

PR #571, opened 2020-12-18
Fix build of docs

daniel-thom:


src/core/cache_utils.jl
```diff
@@ -2,10 +2,7 @@ const ParamCacheKey = NamedTuple{(:stage, :type, :name), NTuple{3, Symbol}}
 
 make_cache_key(stage, type, name) = (stage = stage, type = type, name = name)
 
-"""
-Priority for keeping data in cache to serve reads.
-Currently unused.
-"""
+# Priority for keeping data in cache to serve reads. Currently unused.
```
daniel-thom, line 5: I don't know why a real docstring fails. Not willing to invest more time at the moment.

jd-lara (approved):
I will approve but we should later tests if this will work when adding the module to the docs.

src/core/result_cache.jl
```diff
@@ -47,11 +47,11 @@ end
 """
 Return true if the data for `timestamp` is stored in cache.
 """
-function is_cached!(cache::ResultCache, type, name, stage, timestamp)
-    return is_cached!(cache, make_cache_key(stage, type, name), timestamp)
+function is_cached(cache::ResultCache, type, name, stage, timestamp)
```
daniel-thom, line 50: A docstring was referencing `is_cached`, which also caused building of docs to fail. The name should not have a !

PR #575, opened 2020-12-21
fixing dimensions of hydro_budget param

claytonpbarrows:


jd-lara (approved):
Please comment about why this change was needed so I can consider it in the stage update code, please. The current code is consistent with the single time step parameter and you moved this to a multi-step parameter.

claytonpbarrows:
the issue is:
 1. in the existing hydro budget formulation, the parameter array has length == 1 since the intention is to sum the total energy and not exceed the value in the parameter array
 2. but the time series has length == 48 because we maintain consistent time series lengths 
 3. in update_parameter, we get the time series values and then iterate through the parameter array and set the values. In our case the parameter array (length=1) is only representing the first period in the time series. Instead, we need to pass the full time series and sum it to properly update the parameter.

that might not be totally clear, so @jd-lara let me know..

PR #576, opened 2020-12-23
Fix service model limit functions

kdheepak:


jd-lara (changes requested):
can you please add the discussed tests?

jd-lara (approved):

PR #577, opened 2020-12-23
Cb/pg results

claytonpbarrows:
- [x] align windowing with results periods 
 - [x] print summary stats of results
 - [x] make kwargs args in load_results 
 - [x] load_simulation_results -> load_results 
 - [ ] remove results_timestamps # what about the store reads?
 - [ ] add timestamps to dfs
 - [x] get_*_values -> read_* ... in op_problem_results too 
 - [x] pre-allocate array or df in realized_results #copycols = false


daniel-thom (approved):
Looks good

daniel-thom:
@jd-lara I will merge now so that I can integrate this with my export changes. You can still review the final result.

PR #578, opened 2020-12-24
Export results to CSV

daniel-thom:


src/core/result_cache.jl
```diff
@@ -28,7 +28,7 @@ set_min_flush_size!(cache::ResultCache, x) = cache.min_flush_size = x
 
 function add_param_cache!(cache::ResultCache, key, flush_rule)
     cache.data[key] = ParamResultCache(key, flush_rule)
-    @debug "Added cache container for" key flush_rule
+    #@debug "Added cache container for" key flush_rule
```
daniel-thom, line 31: This is overly noisy but we might want it enabled in the future.

src/core/simulation_results.jl
```diff
@@ -118,26 +116,45 @@ IS.get_parameters(result::StageResults) = result.parameter_values
...
-            end
-            results[name] = _results
+    resolution = PSY.get_time_series_resolution(get_system(res))
+    horizon = PSY.get_forecast_horizon(get_system(res))
+    for name in names
+        _results = SortedDict{Dates.DateTime, DataFrames.DataFrame}()
+        for ts in timestamps
+            out = read_result(DataFrames.DataFrame, store, stage_name, field, name, ts)
+            time_col = range(ts, length = horizon, step = resolution)
+            DataFrames.insertcols!(out, 1, "DateTime" => time_col)
```
daniel-thom, line 152: This means that the timestamps will be stored in-memory if the user calls `load_results!`.

src/core/simulation_results.jl
```diff
@@ -118,26 +116,45 @@ IS.get_parameters(result::StageResults) = result.parameter_values
...
+        _get_store_value(res, field, names, timestamps, store)
+    end
+end
+
 function _get_store_value(
     res::StageResults,
     field::Symbol,
     names::Vector{Symbol},
     timestamps,
+    store::SimulationStore,
```
daniel-thom, line 140: @jd-lara You should note this change. I changed the functions in the readX call stacks to accept an optional store. This will allow us to avoid opening/closing the store repeatedly but still guarantee that the file handle gets closed deterministicly.

src/core/result_cache.jl
claytonpbarrows, line 31: is this intentionally commented and not deleted?

claytonpbarrows (approved):
LGTM, I think you already addressed my only comment.

src/core/simulation_results.jl
jd-lara: noted

PR #581, opened 2021-01-07
Adding new Battery formulations

sourabhdalvi:
New Formulations
- [x] BookKeeping with end of period energy target (hard) constraint 
- [x] BookKeeping with end of period energy target (soft) constraint with slack variable
- [x]  BookKeeping with end of period value of energy

Requires PSY#685 to be merged

### DIFF
--- a/src/devices_models/devices/storage.jl
+++ b/src/devices_models/devices/storage.jl
@@ -4,7 +4,9 @@ abstract type AbstractStorageFormulation <: AbstractDeviceFormulation end
 struct BookKeeping <: AbstractStorageFormulation end
 struct BookKeepingwReservation <: AbstractStorageFormulation end
 struct BookKeepingwTarget <: AbstractStorageFormulation end
+struct BookKeepingwTargetTimeSeries <: AbstractStorageFormulation end
 struct BookKeepingwSoftTarget <: AbstractStorageFormulation end
+struct BookKeepingwSoftTargetTimeSeries <: AbstractStorageFormulation end
 struct BookKeepingwEnergyValue <: AbstractStorageFormulation end
 ########################### ActivePowerInVariable, Storage #################################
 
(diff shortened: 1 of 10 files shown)


src/devices_models/devices/storage.jl
```diff
@@ -3,7 +3,11 @@
 abstract type AbstractStorageFormulation <: AbstractDeviceFormulation end
 struct BookKeeping <: AbstractStorageFormulation end
 struct BookKeepingwReservation <: AbstractStorageFormulation end
-
+struct BookKeepingwTarget <: AbstractStorageFormulation end
+struct BookKeepingwTargetTimeSeries <: AbstractStorageFormulation end
+struct BookKeepingwSoftTarget <: AbstractStorageFormulation end
+struct BookKeepingwSoftTargetTimeSeries <: AbstractStorageFormulation end
+struct BookKeepingwEnergyValue <: AbstractStorageFormulation end
 ########################### ActivePowerInVariable, Storage #################################
```
claytonpbarrows, line 11: abstract type BookKeeping <: AbstractStorageFormulation
abstract type StorageTarget <: BookKeeping
struct StorageTargetTimeSeries <: StorageTarget

src/devices_models/devices/storage.jl
```diff
@@ -335,64 +347,66 @@ end
...
+                UpdateRef{St}(TARGET, target_forecast_label),
+            )
+        else
+            energy_soft_target_timeseries(
+                psi_container,
+                constraint_infos_target,
+                make_constraint_name(ENERGY_TARGET, St),
+                (
+                    make_variable_name(ENERGY, St),
+                    make_variable_name(ENERGY_TARGET_SLACK, St),
```
claytonpbarrows, line 389: I had previously recognized a pattern that all `*._SLACK` variables were subject to a `*._SLACK_COST` defined in https://github.com/NREL-SIIP/PowerSimulations.jl/blob/sd/battery_model/src/core/definitions.jl. This breaks that convention and makes this slack subject to the cost defined in the struct. 
I'm not necessarily saying that we should keep the old convention and change this terminology, but rather highlighting the change and asking if it matters to anyone (@jd-lara)?

sourabhdalvi: `ENERGY_TARGET_SHORTAGE`/`ENERGY_SHORTAGE`?

src/devices_models/devices/storage.jl
```diff
@@ -1,9 +1,16 @@
 #! format: off
 
 abstract type AbstractStorageFormulation <: AbstractDeviceFormulation end
+abstract type EnergyManagementFormulation <: AbstractStorageFormulation end
+abstract type EnergyTargetFormulation <: EnergyManagementFormulation end
+abstract type RelaxedEnergyTargetFormulation <: EnergyManagementFormulation end
 struct BookKeeping <: AbstractStorageFormulation end
 struct BookKeepingwReservation <: AbstractStorageFormulation end
-
+struct EndOfPeriodStaticEnergyTarget <: EnergyTargetFormulation end
```
jd-lara, line 9: We need docstrings for these formulations. It isn't at all clear what the model is from the description.

jd-lara (commented):
I am not totally convinced that the formulation abstractions is adequate since the formulations seem to be basically the same except for the parameters.

PR #585, opened 2021-01-16
add progress meter

jd-lara:


jd-lara:
Adds a progress meter instead of the old print lines as follows: 

Screen Shot 2021-01-15 at 6 41 12 PM
Screen Shot 2021-01-15 at 6 40 44 PM
Screen Shot 2021-01-15 at 6 40 31 PM

src/core/psi_container.jl
```diff
@@ -92,6 +92,13 @@ function _make_jump_model!(psi_container::PSIContainer)
         parameters && PJ.enable_parameters(JuMPmodel)
         psi_container.JuMPmodel = JuMPmodel
     end
+    if !get_optimizer_log_print(settings)
+        JuMP.set_silent(psi_container.JuMPmodel)
+    elseif get_optimizer_log_print(settings)
+        JuMP.unset_silent(psi_container.JuMPmodel)
```
daniel-thom, line 98: This could be simpler.
```
    if get_optimizer_log_print(settings)
        JuMP.unset_silent(psi_container.JuMPmodel)
    else
        JuMP.set_silent(psi_container.JuMPmodel)
    end
```

daniel-thom (approved):
Looks nice! Cool addition.

Just one minor simplification that you can consider.

claytonpbarrows (approved):
This is cool.

src/core/psi_container.jl
jd-lara: I wanted to do this just in case we change the setting later and be able to catch it. If that's ok with you I would ratter leave or change it by and @assert maybe?

daniel-thom: I don't I think I'm following you. The function returns either true or false. It's not really important...whatever you think is best.

PR #588, opened 2021-01-21
A Consolidated Battery Formulation for End effects

sourabhdalvi:


src/core/variables.jl
```diff
@@ -10,6 +10,7 @@ const ENERGY_DOWN = "Edown"
 const ENERGY_BUDGET = "energy_budget"
 const ENERGY_BUDGET_UP = "energy_budget_up"
 const ENERGY_BUDGET_DOWN = "energy_budget_down"
+const ENERGY_TARGET_SLACK = "Eslack"
```
jd-lara, line 13:
```suggestion
const ENERGY_TARGET_SLACK = "energy_slack"
```

src/devices_models/devices/storage.jl
```diff
@@ -3,6 +3,7 @@
 abstract type AbstractStorageFormulation <: AbstractDeviceFormulation end
 struct BookKeeping <: AbstractStorageFormulation end
 struct BookKeepingwReservation <: AbstractStorageFormulation end
+struct EndOfPeriodEnergyTarget <: AbstractStorageFormulation end
```
jd-lara, line 6: We should make some AbstractEnergyManagement that can enable other formulations with energy related constraints.

jd-lara (changes requested):
The PR looks way better but needs some additional improvements

jd-lara (approved):

PR #589, opened 2021-01-21
Simulation store

jd-lara:


daniel-thom:
A unit test produces this error message. It should get suppressed with `@test_logs`.
```
Error: Requested time does not match available results
└ @ PowerSimulations ~/sandboxes/PowerSimulations.jl/src/core/simulation_results.jl:409
```

src/core/simulation.jl
```diff
@@ -884,97 +923,142 @@ function update_stage!(
...
+
+    progress_bar = ProgressMeter.Progress(num_executions, 1)
+    for step in 1:steps
+        TimerOutputs.@timeit RUN_SIMULATION_TIMER "Execution Step $(step)" begin
+            IS.@record :simulation_status SimulationStepEvent(
+                get_current_time(sim),
+                step,
+                "start",
+            )
+            for (ix, stage_number) in enumerate(execution_order)
```
claytonpbarrows, line 1012: Everything in this for loop should go in a separate function that takes the stage. We'll need this to enable the SIIP-MODSIM and Helics style extensions that won't be setting up a sim-store object.

src/core/simulation.jl
```diff
@@ -983,21 +1067,145 @@ function _execute!(sim::Simulation; kwargs...)
...
+                stage_sym,
+                STORE_CONTAINER_DUALS,
+                name,
+                false,
+                CachePriority.LOW,
+            )
         end
-        sim_results = SimulationResultsReference(sim)
+
+        for (name, param_container) in parameters
```
claytonpbarrows, line 1138: `parameters` here can be `nothing` which creates an error. Can we either:
1. add logic to skip this for loop if parameters is nothing
2. modify `get_parameters` to return an empty dict

src/PowerSimulations.jl
```diff
@@ -103,6 +105,7 @@ export GenericOpProblem
 export UnitCommitmentProblem
 export EconomicDispatchProblem
 # export OptimalPowerFlow
+export HdfSimulationStore
```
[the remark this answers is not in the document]
jd-lara: I don't think we should export anything related to the data store, this is a feature for type 3 users.

src/core/simulation.jl
jd-lara: I addressed this one in 26018df

claytonpbarrows (approved):
LGTM

daniel-thom:
> A unit test produces this error message. It should get suppressed with `@test_logs`.
> 
> ```
> Error: Requested time does not match available results
> └ @ PowerSimulations ~/sandboxes/PowerSimulations.jl/src/core/simulation_results.jl:409
> ```

Is this error message fixed?

jd-lara:
> > A unit test produces this error message. It should get suppressed with `@test_logs`.
> > ```
> > Error: Requested time does not match available results
> > └ @ PowerSimulations ~/sandboxes/PowerSimulations.jl/src/core/simulation_results.jl:409
> > ```
> 
> Is this error message fixed?

Yes, 3b8c587

daniel-thom (approved):

PR #591, opened 2021-01-23
Cb/sim results wo system

claytonpbarrows:
enabling SimulationResults without adding a System. This is critical for cases where you run a simulation with `system_to_file = false` because you might have included components not supported for serialization.

### DIFF
--- a/src/core/simulation_results.jl
+++ b/src/core/simulation_results.jl
@@ -638,7 +638,7 @@ end
 """
 Construct SimulationResults from a simulation.
 """
-SimulationResults(sim::Simulation) = SimulationResults(get_simulation_dir(sim))
+SimulationResults(sim::Simulation; kwargs...) = SimulationResults(get_simulation_dir(sim); kwargs...)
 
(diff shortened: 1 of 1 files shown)


src/core/simulation_results.jl
```diff
@@ -638,7 +638,7 @@ end
 """
 Construct SimulationResults from a simulation.
 """
-SimulationResults(sim::Simulation) = SimulationResults(get_simulation_dir(sim))
+SimulationResults(sim::Simulation; kwargs...) = SimulationResults(get_simulation_dir(sim); kwargs...)
```
jd-lara, line 641: I prefer to have the kwargs explicitly to that the function fails if an invalid option is passed

jd-lara (changes requested):
We might want to test this to make sure we caught all cases of systems being passed and or not.

src/core/simulation_results.jl
daniel-thom: IMHO, when you're simply forwarding kwargs to another function, what Clayton did is better. He is explicitly listed the arguments in the function that consumes them.

jd-lara (approved):

PR #595, opened 2021-01-27
add psi container closure

jd-lara:
This PR add a function to make checks for the psi_container. More checks could be added afterwards.


daniel-thom (approved):
Nice addition

PR #598, opened 2021-01-27
Fixing multiplier array NaN issue

sourabhdalvi:


src/core/feedforward.jl
```diff
@@ -288,19 +288,23 @@ function ub_ff(
...
     container = add_param_container!(psi_container, param_reference, set_name)
     param_ub = get_parameter_array(container)
+    multiplier_ub = get_multiplier_array(container)
     con_ub = add_cons_container!(psi_container, ub_name, set_name, time_steps)
 
     for constraint_info in constraint_infos
         name = get_component_name(constraint_info)
         value = JuMP.upper_bound(variable[name, 1])
         param_ub[name] = PJ.add_parameter(psi_container.JuMPmodel, value)
+        multiplier_ub[name] = 1.0
```
jd-lara, line 298: why is this default to 1.0?

src/core/feedforward.jl
```diff
@@ -365,6 +370,8 @@ function range_ff(
             PJ.add_parameter(psi_container.JuMPmodel, JuMP.lower_bound(variable[name, 1]))
         param_ub[name] =
             PJ.add_parameter(psi_container.JuMPmodel, JuMP.upper_bound(variable[name, 1]))
+        multiplier_ub[name] = 1.0
```
jd-lara, line 373: same here

src/core/feedforward.jl
sourabhdalvi: because this implementation doesn't use the multiplier,  so ub_value * multiplier_ub is still ub_value

jd-lara: we can leave this for now but please add a comment so that when we refactor we know why is one.

PR #599, opened 2021-01-28
Jd/refactor sim sequence

jd-lara:
Pending 

- [ ] Address inconsistency between Operations Results
- [ ] Figure out serialization of OperationsProblem
- [x] Use temp path in tests for operation model
- [x] Add tests for templates


jd-lara:
Note this issue with mktempdir() in windows https://github.com/JuliaLang/julia/issues/39457. This is why I removed the cleanup flag

src/core/device_models.jl
```diff
@@ -0,0 +1,78 @@
...
+
+# Accepted Key Words
+- `feedforward::Array{<:AbstractAffectFeedForward}` : use to pass parameters between models
+
+# Example
+```julia
+thermal_gens = DeviceModel(ThermalStandard, ThermalBasicUnitCommitment),
+```
+"""
+mutable struct DeviceModel{D <: PSY.Device, B <: AbstractDeviceFormulation}
```
claytonpbarrows, line 44: The more I use large systems, the more I think we need some additional control on these definitions. We should consider adding a `filter` field that can be applied in `get_components(component_type, sys, filter)`. Of course, we'd also need a check to issue warnings if you inadvertently fail to cover all components.

jd-lara: There is now a possibility to do similar things for the device model. Implementing that filter in this new code won't be too hard but it will effectively NOT model those devices.

PR #600, opened 2021-01-28
Refactoring Storage constraint for energy balance and energy target

sourabhdalvi:


jd-lara (changes requested):
I am not sure why we have different functions for energy target hydro and storage. I see a lot of improvement to simplify code here and make functions more generic between the models. 

In principle the common mathematics of both is the same.

sourabhdalvi:
> I am not sure why we have different functions for energy target hydro and storage. I see a lot of improvement to simplify code here and make functions more generic between the models.
> 
> In principle the common mathematics of both is the same.

jd-lara:
> > I am not sure why we have different functions for energy target hydro and storage. I see a lot of improvement to simplify code here and make functions more generic between the models.
> > In principle the common mathematics of both is the same.
> 
> I think the make difference between the two is for Storage we are using a static target vs in Hydro the target is a time series. I think we need to add support for both methods for both devices going forward, let talk about this .
> 
> Sorry closed the PR by mistake

The input data shouldn't matter here. I think that there is space to match both, the same model could be used for batteries, hydrogen, hydro. Anything with a book keeping of energy + a storage target should can be modeled this way.

sourabhdalvi:
> > > I am not sure why we have different functions for energy target hydro and storage. I see a lot of improvement to simplify code here and make functions more generic between the models.
> > > In principle the common mathematics of both is the same.
> > 
> > 
> > I think the make difference between the two is for Storage we are using a static target vs in Hydro the target is a time series. I think we need to add support for both methods for both devices going forward, let talk about this .
> > Sorry closed the PR by mistake
> 
> The input data shouldn't matter here. I think that there is space to match both, the same model could be used for batteries, hydrogen, hydro. Anything with a book keeping of energy + a storage target should can be modeled this way.
I agree, I was just pointing the difference.

claytonpbarrows (commented):
I believe that [this function](https://github.com/NREL-SIIP/PowerSimulations.jl/pull/600/files#diff-b6353133a2ca8995aa9e66adf95b718ee1decffc1b05ca83340d4dfd8b229770L777-L789) can go away.
As it stands, it removes all `HydroDispatch` variable costs and then throws warnings about not having variable costs. I think it's leftover from when we didn't have a op_cost field in HydroDispatch

src/core/variables.jl
```diff
@@ -52,6 +54,10 @@ struct EnergyVariableUp <: VariableType end
 
 struct EnergyVariableDown <: VariableType end
 
+struct EnergySlackUp <: VariableType end
```
jd-lara, line 57: The way this is implemented, the energy targets are not slacks they are more like a surplus or a deficit. We should use better terminology to not imply these are equivalent to the slacks added via kwargs

src/devices_models/devices/common/cost_functions.jl
```diff
@@ -10,6 +10,7 @@ struct AddCostSpec
     shut_down_cost::Union{Nothing, Function}
     fixed_cost::Union{Nothing, Function}
     has_multistart_variables::Bool
+    has_target_cost::Bool
```
jd-lara, line 13: Why do we need a bool here is there is also the field additional linear terms? we should just add the terms if they exist.

src/devices_models/devices/common/cost_functions.jl
```diff
@@ -701,6 +749,86 @@ function add_to_cost!(
     return
 end
 
+function add_to_cost!(
```
jd-lara, line 752: I don't get what you are trying to do here. You pass the spec and the cost_data simultaneously which in a way have the same information.

src/devices_models/devices/common/energy_balance_constraint.jl
```diff
@@ -235,37 +196,21 @@ If t > 1:
 function energy_balance_hydro!(
```
jd-lara, line 196: Why is this still a separate function?

src/devices_models/devices/common_models.jl
```diff
@@ -0,0 +1,104 @@
...
+
+get_variable_binary(::EnergySlackUp, ::Type{<:PSY.Component}) = false
+get_variable_lower_bound(::EnergySlackUp, d::PSY.Component, _) = 0.0
+
+############## EnergySlackDown, Storage ####################
+
+get_variable_binary(::EnergySlackDown, ::Type{<:PSY.Component}) = false
+get_variable_upper_bound(::EnergySlackDown, d::PSY.Component, _) = 0.0
+
+get_target_multiplier(v::PSY.HydroEnergyReservoir) = PSY.get_storage_capacity(v)
```
jd-lara, line 11: Why are these here, they should be in the respective Storage and Hydro files.

src/devices_models/devices/storage.jl
```diff
@@ -47,7 +47,7 @@ function DeviceRangeConstraintSpec(
     ::Type{<:RangeConstraint},
     ::Type{ActivePowerOutVariable},
     ::Type{T},
-    ::Type{<:AbstractStorageFormulation},
+    ::Type{<:BookKeeping},
```
jd-lara, line 50: We need to rename BookKeeping since in principle both do bookkeeping one just happens to have reservation variables.

jd-lara (changes requested):
We still need to work through some issues in this PR to get it right.

PR #601, opened 2021-01-29
rename structs

jd-lara:
This PR introduces no logic changes. As recommended by @daniel-thom the renaming portions of #599 have been moved to this PR. The objective is to remove acronyms and use easier words for the users to understand what things are. 

**Note** some of these changes make sense in the context of #599 where we are merging stage with OperationsProblem into a single struct to be used in single problem solutions and in simulation.


daniel-thom (approved):
Thank you so much for doing it this way.

claytonpbarrows (approved):

kdheepak (approved):
Looks great!

PR #603, opened 2021-02-03
[WIP]: Hybrid Model Development

sourabhdalvi:
Currently I have few hacks implemented to make the hybrid model work. 
- I have had to use new variable names for each component in the hybrid system  https://github.com/NREL-SIIP/PowerSimulations.jl/blob/d4212f6f898da7f2886fbde8ed7c8036ea3f076d/src/core/variables.jl#L96-L112 
- This limits the formulation for the hybrid device to the simplest version ( ThermalDispatch, RenewableDispatch, BookKeeping, etc..)
- Since all of the HybridSystems done have each component I had use `devices_filter_func` field in DeviceRangeConstraintSpec which resulted in changes to `_apply_range_constraint_spec!` and `_apply_timeseries_range_constraint_spec!` to accept `Union{Vector{T}, IS.FlattenIteratorWrapper{T}}} where {T <: PSY.Device}`  

The following model are currently working with few limitations
 - PhysicalCoupling (only applies input/output active power limits or inverter rating constraints)
 - FinancialCoupling (allows for MarketBidCost to be associated with output power variable)
 - StandardHybridFormulation (PhysicalCoupling + FinancialCoupling)


src/core/constraints.jl
```diff
@@ -52,6 +52,9 @@ const MUST_RUN = "must_run"
 const MUST_RUN_LB = "must_run_lb"
 const NODAL_BALANCE_ACTIVE = "nodal_balance_active"
 const NODAL_BALANCE_REACTIVE = "nodal_balance_reactive"
+const POWER_BALANCE_INFLOW = "power_balance_inflow"
+const POWER_BALANCE_OUTFLOW = "power_balance_outflow"
+const INVERTOR_LIMIT = "invertor_limit"
```
jd-lara, line 57: Inverter, but better would be to have an export limit or pcc limit to be encompasing of an inverter or consider also a transformer.

src/core/variables.jl
```diff
@@ -84,6 +93,24 @@ struct AdditionalDeltaActivePowerDownVariable <: VariableType end
 
 struct SmoothACE <: VariableType end
 
+struct ActivePowerVariableThermal <: VariableType end
```
jd-lara, line 96: We shouldn't need to do this. There has to be a better way to integrate with the existing infrastructure that won't require exploding in the number of variables we support

src/devices_models/device_constructors/hybrid_constructor.jl
```diff
@@ -0,0 +1,485 @@
...
+    S <: PM.AbstractActivePowerModel,
+}
+    devices = get_available_components(T, sys)
+
+    if !validate_available_devices(T, devices)
+        return
+    end
+
+    # Variables
+    add_variables!(psi_container, ActivePowerInVariable, devices)
```
jd-lara, line 18: here you can implement a method add_variables that internally makes the variables for the hybrid systems components.

jd-lara (commented):
We need to re-think the implementation of this PR. This approach won't scale adequately moving forward for many hybrid systems.

model_op.txt
```diff
@@ -0,0 +1,75 @@
+Min 4500 Pout__HybridSystem_{all_hybrid, 1} + 4500 Pout__HybridSystem_{RE+battery, 1} + 4500 Pout__HybridSystem_{load+battery, 1} + 4500 Pout__HybridSystem_{thermal+battery, 1}
```
jd-lara, line 1: Don't add this file.

jd-lara:
closed in favor of #832

PR #605, opened 2021-02-05
Handle case where array size is greater than min flush size

daniel-thom:


src/core/simulation.jl
```diff
@@ -1011,6 +1011,7 @@ function _execute!(
     sim::Simulation,
     store;
     cache_size_mib = 1024,
+    min_cache_flush_size_mib = MIN_CACHE_FLUSH_SIZE_MiB,
```
daniel-thom, line 1014: We can now customize the flush size by passing this kwarg to `execute!`.

src/core/simulation_store.jl
```diff
@@ -34,6 +34,24 @@ struct SimulationStoreStageParams
...
+        horizon,
+        interval,
+        resolution,
+        base_power,
+        system_uuid,
+    )
+        new(
+            num_executions,
+            horizon,
+            Dates.Millisecond(interval),
```
daniel-thom, line 49: In some tests I was getting Dates.Hour here, so I added code to force them to Millisecond.

src/core/simulation.jl
jd-lara: Should we also increase the size of the default value or do something smart to set that up?

daniel-thom: The code will flush this value or the size of each array-at-a-time-point, whichever is bigger. I think this is OK.

jd-lara: My question is how does a user know they need to increase the flush size?

jd-lara (commented):
It seems we ought to add some checks to make sure that the data read from the store isn't empty. Maybe in the return of `_flush_data!()` in line 456 of `hdf_simulation_store`

src/core/simulation.jl
daniel-thom: My hope is that users never need to know about it. There is no need to tempt users to set it. If there ever is a problem then this will allow us to customize behavior without having to change any code.

daniel-thom:
> It seems we ought to add some checks to make sure that the data read from the store isn't empty. Maybe in the return of `_flush_data!()` in line 456 of `hdf_simulation_store`

That was my intention with this addition. Let me know if your concern is not met.
https://github.com/daniel-thom/PowerSimulations.jl/blob/fix-large-array-size/src/core/param_result_cache.jl#L79

jd-lara:
> > It seems we ought to add some checks to make sure that the data read from the store isn't empty. Maybe in the return of `_flush_data!()` in line 456 of `hdf_simulation_store`
> 
> That was my intention with this addition. Let me know if your concern is not met.
> https://github.com/daniel-thom/PowerSimulations.jl/blob/fix-large-array-size/src/core/param_result_cache.jl#L79

Sorry, I didn't realized that this was the intention of that line. This works.

claytonpbarrows (approved):

PR #606, opened 2021-02-08
Jd/refactor sim sequence

jd-lara:


src/core/service_models.jl
```diff
@@ -0,0 +1,104 @@
...
+
+function _set_model!(
+    dict::Dict,
+    service_name::String,
+    model::ServiceModel{D, B},
+) where {D <: PSY.Service, B <: AbstractServiceFormulation}
+    if !model.use_service_name
+        throw(
+            IS.ConflictingInputsError(
+                "The model provided has use_service_name false. This addition method can't be used",
```
daniel-thom, line 83:
```suggestion
                "The model provided has use_service_name false. This additional method can't be used",
```

Falling into this condition would be a coding error, not a user input error, so an assert may be more appropriate.

src/operations_problem_templates.jl
```diff
@@ -0,0 +1,135 @@
...
+```julia
+template = template_economic_dispatch()
+```
+
+# Accepted Key Words
+- `network::Type{<:PM.AbstractPowerModel}` : override default network model settings
+- `devices::Vector{DeviceModel}` : override default `DeviceModel` settings
+- `services::Vector{ServiceModel}` : override default `ServiceModel` settings
+"""
+function template_economic_dispatch(; kwargs...)
```
daniel-thom, line 79: It might be more appropriate to explicitly list the kwargs as optional parameters. As it is a user could pass misspelled arguments and you will not use them or throw errors. Do you want this function to ignore extra arguments? It's fine to use `kwargs...` to generically forward parameters. However, I think it's generally better to have some function in the stack (probably at the bottom) explicitly list them so that you catch errors.

Same comment applies to many other locations.

src/operations_problem_templates.jl
```diff
@@ -0,0 +1,135 @@
...
+    template_agc_reserve_deployment(; kwargs...)
+
+Creates an `OperationsProblemTemplate` with default DeviceModels for an AGC Reserve Deplyment Problem. This model doesn't support customization
+
+# Example
+```julia
+template = agc_reserve_deployment()
+```
+"""
+function template_agc_reserve_deployment(; kwargs...)
```
daniel-thom, line 103:
```suggestion
function template_agc_reserve_deployment()
```

src/core/operations_problem.jl
```diff
@@ -48,520 +124,632 @@ OpModel = OperationsProblem(TestOpProblem, template, system)
...
+    end
+    return get_status(problem)
 end
 
 """
-    OperationsProblem(filename::AbstractString)
+Default implementation of build method for Operational Problems for models conforming with PowerSimulationsOperationsProblem specification. Overload this function to implement a custom build method
+"""
+function problem_build!(problem::OperationsProblem{<:PowerSimulationsOperationsProblem})
+    _build_imp!(
```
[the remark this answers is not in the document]
jd-lara: I felt this makes it easier to read the code base when looking for similar patterns. So readers of code won't copy paste the contents of build_imp to start with.

src/core/service_models.jl
jd-lara: I think the message didn't reflect the actual desired message which was that the method can't be used. I changed it.

src/operations_problem_templates.jl
jd-lara: agreed, I haven't totally cleaned the code in this file.

src/core/optimization_container.jl
```diff
@@ -638,31 +630,37 @@ function check_optimization_container(optimization_container::OptimizationContai
...
+    model = optimization_container.JuMPmodel
+    vars = JuMP.num_variables(model)
+    cons = 0
+    for (exp, c_type) in JuMP.list_of_constraint_types(model)
+        cons += JuMP.num_constraints(model, exp, c_type)
+    end
+    return "The current total number of variables is $(vars) and total number of constraints is $(cons)"
+end
+
+function build_impl(
```
daniel-thom, line 643: Did you inadvertently remove the `!` from this function name?

jd-lara: yes, this was a silly mistake.

jd-lara: I fixed this in the forthcoming PR. I will merge to open a new PR with changes to simulation

PR #608, opened 2021-02-10
refactor sim sequence the sequel

jd-lara:
This PR makes the changes following #606 to implement joining Stage and OperationsProblem. I am sorry for the name changes galore in the PR @daniel-thom.


src/core/operations_problem.jl
```diff
@@ -329,30 +310,30 @@ function build!(
...
-            problem_build!(problem)
-            #serialize_problem(problem, "operations_problem")
-            #serialize_optimization_model(problem)
-            set_status!(problem, BuildStatus.BUILT)
-        end
-    catch e
-        @error "Operation Problem Build Fail" exception = e
-        set_status!(problem, BuildStatus.FAILED)
-    end
+    #try
```
daniel-thom, line 317: What's going on here?

jd-lara: forgot to commit something

src/core/simulation.jl
```diff
@@ -736,32 +722,32 @@ function initial_condition_update!(
             ic,
             quantity,
             previous_value,
-            get_number(stage),
+            get_simulation_number(problem),
```
daniel-thom, line 725: I'm not sure that this is intuitive.

src/core/simulation.jl
```diff
@@ -223,22 +218,34 @@ end
 """
 mutable struct Simulation
     steps::Int
-    stages::Dict{String, OperationsProblem{<:AbstractOperationsProblem}}
+    problems::SimulationProblems
```
daniel-thom, line 221:
```
Simulation
    SimulationSequence
    SimulationProblems
        Dict{OperationsProblem}
            ProblemInternal
                SimulationInfo
```
`SimulationInfo` doesn't seem quite right to me.

src/core/simulation.jl
```diff
@@ -250,77 +257,63 @@ they passed to the original Simulation.
 
 # Arguments
 - `directory::AbstractString`: the directory returned from the call to serialize
-- `stage_info::Dict`: Two-level dictionary containing stage parameters that cannot be
-  serialized. The outer dict should be keyed by the stage name. The inner dict must contain
+- `problem_info::Dict`: Two-level dictionary containing problem parameters that cannot be
```
daniel-thom, line 260: Should this be a struct?

src/core/simulation.jl
```diff
@@ -615,68 +601,68 @@ function build!(
...
+        problem_interval = get_interval(sequence, problem_name)
+        # Note to devs: Here we are setting the number of operations problem executions we
+        # will see for every step of the simulation
+        if ix == 1
+            set_executions!(problem, 1)
+        else
+            step_resolution = get_step_resolution(sequence)
+            get_interval(sequence, problem_name)
+            set_executions!(problem, Int(step_resolution / problem_interval))
+            Int(step_resolution / problem_interval)
```
daniel-thom, line 617: Unused expression?

src/core/simulation.jl
```diff
@@ -615,68 +601,68 @@ function build!(
...
     end
     _assign_feedforward_chronologies(sim)
-    _check_steps(sim, stage_initial_times)
-    _build_stages!(sim)
+    _check_steps(sim, problem_initial_times)
+    _build_problems!(sim)
     if serialize
         TimerOutputs.@timeit BUILD_SIMULATION_TIMER "Serializing Simulation Files" begin
-            serialize_simulation(sim)
+            # serialize_simulation(sim)
```
daniel-thom, line 626: Perhaps mark this somehow so that it doesn't get lost?

src/core/simulation_problems.jl
```diff
@@ -0,0 +1,67 @@
+"""
+Stores the OperationProblem definitions to be used in the simulation. When creating the
+SimulationProblems, the order in which the problems are created determines the order on which
+the simulation is executed.
```
daniel-thom, line 4: Does this relate to SimulationSequence?

src/core/simulation_problems.jl
```diff
@@ -0,0 +1,67 @@
...
+Base.length(problems::SimulationProblems) = length(problems.op_problems)
+Base.first(problems::SimulationProblems) = first(problems.op_problems)
+Base.iterate(problems::SimulationProblems, args...) = iterate(problems.op_problems, args...)
+
+get_problem_names(problems::SimulationProblems) = problems.names
+get_problem_instances(problems::SimulationProblems) = values(problems.op_problems)
+
+function get_problem_numer(problems::SimulationProblems, name)
+    return findfirst(x -> x == Symbol(name), get_problem_names(problems))
+end
```
daniel-thom, line 27: This doesn't seem correct. I don't see anything about a number.

src/core/simulation_store.jl
```diff
@@ -10,32 +10,32 @@ Provides storage of simulation data
 abstract type SimulationStore end
 
 # Required methods:
-# - initialize_stage_storage!
+# - initialize_problem_storage!
 # - read_array
 
-struct SimulationStoreStageRequirements
+struct SimulationStoreProblemRequirements
```
daniel-thom, line 16: As part of this refactor it would be a good idea to unify the structs in this file with the main simulation structs. There is redundant information.

src/core/simulation.jl
jd-lara: Yeah, this needs to be reimplemented

src/core/simulation.jl
jd-lara: In what sense?

src/core/simulation_store.jl
jd-lara: Agreed. But we should probably do it in a follow up PR. This is getting big already and there are a lot of test issues that need to be addressed as well.

src/core/simulation.jl
jd-lara: Do you suggest reverting to get_number? The idea was to indicate this method is used in a simulation.

src/core/simulation.jl
jd-lara: This is for de-serialization which is currently broken.

src/core/simulation_problems.jl
jd-lara: find first returns the location of the function. This gets me the location in the array of names (which is ordered) where the name is.

docs/src/man/simulation_recorder.md
```diff
@@ -3,7 +3,7 @@
 PowerSimulations provides the ability to record structured data as events
 during a simulation. These events can be post-processed to help debug problems.
 
-By default only SimulationStepEvent and SimulationStageEvent are recorded.  Here is an example.
+By default only SimulationStepEvent and SimulationProblemEvent are recorded.  Here is an example.
```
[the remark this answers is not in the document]
jd-lara: SimulationProblemEvent -> ProblemExecutionEvent

src/core/simulation.jl
jd-lara: How about ProblemSimulationData that could get closer to SimulationStoreProblemParams ?

jd-lara:
@daniel-thom the PR is in a good spot. I left comments in the places code is commented out explaining why the code is disabled. We currently only have 3 disabled tests and one of them is the simulation results.

src/core/simulation.jl
jd-lara: I think this can remain a dict

src/core/operations_problem.jl
```diff
@@ -447,12 +505,12 @@ results = solve!(OpModel)
...
 - `optimizer::MOI.OptimizerWithAttributes`: The optimizer that is used to solve the model
 """
-function solve!(problem::OperationsProblem{<:PowerSimulationsOperationsProblem}, kwargs...)
+function solve!(problem::OperationsProblem{<:PowerSimulationsOperationsProblem}; kwargs...)
     return _psi_solve_optimization_problem(problem; kwargs...)
 end
 
 """
-Default simulate method the operational model used inside of a Simulation. Solves problems that conform to the requirements of OperationsProblem{<: PowerSimulationsOperationsProblem}
+Default solve method foran operational model used inside of a Simulation. Solves problems that conform to the requirements of OperationsProblem{<: PowerSimulationsOperationsProblem}
```
kdheepak, line 513:
```suggestion
Default solve method for an operational model used inside of a Simulation. Solves problems that conform to the requirements of OperationsProblem{<: PowerSimulationsOperationsProblem}
```

src/core/optimization_container.jl
```diff
@@ -640,35 +646,51 @@ function get_problem_size(optimization_container::OptimizationContainer)
     return "The current total number of variables is $(vars) and total number of constraints is $(cons)"
 end
 
-function build_impl(
+function build_impl!(
```
kdheepak, line 649: Is `build_impl` short hand for build implementation? What do you think about using `build_problem!` here instead as the name of the function?

src/core/settings.jl
```diff
@@ -1,7 +1,7 @@
 struct Settings
     horizon::Base.RefValue{Int}
     use_forecast_data::Bool
-    use_parameters::Bool
+    use_parameters::Base.RefValue{Bool}
```
kdheepak, line 4: I think we should use `Ref{T}` instead of `Base.RefValue{T}` everywhere.

```suggestion
use_parameters::Ref{Bool}
```

If you use `Ref{Bool}`, you can pass `true` instead of `Ref(true)` to the constructor at the call site and it is stored as a `Ref` in the struct anyway.
If you use `Base.RefValue{Bool}`, you _have_ to use `Ref(true)` for the constructor at the call site. 

See screenshot for an example of this:

Screen Shot 2021-02-16 at 3 59 17 AM

I think is more ergonomic to use `Ref{T}` everywhere.

src/core/settings.jl
```diff
@@ -1,7 +1,7 @@
 struct Settings
     horizon::Base.RefValue{Int}
```
kdheepak, line 2:
```suggestion
    horizon::Base.Ref{Int}
```

src/core/settings.jl
```diff
@@ -1,7 +1,7 @@
...
     use_forecast_data::Bool
-    use_parameters::Bool
+    use_parameters::Base.RefValue{Bool}
     warm_start::Base.RefValue{Bool}
```
kdheepak, line 5:
```suggestion
    warm_start::Ref{Bool}
```

src/core/settings.jl
```diff
@@ -1,7 +1,7 @@
...
     balance_slack_variables::Bool
     services_slack_variables::Bool
```
kdheepak, line 7: Similar suggestion for `initial_time`.

src/core/settings.jl
```diff
@@ -37,7 +37,7 @@ function Settings(
     return Settings(
         Ref(horizon),
         use_forecast_data,
-        use_parameters,
+        Ref(use_parameters),
```
kdheepak, line 40: You can remove all uses of `Ref` here in this function call.

src/core/simulation_problems.jl
```diff
@@ -0,0 +1,66 @@
...
+Base.getindex(problems::SimulationProblems, key::Symbol) = problems.op_problems[key]
+Base.getindex(problems::SimulationProblems, key) = getindex(problems, Symbol(key))
+
+Base.length(problems::SimulationProblems) = length(problems.op_problems)
+Base.first(problems::SimulationProblems) = first(problems.op_problems)
+Base.iterate(problems::SimulationProblems, args...) = iterate(problems.op_problems, args...)
+
+get_problem_names(problems::SimulationProblems) = problems.names
+
+function get_problem_numer(problems::SimulationProblems, name)
```
kdheepak, line 24:
```suggestion
function get_problem_number(problems::SimulationProblems, name)
```

src/core/simulation_problems.jl
kdheepak: Is `index` a better name than `number`? To me `get_problem_index(...)` seems more appropriate.

src/core/simulation_results.jl
```diff
@@ -1,9 +1,6 @@
-const KNOWN_PATHS =
-    ["data_store", "logs", "models_json", "recorder", "results", "simulation_files"]
-
 function check_folder_integrity(folder::String)
     folder_files = readdir(folder)
-    alien_files = [f for f in folder_files if f ∉ KNOWN_PATHS]
+    alien_files = [f for f in folder_files if f ∉ KNOWN_SIMULATION_PATHS]
```
kdheepak, line 3:
```suggestion
    alien_files = filter(!∈(KNOWN_SIMULATION_PATHS), folder_files)
```

I believe in Julia 1.6 [1] you'll be able to do `filter(∉(KNOWN_SIMULATION_PATHS), folder_files)`. 

[1] https://github.com/JuliaLang/julia/pull/38475

src/core/simulation_sequence.jl
```diff
@@ -51,77 +47,46 @@ function _fill_execution_order(
...
+    function _fill_problem(index::Int, problem::Int)
+        if problem < last_problem
+            next_problem = problem + 1
+            for i in 1:interval_run_counts[next_problem]
+                index = _fill_problem(index, next_problem)
             end
         end
-        execution_order[index] = stage
+        execution_order[index] = problem
         index -= 1
```
kdheepak, line 58: What is the point of `index-=1` here?

kdheepak:
Awesome job! Overall it reads very well to me :)

kdheepak (approved):

src/core/optimization_container.jl
jd-lara: I wanted to avoid confusion when reading the code between our internal method and the ones user's would need to overload.

src/core/simulation_sequence.jl
jd-lara: This calculates the sequence from the end through the beggining

src/core/settings.jl
jd-lara: RefValue is used here to allow the field to be mutable. I tested the change and using ref doesn't allow to mutate the field in the same way. 

```julia 
julia> struct foo 
       val::Base.Ref{Bool}
       val_mut::Base.RefValue{Bool}
       end

julia> test = foo(true, Base.RefValue(true))
foo(Base.RefValue{Bool}(true), Base.RefValue{Bool}(true))

julia> test.val_mut[] = false
false

julia> test.val = false
ERROR: setfield! immutable struct of type foo cannot be changed
Stacktrace:
 [1] setproperty!(::foo, ::Symbol, ::Bool) at ./Base.jl:34
 [2] top-level scope at REPL[6]:1
 ```

PR #610, opened 2021-02-18
Cb/ptdf update

claytonpbarrows:
~~requires: https://github.com/lanl-ansi/PowerModels.jl/pull/766~~


src/core/optimization_container.jl
```diff
@@ -671,7 +671,7 @@ function build_impl!(
 
     TimerOutputs.@timeit BUILD_PROBLEMS_TIMER "Construct $(transmission)" begin
         @debug "Building $(transmission) network formulation"
-        construct_network!(optimization_container, sys, transmission)
+        construct_network!(optimization_container, sys, transmission, template)
```
jd-lara, line 674: we need to get rid of the instances where we pass the template. This is no longer needed.

claytonpbarrows: this is still needed for the `PTDFPowerModel` formulation which builds the "sparse" PTDF with PowerModels. It needs the template to understand which branches it should create PTDF rows for. I'm all for alternatives, but this is what I've been able to come up with.

jd-lara: ok, we can refactor this when we review the container construction

jd-lara:
@claytonpbarrows I am merging this PR now that the tests that work are passing so we can focus on fixing the broken tests.

PR #613, opened 2021-02-24
Updating Branch and Service tests

sourabhdalvi:


src/devices_models/device_constructors/regulationdevice_constructor.jl
```diff
@@ -11,7 +11,7 @@ function construct_device!(
         throw(ArgumentError("AGC is only compatible with AreaBalancePowerModel"))
     end
 
-    devices = get_available_components(model.device_type, sys)
+    devices = get_available_components(model.component_type, sys)
```
jd-lara, line 14: do we have getter functions for these fields? If we do please use the getters or make some. Since these are functions that potential users would re-implement then we should always try to make them use getters.

test/test_device_branch_constructors.jl
```diff
@@ -60,44 +59,270 @@ end
...
-    # TODO: use accessors to remove the use of Symbols Directly
-    qFT_line_variable = PSI.get_variable(op_problem_m.internal.optimization_container, :FqFT__MonitoredLine)
-    pFT_line_variable = PSI.get_variable(op_problem_m.internal.optimization_container, :FqFT__MonitoredLine)
     fq = JuMP.value(qFT_line_variable["1", 1])
     fp = JuMP.value(pFT_line_variable["1", 1])
-    flow = sqrt((fp[1])^2 + (fq[1])^2)
-    @test isapprox(flow, limits.from_to, atol = 1e-3)
+    flow = sqrt((fp)^2 + (fq)^2)
+    @test isapprox(flow, limits.from_to, atol = 1e-2)
+    # TODO: investigate why this test fails beyond the 1st period
```
jd-lara, line 94: Can you resolve this before merging, please.

test/test_device_branch_constructors.jl
```diff
@@ -60,44 +59,270 @@ end
...
+              PSI.BuildStatus.BUILT
+        # TODO: use accessors to remove the use of Symbols Directly
+        hvdc_line_variable =
+            PSI.get_variable(op_problem_m.internal.optimization_container, :Fp__HVDCLine)
+        tap_transformer_variable = PSI.get_variable(
+            op_problem_m.internal.optimization_container,
+            :Fp__TapTransformer,
+        )
+        transformer_variable = PSI.get_variable(
+            op_problem_m.internal.optimization_container,
```
jd-lara, line 138: use getter functions for optimization_container, please

test/test_device_branch_constructors.jl
```diff
@@ -60,44 +59,270 @@ end
...
+            optimizer = OSQP_optimizer,
+            PTDF = PSY.PTDF(system),
+        )
+        @test build!(op_problem_m; output_dir = mktempdir(cleanup = true)) ==
+              PSI.BuildStatus.BUILT
+        # TODO: use accessors to remove the use of Symbols Directly
+        hvdc_line_variable =
+            PSI.get_variable(op_problem_m.internal.optimization_container, :Fp__HVDCLine)
+        tap_transformer_variable = PSI.get_variable(
+            op_problem_m.internal.optimization_container,
```
jd-lara, line 206: same here

test/test_device_branch_constructors.jl
```diff
@@ -60,44 +59,270 @@ end
...
+        PSI.get_variable(op_problem_m.internal.optimization_container, :FqTF__HVDCLine)
+    pFT_line_variable =
+        PSI.get_variable(op_problem_m.internal.optimization_container, :FpFT__HVDCLine)
+
+    qFT_tap_variable = PSI.get_variable(
+        op_problem_m.internal.optimization_container,
+        :FqFT__TapTransformer,
+    )
+    pFT_tap_variable = PSI.get_variable(
+        op_problem_m.internal.optimization_container,
```
jd-lara, line 286: same here

jd-lara (changes requested):
Looks almost ready. Please in all the instances where you use optimization container do: 
```
optimization_container = get_optimization(container)
``` 

and then run the getter functions. Using fields with dots is discouraged in PSI

daniel-thom:
> and then run the getter functions. Using fields with dots is discouraged in PSI

At the risk of bike-shedding, can we have a bit more nuance here? I think this rule makes sense when retrieving OptimizationContainer (or things inside it) because the abstraction allows us to reorganize the internal data structures without affecting end users (or users that extend PSI functions). However, we have _internal_ (and simple) data structures like `SimulationStoreProblemParams` where the abstraction is just overkill. It's simpler to use dot access.

jd-lara:
> > and then run the getter functions. Using fields with dots is discouraged in PSI
> 
> At the risk of bike-shedding, can we have a bit more nuance here? I think this rule makes sense when retrieving OptimizationContainer (or things inside it) because the abstraction allows us to reorganize the internal data structures without affecting end users (or users that extend PSI functions). However, we have _internal_ (and simple) data structures like `SimulationStoreProblemParams` where the abstraction is just overkill. It's simpler to use dot access.

To clarify for most of the test code I prefer we use getters where possible. Using `problem.internal.optimization_container` will create maintenance problems if we decide to make changes to internal.

test/test_device_branch_constructors.jl
sourabhdalvi: sorry, that's old code we don't need to grab variable anymore, I'll remove that

test/test_device_branch_constructors.jl
```diff
@@ -52,52 +44,228 @@ end
...
-            @test !JuMP.has_upper_bound(b)
-        end
-        for b in static_line_variable
-            # Broken
-            #@test JuMP.has_lower_bound(b)
-            #@test JuMP.has_upper_bound(b)
-        end
+        @test check_variable_unbounded(op_problem_m, :Fp__MonitoredLine)
+        # Broken
+        # @test check_variable_bounded(op_problem_m, :Fp__MonitoredLine)
```
jd-lara, line 50: Is this true or is it stale?

test/test_device_branch_constructors.jl
```diff
@@ -52,52 +44,228 @@ end
...
+            template,
+            system;
+            optimizer = OSQP_optimizer,
+            PTDF = PSY.PTDF(system),
+        )
+        @test build!(op_problem_m; output_dir = mktempdir(cleanup = true)) ==
+              PSI.BuildStatus.BUILT
+
+        if model == DCPPowerModel
+            # TODO: Currently Broken, remove variable bounds in HVDCUnbounded
```
jd-lara, line 177: Is this broken too?

test/test_device_branch_constructors.jl
sourabhdalvi: Sorry that should be for static line, it does fail that test and that was a comment from the previous test code

test/test_device_branch_constructors.jl
sourabhdalvi: yup, this is where PowerModels is adding a bounds in the unbounded formulation

jd-lara (approved):
Looks good, please check again that there aren't any stale comments before merging.

sourabhdalvi:
> Looks good, please check again that there aren't any stale comments before merging.

I checked again those tests are broken

PR #614, opened 2021-02-25
WIP: network performance upgrades

jd-lara:
This PR is a proof of concept for the speedups we can get from a refactor of how we handle the network expressions, not all the implementations are super elegant (there are several ternaries if/else that look terrible) but the idea is to look into the places where we need to improve. 

@kdheepak I am tagging you for review because I made several changes that require your review.


src/network_models/copperplate_model.jl
```diff
@@ -1,19 +1,28 @@
...
+    expressions = get_expression(optimization_container, :system_balance_active)
+    remove_undef!(expressions)
+    jump_model = get_jump_model(optimization_container)
 
     for t in time_steps
-        constraint_val[t] = JuMP.@constraint(
-            optimization_container.JuMPmodel,
-            sum(optimization_container.expressions[expression].data[1:bus_count, t]) == 0
-        )
+        constraint_val[t] = JuMP.@constraint(jump_model, expressions[t] == 0)
```
jd-lara, line 10: @claytonpbarrows this implementation reduces the time to build the copperplate from 20 seconds to 400 ms.

src/devices_models/devices/common/add_variable.jl
```diff
@@ -93,12 +94,12 @@ function add_variable!(
 
         if !((expression_name === nothing))
             bus_number = PSY.get_number(PSY.get_bus(d))
+            ix = isa(optimization_container.pm, CopperPlatePowerModel) ? t : (bus_number, t)
```
jd-lara, line 97: This is some of the not pretty code

src/devices_models/devices/common/add_parameters.jl
```diff
@@ -19,12 +19,12 @@ function include_parameters!(
         param[get_component_name(r), t] =
             PJ.add_parameter(optimization_container.JuMPmodel, r.timeseries[t])
         mult[get_component_name(r), t] = r.multiplier * multiplier
+        ix = isa(optimization_container.pm, CopperPlatePowerModel) ? t : (r.bus_number, t)
```
jd-lara, line 22: Not pretty code

src/devices_models/device_constructors/branch_constructor.jl
```diff
@@ -63,38 +63,34 @@ function construct_device!(
...
-                flow_variable[name, t] == line_flow
-            )
-        end
+    flow_variables = get_variable(optimization_container, FLOW_ACTIVE_POWER, B)
+    nodal_balance_expressions =
+        get_expression(optimization_container, :nodal_balance_active)
+    jump_model = get_jump_model(optimization_container)
+    for t in time_steps, br in devices
+        br_name = get_name(br)
+        network_flow[br_name, t] = JuMP.@constraint(
```
jd-lara, line 87: @odow I resolved some of the issues we discussed here. WIthout parameters building the lines takes about 115 seconds together between network, transformers, and lines.

image

with ParameterJuMP Parameters takes 10 times more time 

image

In the sum ptdf.axes[2] is an array of integers and the function now doesn't do type inferences at run time.

blegat: Try doing:
```julia
using MutableArithmetics
const MA = MutableArithmetics
expr = zero(MA.promote_operation(MA.add_mul, eltype(flow_variables), eltype(ptdf), eltype(nodal_balance_expressions))
MA.mutable_operate!(+, expr, flow_variables[br_name, t])
for i in ptdf.axes[2]
    MA.mutable_operate!(MA.sub_mul, expr, ptdf[br_name, i], nodal_balance_expressions[i, t])
end
```
The issue with the code written by macros is that it uses `MA.operate!` which falls back to inefficient operation in case none is implemented.
With `MA.mutable_operate!`, you will either have an efficient code or an error, and the error would make it easy to know what is missing in ParameterJuMP

src/core/operations_problem.jl
```diff
@@ -371,10 +376,11 @@ function build!(
     disable_timer_outputs = false,
     serialize = true,
 )
-    !ispath(output_dir) && throw(ArgumentError("$output_dir does not exist"))
+    mkpath(output_dir)
     set_output_dir!(problem, output_dir)
     set_console_level!(problem, console_level)
     set_file_level!(problem, file_level)
+    TimerOutputs.reset_timer!(BUILD_PROBLEMS_TIMER)
```
jd-lara, line 383: @claytonpbarrows please add this time to #610

PR #617, opened 2021-03-02
Jd/network modeling patch

### LINKED ISSUE #525: Large case with PTDF stuck
OperationsProblem of a Large case with PTDF gets stuck on building.

jd-lara:
This PR should resolve most performance issues. @odow and I made several tests and this is probably the best it can be done with what we have. Further improvements will require more code development. 

Closes #525


claytonpbarrows (approved):

src/devices_models/devices/AC_branches.jl
```diff
@@ -140,6 +140,37 @@ function branch_rate_constraints!(
...
+        for br in devices
+            name = PSY.get_name(br)
+            branch_flow[br_name, t] = JuMP.@constraint(
+                jump_model,
+                sum(
+                    ptdf[br_name, i] * nodal_balance_expressions[i, t] for
+                    i in ptdf.axes[2]
+                ) == flow_variables[br_name, t]
+            )
+        end
```
odow, line 170: Did you you extracting the loop-inner into a different function? It made a small difference for me.

src/devices_models/devices/AC_branches.jl
```diff
@@ -140,6 +140,37 @@ function branch_rate_constraints!(
...
+    devices::IS.FlattenIteratorWrapper{B},
+    ::DeviceModel{B, <:AbstractBranchFormulation},
+    ::Type{StandardPTDFModel},
+    ::Nothing,
+) where {B <: PSY.ACBranch}
+    ptdf = get_PTDF(optimization_container)
+    buses = ptdf.axes[2]
+    branches = ptdf.axes[1]
+    time_steps = model_time_steps(optimization_container)
+    constraint_val = JuMPConstraintArray(undef, time_steps)
```
jd-lara, line 154: @claytonpbarrows you can remove this line. This is stale code.

src/devices_models/devices/AC_branches.jl
jd-lara: I will test that after this implementation passes the tests to see if we see the same difference.

PR #619, opened 2021-03-04
Sd/thermal update

jd-lara:


src/core/update_initial_conditions.jl
```diff
@@ -52,6 +52,15 @@ function calculate_ic_quantity(
     return current_status
 end
 
+function _get_active_power_limits(dev::T) where T <: PSY.ThermalGen
```
claytonpbarrows, line 55: `_get_min_active_power`?

jd-lara: the PR needs some adjustments.

PR #620, opened 2021-03-05
Remove redundant serialization of systems in Simulation and OperationsProblem

daniel-thom:


jd-lara (approved):
Thanks, this solves the redundancies.

PR #624, opened 2021-03-09
fix incorrect implementations

jd-lara:
@kdheepak This PR fixes an incorrect implementation of the use of warm start across the package. Due to the use of defaults, it was working but the implementation was incoherent. Some constructors would take settings others would take formulations. 

This situation leads to multiple confusions on the behavior of the package and made it hard to debug. I added types to all the functions to avoid future incorrect behaviors. 

One note about implementation, I implemented the formulations as instances just because it seemed to be your preference, makes the code review easier, and follows better the way things are currently implemented. @daniel-thom any thoughts on the use of instances vs types in the fix for the problem? 

Given the urgency to address these issues, I will keep the PR open for 24 hours. 

@sourabhdalvi please check that the Thermal Formulations have consistent results after the change.

### DIFF
--- a/src/core/service_models.jl
+++ b/src/core/service_models.jl
@@ -9,6 +9,7 @@ struct MyServiceFormulation <: PSI.AbstractServiceFormulation
 ```
 """
 abstract type AbstractServiceFormulation end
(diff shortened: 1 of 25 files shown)


daniel-thom (approved):
LGTM

PR #625, opened 2021-03-09
Moving initial condition function to device files

sourabhdalvi:


src/core/update_initial_conditions.jl
```diff
@@ -173,7 +175,7 @@ function _make_initial_conditions!(
         ini_conds = Vector{InitialCondition}(undef, length_devices)
         set_initial_conditions!(ic_container, key, ini_conds)
         for (ix, dev) in enumerate(devices)
-            val_ = get_val_func(dev, key)
+            val_ = get_val_func(dev, key, device_formulation, variable_type)
```
jd-lara, line 178: I don't fully understand why we are passing here a variable_type

jd-lara (commented):
I get the reasoning behind the changes except the passing of a variable where I made the comment.

src/core/update_initial_conditions.jl
```diff
@@ -160,6 +160,8 @@ end
 function _make_initial_conditions!(
     optimization_container::OptimizationContainer,
     devices::Union{IS.FlattenIteratorWrapper{T}, Vector{T}},
+    device_formulation::Union{AbstractDeviceFormulation, AbstractServiceFormulation},
+    variable_type::Union{Nothing, VariableType},
```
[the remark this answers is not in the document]
sourabhdalvi: The nothing is for the duration initial conditions that still uses the `_get_duration_value` function

src/core/update_initial_conditions.jl
sourabhdalvi: To use the get_variable_initial_value I need variable type, device and formulation.

jd-lara (approved):

PR #628, opened 2021-03-10
fix rh simulation

jd-lara:
Fixes a problem with the RH simulation when interval > resolution 

cc. @sourabhdalvi


claytonpbarrows (approved):
LGTM

PR #630, opened 2021-03-10
pj deprecation

jd-lara:
cc @odow

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -112,6 +112,13 @@ function write_data(
     return
 end
 
+#Given the changes in syntax in ParameterJuMP and the new format to create anonymous parameters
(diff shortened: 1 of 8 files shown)


src/utils.jl
```diff
@@ -112,6 +112,13 @@ function write_data(
     return
 end
 
+#Given the changes in syntax in ParameterJuMP and the new format to create anonymous parameters
+function add_parameter(model::JuMP.Model, val::Number)
```
jd-lara, line 116: @kdheepak does this count as type piracy?

odow: No. You own the function.

jd-lara: @odow thanks.

PR #635, opened 2021-03-11
Improve loading of simulation results

daniel-thom:
1. Serialize the status of a simulation so that if a user tries to load results for a failed simulation, it will fail by default.
2. Do not load systems by default in SimulationResults. Deserialize on the first access by the user or let the user set them manually.


jd-lara (approved):
Looks good to me. This is a great improvement for large systems.

PR #636, opened 2021-03-12
fixing realized results retreval without system

claytonpbarrows:
this PR addresses a bug where `get_realized_results` fails when `ProblemResults` is created without a `System`.


src/core/problem_results.jl
```diff
@@ -387,7 +393,7 @@ function RealizedMeta(
 )
     existing_timestamps = get_existing_timestamps(res)
     interval = existing_timestamps.step
-    resolution = PSY.get_time_series_resolution(get_system(res))
+    resolution = get_resolution(res)
     interval_len = Int(interval / resolution)
```
jd-lara, line 397: We should get this value from the end of interval value.

daniel-thom (approved):

PR #639, opened 2021-03-13
update OperationsProblemResults

claytonpbarrows:
- [x] make `OperationsProblemResults` interface consistent with `ProblemResults`
- [x] handle receeding horizon in `read_realized_...` functions


claytonpbarrows:
This seems to be failing because it's using `PowerSystemsCaseBuilder#v0.3.2`, with `v0.3.0` tests pass locally. I don't know how to pin it in the tests.

jd-lara:
> This seems to be failing because it's using `PowerSystemsCaseBuilder#v0.3.2`, with `v0.3.0` tests pass locally. I don't know how to pin it in the tests.

For some reason the CI isn't catching the compat in the tests Project.toml

src/core/operations_problem_results.jl
```diff
@@ -158,9 +166,79 @@ function write_optimizer_stats(res::OperationsProblemResults, directory::Abstrac
...
+
+function read_duals(
+    res::OperationsProblemResults;
+    names::Union{Vector{Symbol}, Nothing} = nothing,
+)
+    result_values = get_duals(res)
+    return _read_results(result_values, names, first(get_timestamps(res)))
+end
+
+function read_optimizer_stats(res::OperationsProblemResults)
```
daniel-thom, line 240: I started down this path in my PR to refactor these results because I understood that users might want consistency. However, I decided against including the `read_X` functions because it didn't seem useful. Getting back a DataFrame that can only have a single entry is less useful that just returning that single entry. If you have an OperationsProblemResults and want the optimizer stats, you should call `get_optimizer_stats()`. It appears that I included a partial `read_variables` function. That was inadvertent.

There was already existing code doing things like `res.variable_values[:P__ThermalStandard]`. I thought that was more convenient than having to do `first(values(read_variables(res)[:P__ThermalStandard]))`.

Just explaining why I didn't do this. I don't disagree with the notion of having consistent interfaces.

jd-lara:
@claytonpbarrows I merged #641 that should solve the testing issues. If you rebase we should be ok

test/Project.toml
```diff
@@ -26,4 +26,5 @@ TimerOutputs = "a759f4b9-e2f1-59dc-863e-4aeb61b1ea8f"
 UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 
 [compat]
+PowerSystemCaseBuilder = "0.3.0"
```
jd-lara, line 29: we can remove this now and update the Julia UB to 1.5

test/test_simulation_results.jl
```diff
@@ -172,11 +172,10 @@ function test_simulation_results(file_path::String, export_path)
         ren_dispatch_params =
             read_parameter(results_ed, :P__max_active_power__RenewableDispatch)
         @test length(keys(ren_dispatch_params)) == 48
-        for v in values(p_thermal_standard_ed)
-            @test size(v) == (12, 6)
+        for v in values(ren_dispatch_params)
```
jd-lara, line 175: good catch

test/test_simulation_results.jl
```diff
@@ -328,6 +327,164 @@ function test_simulation_results(file_path::String, export_path)
             SimulationResults(sim, ignore_status = true),
         )
     end
+
+    @testset "Test receeding horizon simulation results" begin
```
jd-lara, line 331: Typo Receding

jd-lara (commented):
Overall looks great. I am guessing the changes we discussed on Monday aren't incorporated yet

jd-lara (approved):

PR #641, opened 2021-03-15
Hydro Refactor

sourabhdalvi:
- Adds new constraint type for Energy Balance
- Adds new Target constraint
- Missing test for new formulation ( will be included with the storage refactor PR, as they use the new StorageTargetCost)
- Disabled Storage test ( will be re-enabled in storage refactor PR)


jd-lara (approved):
This PR is an improvement in the modeling of hydro with storage but we should be able to arrive at an even cleaner solution.

PR #643, opened 2021-03-17
Prefetch sequential reads of time series data

daniel-thom:
This PR improves performance when time series data is read during a simulation.

The current implementation repeatedly reads in very small increments. Suppose a simulation has many components with Deterministic forecasts with a horizon of 24. It will read 24 floats for every component at every execution. 24 x 8 = 192 bytes. This is very inefficient for the OS and storage system. When reading sequentially, which is what we do here, it is much better to read in large chunks.

This PR uses a feature in InfrastructureSystems to prefetch 1 MiB of chronological time series data for each time array into cache at a time and then service the small reads from that cache.

Here is before & after data from a simulation with a System that has 1350 Deterministic forecasts across components. This shows that the implementation will scale very well with simulations with increasing numbers of steps.

These simulations were run on a system with a modern, fast SSD. The gains will be significantly higher if the simulation is run on a computer with spinning disks.

```
➜  test grep ReadForecast standard/4/logs/simulation.log | head
│          Section       ncalls   time   %tot     avg     alloc   %tot      avg
│          ReadForecast  1.35k    618ms  0.10%   456μs   14.9MiB  0.02%  11.3KiB
│          ReadForecast  1.35k    593ms  0.10%   437μs   14.9MiB  0.02%  11.3KiB
│          ReadForecast  1.35k    561ms  0.09%   414μs   14.9MiB  0.02%  11.3KiB
│          ReadForecast  1.35k    562ms  0.09%   415μs   14.9MiB  0.02%  11.3KiB
│          ReadForecast  1.35k    607ms  0.10%   448μs   14.9MiB  0.02%  11.3KiB
│          ReadForecast  1.35k    557ms  0.09%   411μs   14.9MiB  0.02%  11.3KiB
│          ReadForecast  1.35k    580ms  0.09%   428μs   14.9MiB  0.02%  11.3KiB
│          ReadForecast  1.35k    562ms  0.09%   415μs   14.9MiB  0.02%  11.3KiB
│          ReadForecast  1.35k    594ms  0.10%   439μs   14.9MiB  0.02%  11.3KiB
│          ReadForecast  1.35k    570ms  0.09%   420μs   14.9MiB  0.02%  11.3KiB
➜  test grep ReadForecast standard/8/logs/simulation.log | head
│          Section       ncalls   time   %tot     avg     alloc   %tot      avg
│          ReadForecast  1.35k    11.4s  2.26%  8.41ms   7.02GiB  9.50%  5.31MiB
│          ReadForecast  1.35k   8.29ms  0.00%  6.12μs   2.54MiB  0.00%  1.92KiB
│          ReadForecast  1.35k   8.54ms  0.00%  6.30μs   2.54MiB  0.00%  1.92KiB
│          ReadForecast  1.35k   8.76ms  0.00%  6.47μs   2.54MiB  0.00%  1.92KiB
│          ReadForecast  1.35k   8.09ms  0.00%  5.97μs   2.54MiB  0.00%  1.92KiB
│          ReadForecast  1.35k   8.48ms  0.00%  6.26μs   2.54MiB  0.00%  1.92KiB
│          ReadForecast  1.35k   8.41ms  0.00%  6.21μs   2.54MiB  0.00%  1.92KiB
│          ReadForecast  1.35k   8.66ms  0.00%  6.39μs   2.54MiB  0.00%  1.92KiB
│          ReadForecast  1.35k   8.43ms  0.00%  6.22μs   2.54MiB  0.00%  1.92KiB
│          ReadForecast  1.35k   8.48ms  0.00%  6.26μs   2.54MiB  0.00%  1.92KiB
```

Before merging we need to run a comparable test on a system that transforms SingleTimeSeries to DeterministicSingleTimeSeries.

### DIFF
--- a/src/core/operations_problem.jl
+++ b/src/core/operations_problem.jl
@@ -11,6 +11,12 @@ mutable struct SimulationInfo
     sequence_uuid::Base.UUID
 end
 
+struct TimeSeriesCacheKey
+    component_uuid::Base.UUID
+    time_series_type::Type{<:IS.TimeSeriesData}
+    name::String
+end
+
 mutable struct ProblemInternal
     optimization_container::OptimizationContainer
     status::BuildStatus
@@ -18,6 +24,7 @@ mutable struct ProblemInternal
     base_conversion::Bool
     output_dir::Union{Nothing, String}
     simulation_info::Union{Nothing, SimulationInfo}
+    time_series_cache::Dict{TimeSeriesCacheKey, <:IS.TimeSeriesCache}
     ext::Dict{String, Any}
     console_level::Base.CoreLogging.LogLevel
     file_level::Base.CoreLogging.LogLevel
@@ -34,6 +41,7 @@ function ProblemInternal(
         true,
         nothing,
         nothing,
+        Dict{TimeSeriesCacheKey, IS.TimeSeriesCache}(),
         ext,
         Logging.Warn,
         Logging.Info,
@@ -273,6 +281,76 @@ get_output_dir(problem::OperationsProblem) = problem.internal.output_dir
 get_variables(problem::OperationsProblem) = get_optimization_container(problem).variables
 get_run_status(problem::OperationsProblem) = problem.internal.run_status
 set_run_status!(problem::OperationsProblem, status) = problem.internal.run_status = status
+get_time_series_cache(problem::OperationsProblem) = problem.internal.time_series_cache
+empty_time_series_cache!(x::OperationsProblem) = empty!(get_time_series_cache(x))
+
+function get_time_series_array!(
+    ts_type::Type{<:IS.TimeSeriesData},
+    problem::OperationsProblem,
+    component,
+    name,
+    initial_time;
+    ignore_scaling_factors = true,
+)
+    cache = get_time_series_cache(problem)
+    key = TimeSeriesCacheKey(IS.get_uuid(component), ts_type, name)
+    if haskey(cache, key)
+        ts_cache = cache[key]
+    else
+        ts_cache = make_time_series_cache(
+            ts_type,
+            component,
+            name,
+            initial_time,
+            ignore_scaling_factors = true,
+        )
+        cache[key] = ts_cache
+    end
+
+    ts = IS.get_next_time_series_array!(ts_cache)
+    @assert_op IS.get_initial_timestamp(ts) == initial_time
+    return ts
+end
+
+function make_time_series_cache(
+    time_series_type::Type{T},
+    component,
+    name,
+    initial_time;
+    ignore_scaling_factors = true,
+) where {T <: IS.TimeSeriesData}
+    key = TimeSeriesCacheKey(IS.get_uuid(component), T, name)
+    if T <: IS.SingleTimeSeries
+        cache = IS.StaticTimeSeriesCache(
+            PSY.SingleTimeSeries,
+            component,
+            name,
+            start_time = initial_time,
+            ignore_scaling_factors = ignore_scaling_factors,
+        )
+    elseif T <: IS.Deterministic
+        cache = IS.ForecastCache(
+            IS.AbstractDeterministic,
+            component,
+            name,
+            start_time = initial_time,
+            ignore_scaling_factors = ignore_scaling_factors,
+        )
+    elseif T <: IS.Probabilistic
+        cache = IS.ForecastCache(
+            IS.Probabilistic,
+            component,
+            name,
+            start_time = initial_time,
+            ignore_scaling_factors = ignore_scaling_factors,
+        )
+    else
+        error("not supported yet: $T")
+    end
+
+    @debug "Made time series cache for $(summary(component))" name initial_time
+    return cache
+end
 
 function get_initial_conditions(
     problem::OperationsProblem,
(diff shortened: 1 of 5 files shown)


src/core/operations_problem.jl
```diff
@@ -273,6 +281,76 @@ get_output_dir(problem::OperationsProblem) = problem.internal.output_dir
...
+    if T <: IS.SingleTimeSeries
+        cache = IS.StaticTimeSeriesCache(
+            PSY.SingleTimeSeries,
+            component,
+            name,
+            start_time = initial_time,
+            ignore_scaling_factors = ignore_scaling_factors,
+        )
+    elseif T <: IS.Deterministic
+        cache = IS.ForecastCache(
```
daniel-thom, line 332: There are customizations we could make here.
- IS allows custom cache sizes per array.
- We could disable the feature entirely with a flag. I'm inclined to add that.

jd-lara: we can also adjust that with the settings.

src/core/operations_problem.jl
```diff
@@ -365,6 +463,7 @@ function _build!(problem::OperationsProblem{<:AbstractOperationsProblem}, serial
             serialize && serialize_problem(problem)
             serialize && serialize_optimization_model(problem)
             set_status!(problem, BuildStatus.BUILT)
+            log_values(get_settings(problem))
```
daniel-thom, line 466: @jd-lara I added this so that we can have an easy-to-access record of what the user set for time_series_cache_size. I think it's a good addition overall, but let me know if you find it too noisy. It could be debug.

daniel-thom:
There is another outstanding item. I need to ensure that this feature gets disabled if the user runs a System with InMemoryTimeSeriesStorage instead of HdfTimeSeriesStorage.

claytonpbarrows:
@daniel-thom there an updated RTS example that should be suitable [here](https://github.com/NREL-SIIP/SIIPExamples.jl/blob/cb/sim_store/notebook/3_PowerSimulations_examples/02_sequential_simulations.ipynb)

src/core/simulation.jl
```diff
@@ -840,21 +840,17 @@ function update_parameter!(
...
-            count = 1,
-        )
-        ts_vector = IS.get_time_series_values(
-            d,
-            forecast,
-            initial_forecast_time;
-            len = horizon,
-            ignore_scaling_factors = true,
-        )
+        TimerOutputs.@timeit RUN_SIMULATION_TIMER "GetTimeSeries" begin
```
jd-lara, line 843: Can we remove this @timeit from here please? This operation happens way too often to record individually for every component. If anything, it might be delaying the loop. We can add the recorder too.

src/core/simulation.jl
```diff
@@ -878,20 +874,17 @@ function update_parameter!(
...
-            count = 1,
-        )
-        ts_vector = IS.get_time_series_values(
-            service,
-            forecast,
-            initial_forecast_time;
-            len = horizon,
-            ignore_scaling_factors = true,
-        )
+        TimerOutputs.@timeit RUN_SIMULATION_TIMER "GetTimeSeries" begin
```
jd-lara, line 877: same here

src/core/simulation.jl
```diff
@@ -1287,6 +1281,14 @@ struct SimulationSerializationWrapper
     name::String
 end
 
+function _empty_problem_caches!(sim::Simulation)
```
jd-lara, line 1284: can you specify this is about time series cache? since we have the simulation cache it can be confusing

daniel-thom: I actually made the name generic in case we ever wanted to empty other caches in the same place. If this doesn't change your mind then I'll change the name.

daniel-thom:
> @daniel-thom there an updated RTS example that should be suitable [here](https://github.com/NREL-SIIP/SIIPExamples.jl/blob/cb/sim_store/notebook/3_PowerSimulations_examples/02_sequential_simulations.ipynb)

Here are before & after results:
GetTimeSeries  13.2k    8.58s  10.8%   652μs    168MiB  11.3%  13.1KiB
GetTimeSeries  13.2k    383ms  0.59%  29.1μs    325MiB  19.8%  25.3KiB

src/core/simulation.jl
```diff
@@ -835,30 +835,26 @@ function update_parameter!(
...
-            forecast,
-            initial_forecast_time;
-            len = horizon,
-            ignore_scaling_factors = true,
-        )
-        component_name = PSY.get_name(d)
-        for (ix, val) in enumerate(get_parameter_array(container)[component_name, :])
-            value = ts_vector[ix]
-            JuMP.set_value(val, value)
+    TimerOutputs.@timeit RUN_SIMULATION_TIMER "update_parameter!" begin
```
jd-lara, line 838: Since this is a very specific update parameter from the forecast, as opposed to updating parameters from other results or cache. Can we make this more specific? or move it to the overall update parameters function?

src/core/simulation.jl
```diff
@@ -835,30 +835,26 @@ function update_parameter!(
...
+        components = get_available_components(T, problem.sys)
+        initial_forecast_time = get_simulation_time(sim, get_simulation_number(problem))
+        horizon = length(model_time_steps(problem.internal.optimization_container))
+        for d in components
+            # RECORDER TODO: Parameter Update from forecast
```
jd-lara, line 843: This comment can be removed given the recorder on the top

src/core/simulation.jl
```diff
@@ -872,28 +868,25 @@ function update_parameter!(
     sim::Simulation,
 ) where {T <: PSY.Service}
     # RECORDER TODO: Parameter Update from forecast
```
jd-lara, line 870: Same here

jd-lara (approved):
Some minor comments on the time series updates. We might need/want to remove the timeoutputs if the reads are faster than the time outputs over head.

PR #646, opened 2021-03-17
DNMY: Storage Refactor

sourabhdalvi:


src/devices_models/devices/common_models.jl
```diff
@@ -1,3 +1,31 @@
...
+        range_constraint_spec = RangeConstraintSpec(;
+            constraint_name = make_constraint_name(
+                RangeConstraint,
+                EnergyShortageVariable,
+                T,
+            ),
+            variable_name = make_variable_name(EnergyShortageVariable, T),
+            limits_func = x -> (
+                min = 0.0,
+                max = PSY.get_energy_shortage_cost(PSY.get_operation_cost(x)) * M_VALUE,
```
jd-lara, line 21: Why are we multiplying by M here? It feels that this adds to the model constants that the user might not be aware of and can't change. This is delicate and can massively change the behavior of the model.

src/devices_models/devices/storage.jl
```diff
@@ -40,6 +40,20 @@ get_variable_initial_value(::EnergyVariable, d::PSY.Storage, ::AbstractStorageFo
 
 get_variable_binary(::ReserveVariable, ::Type{<:PSY.Storage}, ::AbstractStorageFormulation) = true
 
+get_target_multiplier(v::PSY.BatteryEMS) = PSY.get_rating(v)
```
jd-lara, line 43: Why is this method here? This is not a model building function, it seems to just forward a method from PSY.

jd-lara (changes requested):
Some parts of the code need some further clarification. Overall looks like a great improvement in the models.

src/devices_models/devices/common_models.jl
sourabhdalvi: This is to turn off the energy shortage slack when the short cost is zero. This is mainly to avoid unbounded obj func condition when the target energy is zero e.g. in a ED problem where no target is provided but just a value of energy. This bound will never be enforce as the slack is already has ub & lb ( 0 <= slack <= storage capacity), only condition that triggers this is when the slack has no penalty cost.

src/devices_models/devices/storage.jl
sourabhdalvi: I think its just stale code from previous implementation, I'll clean it up and also review the PR again to find anything like this

src/devices_models/devices/common_models.jl
jd-lara: if the shortage cost is zero then zero*M = zero ?

sourabhdalvi: yup, this turns off the slack and you can do a energy value formulation. I guess this is a protection again a corner case where the target is zero and shortage slack has no penalty but the surplus slack as a negative cost. So, in this case you have : E + energy_shortage + energy_surplus = E_target => E + energy_shortage = - energy slack . So by taking out energy_shortage slack we are avoiding the unbounded obj func issue

src/devices_models/devices/storage.jl
sourabhdalvi: I could move the `energy_target_constraint!` in the device file and use PSY method but that would case code repetition. as the multiplier for both devices is different 
https://github.com/NREL-SIIP/PowerSimulations.jl/blob/1ebf8edb897c634234cd555e1b8056e0a8d965a9/src/devices_models/devices/storage.jl#L43
https://github.com/NREL-SIIP/PowerSimulations.jl/blob/1ebf8edb897c634234cd555e1b8056e0a8d965a9/src/devices_models/devices/hydro_generation.jl#L103

src/devices_models/devices/common_models.jl
jd-lara: Then this should be added in the implementation of the constraint and not here. Can you move the M_VALUE to be used in the implementation of the constraint to catch that corner case and throw a warning?

src/devices_models/devices/storage.jl
jd-lara: Some code repetition is preferable to make code more readable, we prefer dispatch at the construct device level. That's a tradeoff to make it easier to track what the models are doing.

sourabhdalvi: ok, I'll move this to the device file

jd-lara: @sourabhdalvi did you implement this change?

sourabhdalvi: I did, just forgot to remove it

jd-lara (approved):
LGTM. Please merge once is has been properly demo-ed

PR #655, opened 2021-03-26
Fix bug by checking for empty deque

daniel-thom:
This resulted in a runtime exception if is_dirty was called when the
deque is empty.


jd-lara (approved):

jd-lara:
The release to Julia 1.6 broke the CI all over. Formatter, tests, etc.

PR #662, opened 2021-04-05
Add option to store simulation results in memory

daniel-thom:
This PR lets the user choose to store all simulation results in memory instead of in an HDF5 file. The purpose is to test & dev of simulations that store result types that are not yet supported by the HDF5 serialization layer. Even with this implementation the result data must be able to be converted to a DataFrame (we could add other types if necessary).

To enable the feature pass `in_memory = true` to `execute!`.


src/core/operations_problem.jl
```diff
@@ -750,19 +750,16 @@ function _write_model_dual_results!(
...
            should_export_dual(exports[:exports], timestamp, problem_name_str, name)
             horizon = exports[:horizon]
             resolution = exports[:resolution]
             file_type = exports[:file_type]
-            df = axis_array_to_dataframe(constraint)
-            if names(df) == ["var"]
-                # Workaround for limitation in axis_array_to_dataframe.
-                DataFrames.rename!(df, [name])
-            end
+            df = axis_array_to_dataframe(constraint, [name])
```
daniel-thom, line 762: @jd-lara The resulting dataframes for duals will now have a relevant column name instead of `var`. I think this is an improvement but let me know if I got it wrong..

src/core/simulation.jl
```diff
@@ -981,18 +990,23 @@ function execute!(sim::Simulation; kwargs...)
     file_mode = "a"
     logger = configure_logging(sim.internal, file_mode)
     register_recorders!(sim.internal, file_mode)
-    open_func = get_simulation_store_open_func(sim)
+
+    # Undocumented option for test & dev only.
```
daniel-thom, line 994: Note that we are not advertising this to users.

test/test_simulation_results.jl
```diff
@@ -399,10 +416,15 @@ function test_simulation_results(file_path::String, export_path)
...
         end
 
-        network_duals = read_dual(results_rh, :CopperPlateBalance)
-        @test length(keys(network_duals)) == 2
-        for v in values(network_duals)
-            @test size(v) == (24, 2)
+        if !in_memory
+            # TODO DT: this test doesn't cause any duals to get written. It passes
+            # for HDF5 because the datasets get initialized.
+            # Fails for InMemory because they only get assigned on a write.
```
daniel-thom, line 422: @claytonpbarrows Can you confirm that the test is incorrect?

src/core/operations_problem.jl
jd-lara: I think this is fine but we should review the code in axis_array_to_df later too.

jd-lara (approved):

PR #665, opened 2021-04-13
fixing bug in duration ic from indexing with float

claytonpbarrows:


src/devices_models/devices/common/duration_constraints.jl
```diff
@@ -268,10 +268,10 @@ function device_duration_parameters!(
             for i in (t - duration_data[ix].up + 1):t
```
jd-lara, line 268:
```suggestion
            for i in UnitRange{Int}(t - duration_data[ix].up + 1, t)
```

This resolves the problem without that many type conversions. Also, we have to be careful here if there are fractional hour time constraints how are we handling it.

jd-lara (changes requested):

jd-lara (approved):
We will probably want to refactor the time constraints to use minutes instead of hours to handle this case better.

PR #668, opened 2021-04-17
fixing increment calculation

claytonpbarrows:
If `get_increment` is intended to calculate the number of periods in the interval, then I think this PR corrects the calculation

### DIFF
--- a/src/core/simulation.jl
+++ b/src/core/simulation.jl
@@ -814,9 +814,8 @@ function get_increment(sim::Simulation, problem::OperationsProblem, cache::TimeS
     problem_name = get_name(problem)
     sequence = get_sequence(sim)
     problem_interval = get_interval(sequence, problem_name)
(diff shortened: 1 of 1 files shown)


src/core/simulation.jl
```diff
@@ -815,8 +815,8 @@ function get_increment(sim::Simulation, problem::OperationsProblem, cache::TimeS
     sequence = get_sequence(sim)
     problem_interval = get_interval(sequence, problem_name)
     horizon = get_horizon(problem)
-    problem_resolution = problem_interval / horizon
-    return float(problem_resolution / units)
+    problem_resolution = problem_interval / (horizon * units)
+    return float(problem_resolution * horizon)
```
jd-lara, line 819: I know this float conversion was here before. But we don't need it anymore

jd-lara (approved):

Issue #657, opened 2021-03-29
multi-problem simulation executions counted improperly


claytonpbarrows:
```julia
problem_def = SimulationProblems(
    MD = OperationsProblem(
        GenericOpProblem,
        template_md,
        wk_sys,
        optimizer = solver,
        system_to_file = false,
    ),
    DA = OperationsProblem(
        GenericOpProblem,
        template_da,
        da_sys,
        optimizer = solver,
        system_to_file = false,
    ),
    ED = OperationsProblem(
        GenericOpProblem,
        template_da,
        rt_sys,
        optimizer = solver,
        system_to_file = false,
    ),
)

sequence = SimulationSequence(
    problems = problem_def,
    feedforward_chronologies = Dict(
        ("MD" => "DA") => Synchronize(periods = 1),
        ("DA" => "ED") => Synchronize(periods = 24),
    ),
    intervals = Dict(
        "MD" => (Hour(72), Consecutive()),
        "DA" => (Hour(24), Consecutive()),
        "ED" => (Hour(1), Consecutive()),
    ),
    feedforward = Dict(
        ("DA", :devices, :HydroEnergyReservoir) => IntegralLimitFF(
            variable_source_problem = PSI.ACTIVE_POWER,
            affected_variables = [PSI.ACTIVE_POWER],
        ),
        ("ED", :devices, :HydroEnergyReservoir) => IntegralLimitFF(
            variable_source_problem = PSI.ACTIVE_POWER,
            affected_variables = [PSI.ACTIVE_POWER],
        ),
    ),
    #cache = Dict(("MD", "DA") => StoredEnergy(HydroEnergyReservoir, PSI.ENERGY)),
    ini_cond_chronology = InterProblemChronology(),
);

sim = Simulation(
    name = "hydro",
    steps = 1,
    problems = problem_def,
    sequence = sequence,
    simulation_folder = file_path,
)

build!(sim)
```

```julia
ERROR: AssertionError: 24.0 == 72
Stacktrace:
  [1] macro expansion
    @ ~\.julia\packages\InfrastructureSystems\sN3Ed\src\utils\assert_op.jl:34 [inlined]
  [2] _check_steps(sim::Simulation, problem_initial_times::Dict{Int64, Vector{DateTime}})
    @ PowerSimulations ~\.julia\packages\PowerSimulations\DWqCz\src\core\simulation.jl:484
  [3] macro expansion
[...]

PR #669, opened 2021-04-18
Cb/exec order

claytonpbarrows:
Addresses #657 by removing bad calculation and assertion.

### DIFF
--- a/src/core/simulation.jl
+++ b/src/core/simulation.jl
@@ -486,12 +486,6 @@ function _check_steps(
     execution_order = get_execution_order(sequence)
     for (problem_number, (problem_name, problem)) in enumerate(get_problems(sim))
         execution_counts = get_executions(problem)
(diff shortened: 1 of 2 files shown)


jd-lara:
I remember this assertion was there for some reason. Did you identified the mistake?

claytonpbarrows:
> I remember this assertion was there for some reason. Did you identified the mistake?

Yes, the `get_executions` function, returns the number of executions per simulation step as set in the `build!`. Prior to v0.12 `get_executions` would return the number of executions of a stage before transitioning to the previous stage. So, when we have execution orders like 1-2-3-3-2-3-3-1..., the old `get_executions(3)` would return 2, and the new one returns 4.

The new calculation of executions is much simpler, this assertion double checks the execution count correctly.

jd-lara (approved):

PR #671, opened 2021-04-20
Bug fix in Multi-start start type constraint 

sourabhdalvi:
Creating new time limits by rounding the start time limits from hours to time periods.

### DIFF
--- a/src/devices_models/devices/thermal_generation.jl
+++ b/src/devices_models/devices/thermal_generation.jl
@@ -690,7 +690,7 @@ function _convert_hours_to_timesteps(
     resolution::Dates.TimePeriod,
 )
     _start_times_ts = (
-        round((hr * 60) / Dates.value(Dates.Minute(resolution)), RoundUp) for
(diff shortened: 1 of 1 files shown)


src/devices_models/devices/thermal_generation.jl
```diff
@@ -685,6 +685,18 @@ end
 
 ########################### start up trajectory constraints ######################################
 
+function _convert_hours_to_timesteps(
+    start_times_hr::StartUpStages,
+    resolution::Dates.TimePeriod,
+)
+    _start_times_ts = (
+        round((hr * 60) / Dates.value(Dates.Minute(resolution)), RoundUp) for
```
jd-lara, line 693: use a const for the 60. MINUTES_IN_HOUR

jd-lara (approved):

jd-lara:
@kaydayday and @sourabhdalvi, we currently have the hot start times as the default for the case with 1 stage and the data is coming as ` start_time_limits: (hot = 0.0, warm = 0.0, cold = 1.0)`. Is this the expected outcome?

PR #672, opened 2021-04-22
Adding Ramp Reserve

sourabhdalvi:


src/services_models/reserves.jl
```diff
@@ -144,6 +146,109 @@ function service_requirement_constraint!(
...
+    initial_time = model_initial_time(optimization_container)
+    data = _get_data_for_ramp_limit(optimization_container, service, contributing_devices)
+    service_name = PSY.get_name(service)
+    if !isempty(data)
+        service_upward_rateofchange!(
+            optimization_container,
+            data,
+            make_constraint_name(RAMP, SR),
+            make_variable_name(service_name, SR),
+            service_name,
```
sourabhdalvi, line 216: the `make_constraint_name` doesn't make unique names current it will just combine the String and Service type. If you more than 1 service of type `RampReserve` you need to make multiple constraints like this, so I pass the service name and concatenate it to the constraint name. Do you have any suggestion here ?

jd-lara: This is ok for now. Please add a comment about this so we can handle it when refactoring containers.

jd-lara (approved):

PR #689, opened 2021-05-25
performance patch1

### LINKED ISSUE #685: BatteryAncialliryServices is misspelled
Should be "Ancillary"

jd-lara:
This is the first of a series of PR's to address several performance and flexibility of use issues. 

Notably 

1. Abstract the optimization container to support other containers for extensions. For instance, SDDP containers, ADMM containers or HELICs containers. 
2. Instantiate the JuMP model in the container to remove a `Union{Nothing, JuMP.Model}` that introduced type instabilities in the build functions. 
3. Enables direct mode of the optimizer as a settings kwarg. This is disabled by default since we don't know which variables/solvers might break compatibility. 
4. Removes other `Union{Nothing, Something}` in the optimization container. 

Minor 

closes #685


src/core/operations_problem.jl
```diff
@@ -429,15 +438,17 @@ function build_pre_step!(problem::OperationsProblem)
             reset!(problem)
         end
         settings = get_settings(problem)
+        if !get_use_parameters(settings)
+            set_use_parameters!(settings, built_for_simulation(problem))
+            @debug "Set use_parameters = true for use in simulation"
```
daniel-thom, line 443: Tag with a `_group`?

daniel-thom (commented):
Looks good. Just a question about a debug log message.

src/core/operations_problem.jl
jd-lara: I think we should address the groupings of the @debug in a separate PR to implement it for all.

daniel-thom (approved):

PR #690, opened 2021-05-25
remove unused declarations

jd-lara:
Removes unused declarations

### DIFF
--- a/src/devices_models/devices/hydro_generation.jl
+++ b/src/devices_models/devices/hydro_generation.jl
@@ -734,15 +734,13 @@ function NodalExpressionSpec(
 end
 
 ##################################### Water/Energy Budget Constraint ############################
-function energy_budget_constraints!(
-    optimization_container::OptimizationContainer,
-    devices::IS.FlattenIteratorWrapper{H},
-    model::DeviceModel{H, <:AbstractHydroFormulation},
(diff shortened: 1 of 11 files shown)


src/devices_models/devices/hydro_generation.jl
```diff
@@ -753,9 +751,9 @@ active power budget formulation.
 function energy_budget_constraints!(
     optimization_container::OptimizationContainer,
     devices::IS.FlattenIteratorWrapper{H},
-    model::DeviceModel{H, <:AbstractHydroFormulation},
-    system_formulation::Type{<:PM.AbstractPowerModel},
-    feedforward::Union{Nothing, AbstractAffectFeedForward},
+    ::DeviceModel{H, <:AbstractHydroFormulation},
```
daniel-thom, line 754: Do we need to put this in the style guide? I'm OK with it. Some people prefer including the name because it does inform the reader about what is being passed even if it is unused. That isn't too important if the type isn't specified but there is at least one example here where the type is not specified.

Consistency is better than anything.

jd-lara: we could add it to the style guide. To me, it makes it easier to read @code_warntype returns

daniel-thom (approved):

PR #707, opened 2021-06-25
Updating InitialConditions structs and methods

sourabhdalvi:


src/core/initialization.jl
```diff
@@ -0,0 +1,109 @@
...
+    :ThermalStandard => DeviceModel(PSY.ThermalStandard, ThermalBasicUnitCommitment),
+    :ThermalMultiStart =>
+        DeviceModel(PSY.ThermalMultiStart, ThermalBasicUnitCommitment), # Compact vs Standard representation
+    :HydroDispatch => DeviceModel(PSY.HydroDispatch, HydroDispatchRunOfRiver),
+    :HydroEnergyReservoir =>
+        DeviceModel(PSY.HydroEnergyReservoir, HydroDispatchRunOfRiver),
+    :HydroPumpedStorage =>
+        DeviceModel(PSY.HydroPumpedStorage, HydroDispatchPumpedStorage),
+    :RenewableFix => DeviceModel(PSY.RenewableFix, FixedOutput),
+    :RenewableDispatch => DeviceModel(PSY.RenewableDispatch, RenewableFullDispatch),
```
jd-lara, line 11: we should initialize renewable with Fix

src/core/initialization.jl
```diff
@@ -0,0 +1,109 @@
...
+    :GenericBattery => DeviceModel(PSY.GenericBattery, BookKeeping),
+    :BatteryEMS => DeviceModel(PSY.BatteryEMS, BookKeeping),
+    :TapTransformer => DeviceModel(PSY.TapTransformer, StaticBranch),
+    :Transformer2W => DeviceModel(PSY.Transformer2W, StaticBranch),
+    :MonitoredLine => DeviceModel(PSY.MonitoredLine, StaticBranchUnbounded),
+    :Line => DeviceModel(PSY.Line, StaticBranch),
+    :HVDCLine => DeviceModel(PSY.HVDCLine, HVDCDispatch),
+    :PowerLoad => DeviceModel(PSY.PowerLoad, StaticPowerLoad),
```
jd-lara, line 19: same for loads, they should be initialized with fix

src/core/initialization.jl
```diff
@@ -0,0 +1,109 @@
...
+    :InterruptibleLoad => DeviceModel(PSY.InterruptibleLoad, InterruptiblePowerLoad),
+    :VariableReserve => ServiceModel(PSY.VariableReserve{PSY.ReserveUp}, RangeReserve),
```
jd-lara, line 21: I don't think we need to have a default for reserves.

src/core/initialization.jl
```diff
@@ -0,0 +1,109 @@
...
+
+function _build_initialization_problem(
+    problem::OperationsProblem{M},
+    sim::Simulation,
+) where {M <: AbstractOperationsProblem}
+    settings = deepcopy(get_settings(problem))
+    set_horizon!(settings, 1)
+    template = _build_initialization_template(problem)
+    ic_op_problem = OperationsProblem{M}(template, problem.sys, settings)
+    build!(ic_op_problem, output_dir = get_internal(problem).output_dir)
```
jd-lara, line 51: we should pass the flags needed to avoid serialization of the system and the problem.

src/core/optimization_container.jl
```diff
@@ -657,13 +656,18 @@ function get_initial_conditions(
 end
 
 function get_initial_conditions(optimization_container::OptimizationContainer, key::ICKey)
-    return get_initial_conditions(optimization_container.initial_conditions, key)
+    initial_conditions = get(optimization_container.initial_conditions, key, nothing)
+    if initial_conditions === nothing
+        throw(IS.InvalidValue("initial conditions are not stored for $(key)"))
```
daniel-thom, line 661: In similar cases with variables and parameters we print a sorted list of keys that are present. That can help debug the problem.

src/core/update_initial_conditions.jl
```diff
@@ -155,123 +155,70 @@ function calculate_ic_quantity(
...
     length_devices = length(devices)
     parameters = model_has_parameters(optimization_container)
     ic_container = get_initial_conditions(optimization_container)
-    if !has_initial_conditions(ic_container, key)
+    if !haskey(ic_container, key)
         @debug "Setting $(get_entry_type(key)) initial conditions for all devices $(T) based on system data" _group =
             LOG_GROUP_INITIAL_CONDITIONS
         ini_conds = Vector{InitialCondition}(undef, length_devices)
-        set_initial_conditions!(ic_container, key, ini_conds)
+        ic_container[key] = ini_conds
```
daniel-thom, line 208: You might want to use the accessor functions defined in optimization container. One reason is that you have a debug log statement whenever a set occurs. You defeat that by performing a set here.

jd-lara: we probably want to clean up correctly the containers for IC's too

src/services_models/agc.jl
```diff
@@ -83,24 +83,20 @@ function add_variables!(optimization_container::OptimizationContainer, ::Type{T}
...
+function add_initial_condition!(
     optimization_container::OptimizationContainer,
-    services::Vector{PSY.AGC},
+    devices::Union{Vector{T}, IS.FlattenIteratorWrapper{T}},
     ::D,
-) where {D <: AbstractAGCFormulation}
-    key = ICKey(AreaControlError, PSY.AGC)
+    initial_conditions_type::Type{<:InitialConditionType},
+) where {T <: PSY.Component, D <: AbstractServiceFormulation}
     _make_initial_conditions!(
```
jd-lara, line 92: Do we still need this function?

PR #709, opened 2021-06-26
support range ff for services

jd-lara:
This PR is a bit hacky but enables passing a range_ff to services to set the reserve levels between stages. This code needs to be redone for v015 anyway so this is a temporary patch 

cc. @kdayday


claytonpbarrows (approved):
This is fine. We'll be rewriting all this soon anyway.

PR #718, opened 2021-07-08
Jd/emulation problem

jd-lara:
replacement of #716 + other breaking changes. Tests are expected to fail.


src/initial_conditions/initial_condition.jl
```diff
@@ -1,3 +1,6 @@
+"""
+    Container for the initial condition data
```
daniel-thom, line 2: Tab to the left

src/operation/decision_model.jl
```diff
@@ -320,197 +251,169 @@ function make_time_series_cache(
...
     mkpath(output_dir)
-    set_output_dir!(problem, output_dir)
-    set_console_level!(problem, console_level)
-    set_file_level!(problem, file_level)
+    set_output_dir!(model, output_dir)
+    set_console_level!(model, console_level)
+    set_file_level!(model, file_level)
     TimerOutputs.reset_timer!(BUILD_PROBLEMS_TIMER)
     disable_timer_outputs && TimerOutputs.disable_timer!(BUILD_PROBLEMS_TIMER)
     logger = configure_logging(problem.internal, "w")
```
daniel-thom, line 353: `problem` is undefined.

src/operation/decision_model.jl
```diff
@@ -584,42 +487,31 @@ results = solve!(OpModel)
...
-    @info "Write results to Store not implemented for problems $T"
+    stats = OptimizerStats(model, step)
+    write_optimizer_stats!(store, get_name(model), stats, start_time)
+    write_model_results!(store, model, start_time; exports = exports)
     return
 end
 
 """
-Default solve method for an operational model used inside of a Simulation. Solves problems that conform to the requirements of OperationsProblem{<: PowerSimulationsOperationsProblem}
+Default solve method for an operational model used inside of a Simulation. Solves problems that conform to the requirements of DecisionModel{<:  DecisionProblem}
```
daniel-thom, line 510: Nit: extra space

src/operation/operation_model_interface.jl
```diff
@@ -0,0 +1,42 @@
+# Default implementations of getter/setter functions for OperationsModel.
```
daniel-thom, line 1: This file indicates that we would require specific fields for all subtypes of OperationsModel. Are you sure about this?

src/simulation/simulation_store.jl
```diff
@@ -4,26 +4,6 @@ const STORE_CONTAINER_VARIABLES = :variables
 const STORE_CONTAINERS =
     Set((STORE_CONTAINER_DUALS, STORE_CONTAINER_PARAMETERS, STORE_CONTAINER_VARIABLES))
 
-"""
```
daniel-thom, line 7: I'm not sure why you moved this, but the filename now seems odd.

jd-lara: I moved it to the abstract types definition file to be able to define the DecisionModel before the simulation code.

src/operation/operation_model_interface.jl
jd-lara: I think we want to have those requirements but probably there is more here than what we want. A lot of these functions need to be eliminated in favor of interfaces directly with the optimization container.

src/devices_models/devices/electric_loads.jl
```diff
@@ -84,21 +83,7 @@ function DeviceRangeConstraintSpec(
     ::Type{<:PM.AbstractPowerModel},
     feedforward::Union{Nothing, AbstractAffectFeedForward},
     use_parameters::Bool,
-    use_forecasts::Bool,
 ) where {T <: PSY.ElectricLoad}
-    if (!use_parameters && !use_forecasts)
```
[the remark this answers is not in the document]
jd-lara: we are getting rid of the `use_forecasts` option

daniel-thom (approved):

PR #719, opened 2021-07-12
Propagating changes from v015 

sourabhdalvi:


src/core/variables.jl
```diff
@@ -25,6 +25,9 @@ get_component_type(::VariableKey{T, U}) where {T <: VariableType, U <: PSY.Compo
 """Struct to dispatch the creation of Active Power Variables"""
 struct ActivePowerVariable <: VariableType end
 
+"""Struct to dispatch the creation of Active Power Variables above minimum power"""
```
jd-lara, line 28: Mention in the docstring the formulations that use this.

src/devices_models/devices/common/add_dual_variable.jl
```diff
@@ -0,0 +1,101 @@
...
+    constriant_type::Type{<:ConstraintType},
+    sys::U,
+    formulation,
+) where {U <: PSY.System}
+    time_steps = get_time_steps(container)
+    add_dual_container!(container, constriant_type, U, time_steps)
+    return
+end
+
+function calculate_dual_variable_value!(
```
jd-lara, line 79: make this an internal method since it dispatches on the key. 
```suggestion
function _calculate_dual_variable_value!(
```

src/devices_models/devices/thermal_generation.jl
```diff
@@ -517,6 +563,28 @@ function calculate_aux_variable_value!(
 
     return
 end
+
+function calculate_aux_variable_value!(
+    optimization_container::OptimizationContainer,
+    key::AuxVarKey{PowerOutput, T},
```
jd-lara, line 569: we don't need the key here. Please use PowerOutput and T separately.

src/devices_models/devices/thermal_generation.jl
```diff
@@ -517,6 +563,28 @@ function calculate_aux_variable_value!(
...
+    system::PSY.System,
+) where {T <: PSY.ThermalGen}
+    devices = PSY.get_components(T, system)
+    time_steps = model_time_steps(optimization_container)
+    on_var_results = get_variable(optimization_container, OnVariable(), T)
+    p_var_results = get_variable(optimization_container, PowerAboveMinimumVariable(), T)
+    aux_var_container = get_aux_variables(optimization_container)[key]
```
jd-lara, line 576: use the appropriate getter function with the var type and T

src/devices_models/devices/thermal_generation.jl
```diff
@@ -517,6 +563,28 @@ function calculate_aux_variable_value!(
...
+
+    for d in devices, t in time_steps
+        name = PSY.get_name(d)
+        active_limits = PSY.get_active_power_limits(d)
```
jd-lara, line 580:
```suggestion
        min = PSY.get_active_power_limits(d).min
```

src/devices_models/devices/thermal_generation.jl
```diff
@@ -517,6 +563,28 @@ function calculate_aux_variable_value!(
...
+        aux_var_container[name, t] =
+            JuMP.value(on_var_results[name, t]) * active_limits.min +
```
jd-lara, line 582:
```suggestion
            JuMP.value(on_var_results[name, t]) * min +
```

src/network_models/network_constructor.jl
```diff
@@ -1,13 +1,13 @@
...
     sys::PSY.System,
-    ::Type{CopperPlatePowerModel},
+    model::NetworkModel{CopperPlatePowerModel},
     template::ProblemTemplate,
 )
     buses = PSY.get_components(PSY.Bus, sys)
     bus_count = length(buses)
 
-    if get_balance_slack_variables(container.settings)
+    if get_use_slacks(model)
```
jd-lara, line 10:
```suggestion
    get_use_slacks(model) && add_slacks!(container, CopperPlatePowerModel)
```

src/core/device_model.jl
```diff
@@ -32,15 +32,15 @@ thermal_gens = DeviceModel(ThermalStandard, ThermalBasicUnitCommitment),
 mutable struct DeviceModel{D <: PSY.Device, B <: AbstractDeviceFormulation}
     feedforward::Union{Nothing, AbstractAffectFeedForward}
     use_slacks::Bool
-    duals::Vector{<:ConstraintType}
+    duals::Vector
```
daniel-thom, line 35: Just curious, what is the reason for this change? Are we no longer bounding the types?

src/core/optimization_container.jl
```diff
@@ -409,6 +416,28 @@ function get_aux_variable_keys(container::OptimizationContainer)
     return collect(keys(container.aux_variables))
 end
 
+##################################### DualVariable Container ################################
+function add_dual_container!(
```
jd-lara, line 420: Wherever this is called we need to check if the container is compatible i.e., check if its MILP

src/core/device_model.jl
jd-lara: This is actually a `Vector{DataType}`.

jd-lara (approved):
There are some minor changes requested. You can merge after those are addressed.

src/devices_models/devices/thermal_generation.jl
sourabhdalvi: we do use the key to get the aux container `aux_var_container = get_aux_variables(container)[key]`. This is just following the same template as TimeDurationOff/On

src/devices_models/devices/thermal_generation.jl
sourabhdalvi: I'm not sure what you mean, there are only two `get_variable` methods one uses VariableKey and the other is the one I use.

src/devices_models/devices/thermal_generation.jl
jd-lara: we then need a method similar to: 

```Julia
function get_variable(
    ::T,
    ::Type{U},
    meta::String = CONTAINER_KEY_EMPTY_META,
) where {T <: VariableType, U <: PSY.Component}
    return get_variable(container, VariableKey(T, U, meta))
end
```

for get_aux_variable

src/core/device_model.jl
daniel-thom: So is it supposed to be defined as `Vector{DataType}`?

jd-lara: yes, @sourabhdalvi we can make this `Vector{DataType}` since it stores types of constraints.

daniel-thom (approved):

test/test_operations_problem.jl
```diff
@@ -151,18 +141,19 @@ end
     parameters = [true, false]
     ptdf = PTDF(sys)
     # These are the duals of interest for the test
-    dual_constraint =
-        [[:nodal_balance_active__Bus], [:CopperPlateBalance, :network_flow__Line]]
+    dual_constraint = [[NodalBalanceActiveConstraint], [CopperPlateBalanceConstraint]]
```
daniel-thom, line 144: @sourabhdalvi @jd-lara I'm now working with this code. Should this interface take a ConstraintKey instead of ConstraintType (or tuple that can be used to create a ConstraintKey)? Before this PR I had written some code like this:
```
dual_constraint = [[(NodalBalanceActiveConstraint, PSY.Bus)], [(CopperPlateBalanceConstraint, PSY.System]]
```

test/test_operations_problem.jl
```diff
@@ -171,52 +162,45 @@ end
...
 
     container = PSI.get_optimization_container(model)
-    constraints = PSI.get_constraints(container)[PSI.ConstraintKey(
-        CopperPlateBalanceConstraint,
-        PSY.System,
-    )]
-    dual_results = read_duals(container, [:CopperPlateBalance])
+    constraint_key = PSI.ConstraintKey(CopperPlateBalanceConstraint, PSY.System)
+    constraints = PSI.get_constraints(container)[constraint_key]
+    dual_results = read_duals(container)[:CopperPlateBalanceConstraint_System]
```
daniel-thom, line 185: @sourabhdalvi @jd-lara Similar to above, I was planning to have something like
```
dual_results = read_duals(container)[(CopperPlateBalanceConstraint, PSY.System)]
```

test/test_operations_problem.jl
jd-lara: Since the dual constraint type gets passed with the model we can build the key at the construction time. The only exception is the case of the network where the device portion can be a system or a bus depending on the constraint type we want de dual from. 

I think we can still pass the constraint type and fill the remaining portion after since the system applies only in the case of the CopperPlateBalanceConstraint and Bus in any other constraint type.

test/test_operations_problem.jl
jd-lara: forgetting the results I think this is fine, albeit inconsistent with the other interface but the dual interface will be for programmatic querying of the duals and it isn't too bad.

daniel-thom: Shouldn't the interfaces in the user code be symmetric? Otherwise, how will they know what to pass to `read_duals`?

To be clear I mean
```
NetworkModel(network; PTDF = ptdf, duals = (CopperPlateBalanceConstraint, PSY.System))
read_duals(container, (CopperPlateBalanceConstraint, PSY.System))
```

jd-lara: I get what you mean 100% and agree that there is inconsistency. I am not sure what is worse that's all because in the case of the device and service model we don't need to pass the device type for the duals.

PR #723, opened 2021-07-22
WIP: Refactor results to use OptimizationContainerKey

daniel-thom:
Functionality is limited and this code still needs some work. This branch does allow:
- Store results in the simulation store.
- Read results back.
- Use container keys everywhere.


src/core/abstract_types.jl
```diff
@@ -10,24 +10,41 @@ abstract type AuxVariableType end
 abstract type ParameterType end
 abstract type InitialConditionType end
 
+const _DELIMITER = "_"
```
jd-lara, line 13: Given that the key code is growing should we separate it into a new file?

src/core/optimization_container.jl
```diff
@@ -1,12 +1,40 @@
+struct OptimizationContainerMetadata
```
jd-lara, line 1: Do we need to make this a separate struct?

src/core/optimization_container.jl
```diff
@@ -1,12 +1,40 @@
...
+
+add_container_key!(x::OptimizationContainerMetadata, key, val) =
+    x.container_key_lookup[key] = val
+get_container_key(x::OptimizationContainerMetadata, key) = x.container_key_lookup[key]
+has_container_key(x::OptimizationContainerMetadata, key) =
+    haskey(x.container_key_lookup, key)
+
 mutable struct OptimizationContainer <: AbstractModelContainer
-    JuMPmodel::JuMP.Model
+    JuMPmodel::Union{Nothing, JuMP.Model}
```
jd-lara, line 30: Why did this one go back to `::Union{Nothing, JuMP.Model}`?

src/devices_models/devices/common/add_dual_variable.jl
```diff
@@ -5,8 +5,8 @@ function add_dual_variable!(
 ) where {T <: PSY.Component, D <: AbstractDeviceFormulation}
     if !isempty(get_duals(model))
         devices = get_available_components(T, sys)
-        for constriant_key in get_duals(model)
-            assign_dual_variable!(container, constriant_key, devices, D)
+        for constraint_type in get_duals(model)
```
jd-lara, line 8: The duals are keyed with a ConstraintKey`, are you sure you want to keep this name?

src/devices_models/devices/storage.jl
```diff
@@ -300,7 +300,7 @@ function energy_target_constraint!(
     ::Union{Nothing, AbstractAffectFeedForward},
 ) where {T <: PSY.Storage}
     time_steps = get_time_steps(container)
-    target_forecast_name = "storage_target"
+    target_forecast_name = "storage_target"  # TODO DT: should this be a const?
```
jd-lara, line 303: These will be extracted from the `DeviceModel` for the final release of v015

src/simulation/hdf_simulation_store.jl
```diff
@@ -25,16 +25,16 @@ DEFAULT_MAX_CHUNK_BYTES = 128 * KiB
 Stores HDF5 datasets for one problem.
 """
 mutable struct ProblemDatasets
-    duals::Dict{Symbol, Dataset}
-    parameters::Dict{Symbol, Dataset}
-    variables::Dict{Symbol, Dataset}
+    duals::Dict{OptimizationContainerKey, Dataset}
```
jd-lara, line 28: Should we make this the appropriate key?

src/simulation/hdf_simulation_store.jl
```diff
@@ -104,22 +108,32 @@ df = PowerSimulations.read_result(DataFrame, store, model, :variables, var_name,
...
-    mode = "r",
+    mode = "r";
     filename = HDF_FILENAME,
+    problem_path = nothing,
 )
-    return HdfSimulationStore(joinpath(directory, filename), mode)
+    return HdfSimulationStore(
+        joinpath(directory, filename),
+        mode,
+        problem_path = problem_path,
```
jd-lara, line 118: Should the path be the same as the directory? I don't understand why we pass 2 different paths.

test/test_simulation_execute.jl
```diff
@@ -203,20 +206,20 @@ end
...
     c_sys5_hy_uc = PSB.build_system(PSITestSystems, "c_sys5_hy_uc")
     c_sys5_hy_ed = PSB.build_system(PSITestSystems, "c_sys5_hy_ed")
-    problems = SimulationProblems(
+    problems = SimulationModels(
         UC = DecisionModel(
             template_uc,
             c_sys5_hy_uc;
             optimizer = GLPK_optimizer,
-            constraint_duals = [:CopperPlateBalance],
+            constraint_duals = [ConstraintKey(CopperPlateBalanceConstraint, PSY.System)],
```
jd-lara, line 214: This interface has changed so we probably want to remove all the calls `constraint_duals =` in `DecisionModel` constructors.

jd-lara (commented):
Generally speaking, looks good, I have some questions about the paths. It seems we need to handle better the folder structure of the simulations.

src/devices_models/devices/common/add_dual_variable.jl
daniel-thom: This should actually fail. It should be
```
        for constraint_key in keys(get_duals(model))
            assign_dual_variable!(container, get_entry_type(constraint_key), devices, D)
        end
```
Same with other cases below.

src/core/optimization_container.jl
daniel-thom: This allows us to deserialize this one object apart from the OptimizationContainer.

src/simulation/hdf_simulation_store.jl
daniel-thom: I did it correctly for the in_memory store. Fixed now.

test/test_simulation_execute.jl
daniel-thom: This whole section is wrong and will fail to run. I'll leave it here so that we remember to fix the definition of the duals within the model.

jd-lara (approved):
Thanks, the changes looks great we can now continue with the overall refactor of simulations to re-enable the other tests.

PR #724, opened 2021-07-23
Adding methods to check numerical bounds for JuMP model

sourabhdalvi:


src/jump_model_utils.jl
```diff
@@ -0,0 +1,147 @@
+########### JuMP model utils #########
+
+mutable struct NumericalBounds
+    min::Float64
+    max::Float64
+    min_index::Any
+    max_index::Any
+end
+
+NumericalBounds() = NumericalBounds(Inf, -Inf, nothing, nothing)
```
jd-lara, line 10: We should add at the top of the file that is based on the code in SDDP and reference the file and commit.

jd-lara (changes requested):
We need test code for these utilities. You can start a new test file and use a very small model. Like a 5 bus ED.

src/jump_model_utils.jl
sourabhdalvi: Yup, I'll add that

odow: Don't bother. I approve the release of this under the current PowerSimulations.jl license without the copyright notice.

src/jump_model_utils.jl
```diff
@@ -0,0 +1,157 @@
+# The Numerical stability checks code in this file is based on the code from the SDDP.jl package, 
+ # from the below mentioned commit and file.
+ # commit :8cd305188caffc50a1734913053fc81bba613778 
+ # link to file :https://github.com/odow/SDDP.jl/blob/d353fe5a2903421e7fed6d609eb9377c35d715a1/src/print.jl#L190
+
```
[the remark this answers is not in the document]
odow: Woah. How do I do this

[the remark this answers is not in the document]
jd-lara: @odow you can copy this: https://github.com/NREL-SIIP/PowerSimulations.jl/blob/master/.github/workflows/format-check.yml

jd-lara (approved):

PR #728, opened 2021-08-02
Constraint refactor for Hydro and Storage devices

sourabhdalvi:


src/devices_models/devices/interfaces.jl
```diff
@@ -7,3 +7,4 @@ get_variable_expression_name(_, ::Type{<:PSY.Component}) = nothing
 get_variable_initial_value(_, ::PSY.Component, __) = nothing
 get_variable_lower_bound(_, ::PSY.Component, __) = nothing
 get_variable_upper_bound(_, ::PSY.Component, __) = nothing
+get_multiplier_value(_, ::PSY.Component, __) = 1.0
```
jd-lara, line 10: This default value can be problematic since it is a valid default. I would prefer this to be a `NaN`

jd-lara (approved):
LGTM, I will make the change for the default value and merge.

PR #729, opened 2021-08-02
refactoring branch constraint construction

claytonpbarrows:


src/devices_models/device_constructors/branch_constructor.jl
```diff
@@ -111,8 +136,25 @@ function construct_device!(
     if !validate_available_devices(B, devices)
         return
     end
+
     branch_rate_bounds!(container, devices, model, S)
```
claytonpbarrows, line 140: Note that the bound setting needs to be refactored

src/devices_models/devices/AC_branches.jl
```diff
@@ -109,52 +110,66 @@ function branch_rate_constraints!(
...
-function branch_flow_values!(
+function add_constraints!(
     container::OptimizationContainer,
+    cons_type::Type{NetworkFlowConstraint},
     devices::IS.FlattenIteratorWrapper{B},
-    ::DeviceModel{B, <:AbstractBranchFormulation},
-    network_model::NetworkModel{StandardPTDFModel},
-) where {B <: PSY.ACBranch}
+    model::DeviceModel{B, <:AbstractBranchFormulation},
+    network_model::NetworkModel{S},
```
claytonpbarrows, line 170: note that this signature differs from the other `add_constraints!` methods because we need the entire `network_model` to get the PTDF

jd-lara: good stuff here with the PTDF.

jd-lara (approved):

PR #731, opened 2021-08-12
Adding attributes to device model

sourabhdalvi:


src/core/device_model.jl
```diff
@@ -68,7 +71,7 @@ get_services(::Nothing) = nothing
 get_use_slacks(m::DeviceModel) = m.use_slacks
 get_duals(m::DeviceModel) = m.duals
 get_time_series_labels(m::DeviceModel) = m.time_series_labels
-
+get_attributes(m::DeviceModel) = m.attributes
```
jd-lara, line 74: also add a getter like this for simplicity. 

```
get_attribute(m, attr::String) = m.attributed[attr]
```

src/core/device_model.jl
```diff
@@ -82,9 +85,20 @@ function _set_model!(
...
 end
+
+function initialize_attributes(
+    ::Type{D},
+    ::Type{T},
+) where {D <: PSY.Device, T <: AbstractDeviceFormulation}
+    return Dict{String, Any}()
+end
+
+function apply_reservations(model::DeviceModel)
```
jd-lara, line 102: We should use the get attribute function so that it is easier to read in the construction code. Since apply reservation isn't defined for all device models, this could be confusing

src/devices_models/devices/common/range_constraint.jl
```diff
@@ -152,6 +152,159 @@ function add_semicontinuous_range_constraints!(
...
+    feedforward::Union{Nothing, AbstractAffectFeedForward},
+) where {V <: PSY.Component, W <: AbstractDeviceFormulation}
+    use_parameters = built_for_simulation(container)
+    constraint = T()
+    variable = U()
+    component_type = V
+    time_steps = get_time_steps(container)
+    jump_variable = get_variable(container, variable, component_type)
+    names = [PSY.get_name(d) for d in devices]
+    binary_variables = [ReservationVariable()]
```
jd-lara, line 205: @kdheepak can we just use the variable. I don't see a reason to have a vector or size 1 here.

src/devices_models/devices/common/range_constraint.jl
```diff
@@ -152,6 +152,159 @@ function add_semicontinuous_range_constraints!(
...
+
+where limits in constraint_infos.
+
+# LaTeX
+
+`` limits^{min} x^{bin} \leq x^{cts} \leq limits^{max} x^{bin},``
+"""
+function add_reserve_range_constraints!(
+    container::OptimizationContainer,
+    T::Type{<:PowerVariableLimitsConstraint},
```
jd-lara, line 266: Should this be `OutputActivePowerVariableLimitsConstraint`

jd-lara (changes requested):
There are some places where it isn't clear if the variable is the input or output one. We need to correct that.

src/devices_models/devices/common/range_constraint.jl
sourabhdalvi: This is also used for the ReactivePowerLimits constraints, that's why I left it like that

src/core/device_model.jl
sourabhdalvi: ok, I'll remove this and use the getter function

src/devices_models/devices/hydro_generation.jl
```diff
@@ -217,29 +187,22 @@ function add_constraints!(
...
 """
-function DeviceRangeConstraintSpec(
-    ::Type{<:ActivePowerVariableLimitsConstraint},
-    ::Type{ActivePowerVariable},
-    ::Type{T},
-    ::Type{<:AbstractHydroDispatchFormulation},
-    ::Type{<:PM.AbstractPowerModel},
+function add_constraints!(
+    container::OptimizationContainer,
+    T::Type{<:PowerVariableLimitsConstraint},
```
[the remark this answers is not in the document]
sourabhdalvi: the function would be be the same for input and output, we are dispatching on the constraint type in the `add_reserve_range_constraints!` so I thought just combining them would reduce code repetition

src/devices_models/devices/common/range_constraint.jl
```diff
@@ -263,13 +263,18 @@ where limits in constraint_infos.
 """
 function add_reserve_range_constraints!(
     container::OptimizationContainer,
-    T::Type{<:PowerVariableLimitsConstraint},
+    T::Type{Z},
```
jd-lara, line 266: Nit: can you order the variables correctly here, please. The last is the first and has the Z value.

src/devices_models/devices/common/range_constraint.jl
jd-lara: bump @kdheepak about this

PR #733, opened 2021-08-19
Record relative path to file

### LINKED ISSUE #684: Reading in a simulation results after changing directory fails
Expect this to issue for most people running simulation on HPC, I save my results on the tmp/scratch/ on the compute node and move the results directory to Lustre after the simulation is complete. The error occurs when reading in the results after it's moved, because its using the file path stored in the simulation_store.h5 and not the path passed.
https://github.com/NREL-SIIP/PowerSimulations.jl/blob/772adeccde3370a5e19c72f54bcb134dacec61f7/src/utils.jl#L386-L389

daniel-thom:
This change allows a simulation results directory to be moved or renamed. #Closes #684

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -39,7 +39,8 @@ function compute_file_hash(path::String, files::Vector{String})
     data = Dict("files" => [])
     for file in files
         file_path = joinpath(path, file)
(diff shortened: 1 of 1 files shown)


jd-lara (approved):
This is a great improvement. We should apply this change to V015 too.

sourabhdalvi (approved):

PR #734, opened 2021-08-20
Generalize exception type

timholy:
In preparation for https://github.com/JuliaLang/julia/pull/41885


jd-lara:
> In preparation for [JuliaLang/julia#41885](https://github.com/JuliaLang/julia/pull/41885)

Thanks for the heads up

jd-lara (approved):

PR #746, opened 2021-08-30
Fixing multiple issue

sourabhdalvi:
-  updating ICkey methods
- Bug fixes  hydro
- Bug fixes  load
- Fixes the PTDFPowerModel (but one test is still failing obj values don't match)
- Changes to VariableKey code

### DIFF
--- a/src/core/optimization_container.jl
+++ b/src/core/optimization_container.jl
@@ -310,8 +310,8 @@ end
 function _make_system_expressions!(
     container::OptimizationContainer,
     bus_numbers::Vector{Int},
-    ::Type{StandardPTDFModel},
-)
+    ::Type{T},
+) where {T <: Union{PTDFPowerModel, StandardPTDFModel}}
     parameter_jump = built_for_simulation(container)
     time_steps = get_time_steps(container)
     container.expressions = Dict(
@@ -524,7 +524,7 @@ function _add_var_container!(
     var_key::VariableKey{T, U},
     sparse::Bool,
     axs...,
-) where {T <: VariableType, U <: PSY.Component}
+) where {T <: VariableType, U <: Union{PSY.Component, PSY.System}}
     if sparse
         var_container = sparse_container_spec(JuMP.VariableRef, axs...)
     else
@@ -540,7 +540,7 @@ function add_var_container!(
     ::Type{U},
     axs...;
     sparse = false,
-) where {T <: VariableType, U <: PSY.Component}
+) where {T <: VariableType, U <: Union{PSY.Component, PSY.System}}
     var_key = VariableKey(T, U)
     return _add_var_container!(container, var_key, sparse, axs...)
 end
@@ -552,7 +552,7 @@ function add_var_container!(
     meta::String,
     axs...;
     sparse = false,
-) where {T <: VariableType, U <: PSY.Component}
+) where {T <: VariableType, U <: Union{PSY.Component, PSY.System}}
     var_key = VariableKey(T, U, meta)
     return _add_var_container!(container, var_key, sparse, axs...)
 end
@@ -577,7 +577,7 @@ function get_variable(
     ::T,
     ::Type{U},
     meta::String = CONTAINER_KEY_EMPTY_META,
-) where {T <: VariableType, U <: PSY.Component}
+) where {T <: VariableType, U <: Union{PSY.Component, PSY.System}}
     return get_variable(container, VariableKey(T, U, meta))
 end
 
@@ -836,7 +836,7 @@ function read_parameters(container::OptimizationContainer)
     parameters = get_parameters(container)
     (isnothing(parameters) || isempty(parameters)) && return params_dict
     for (k, v) in parameters
-        !isa(v.update_ref, UpdateRef{<:PSY.Component}) && continue
+        !isa(get_component_type(k), PSY.Component) && continue
         param_array = axis_array_to_dataframe(get_parameter_array(v))
         multiplier_array = axis_array_to_dataframe(get_multiplier_array(v))
         params_dict[k] = param_array .* multiplier_array
--- a/src/core/optimization_container.jl
+++ b/src/core/optimization_container.jl
@@ -944,10 +944,11 @@ function get_initial_conditions_keys(container::OptimizationContainer)
     return collect(keys(container.initial_conditions))
 end
 
-function set_initial_conditions!(container::OptimizationContainer,
+function set_initial_conditions!(
+    container::OptimizationContainer,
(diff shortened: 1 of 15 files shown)


src/core/optimization_container.jl
```diff
@@ -944,6 +944,15 @@ function get_initial_conditions_keys(container::OptimizationContainer)
     return collect(keys(container.initial_conditions))
 end
 
+function set_initial_conditions!(
+    container::OptimizationContainer,
+    ::Type{T},
+    ::Type{D},
+    value,
```
jd-lara, line 951: Should the value be a Float64?

src/devices_models/devices/common/range_constraint.jl
```diff
@@ -549,47 +549,47 @@ function add_parameterized_upper_bound_range_constraints(
...
-#
-#         # TODO: deal with additional terms
-#
-#         constraint[name, t] = JuMP.@constraint(
-#             container.JuMPmodel,
-#             expression_ub <= multiplier[name, t] * parameter[name, t]
-#         )
-#     end
-# end
+function add_parameterized_upper_bound_bigM_range_constraints(
```
jd-lara, line 552: which model uses this constraint? I looked into the code and I think we never implemented it because it was better to have range + ts_ub

src/devices_models/devices/common/range_constraint.jl
```diff
@@ -549,47 +549,47 @@ function add_parameterized_upper_bound_range_constraints(
...
+    T::Type{<:ConstraintType},
+    U::Type{<:VariableType},
+    P::Type{<:ParameterType},
+    devices::IS.FlattenIteratorWrapper{V},
+    model::DeviceModel{V, W},
+    X::Type{<:PM.AbstractPowerModel},
+    feedforward::Union{Nothing, AbstractAffectFeedForward},
+) where {V <: PSY.Component, W <: AbstractDeviceFormulation}
+
+    # TODO: the following is incorrect implementation of bigM constraints
```
jd-lara, line 563: Also this

src/devices_models/devices/hydro_generation.jl
```diff
@@ -178,6 +178,21 @@ function get_min_max_limits(
...
+function add_constraints!(
+    container::OptimizationContainer,
+    T::Type{<:PowerVariableLimitsConstraint},
+    U::Type{<:VariableType},
+    devices::IS.FlattenIteratorWrapper{V},
+    model::DeviceModel{V, W},
+    X::Type{<:PM.AbstractPowerModel},
+    feedforward::Union{Nothing, AbstractAffectFeedForward},
+) where {V <: PSY.HydroGen, W <: AbstractHydroUnitCommitment}
+    add_range_constraints!(container, T, U, devices, model, X, feedforward)
```
jd-lara, line 193: In UC should this be a semicontinous?

src/devices_models/devices/thermal_generation.jl
```diff
@@ -316,10 +316,8 @@ function add_constraints!(
     W::Type{<:PM.AbstractPowerModel},
     feedforward::Union{Nothing, AbstractAffectFeedForward},
 ) where {T <: PSY.ThermalGen, S <: AbstractCompactUnitCommitment}
-    key_power = ICKey(DevicePower, T)
-    key_status = ICKey(DeviceStatus, T)
-    initial_conditions_power = get_initial_conditions(container, key_power)
-    initial_conditions_status = get_initial_conditions(container, key_status)
+    initial_conditions_power = get_initial_conditions(container, DevicePower, T)
+    initial_conditions_status = get_initial_conditions(container, DeviceStatus, T)
```
jd-lara, line 320: same here 

```suggestion
    initial_conditions_status = get_initial_conditions(container, DeviceStatus(), T)
```

src/devices_models/devices/thermal_generation.jl
```diff
@@ -1114,10 +1110,8 @@ function add_constraints!(
     parameters = built_for_simulation(container)
     resolution = get_resolution(container)
     # Use getter functions that don't require creating the keys here
```
jd-lara, line 1112: remove stale comment

src/services_models/agc.jl
```diff
@@ -219,7 +219,7 @@ function smooth_ace_pid!(container::OptimizationContainer, services::Vector{PSY.
                 )
             if t == 1
                 SACE_ini =
-                    get_initial_conditions(container, ICKey(AreaControlError, PSY.AGC))[ix]
+                    get_initial_conditions(container, AreaControlError, PSY.AGC)[ix]
```
jd-lara, line 222:
```suggestion
                    get_initial_conditions(container, AreaControlError(), PSY.AGC)[ix]
```

jd-lara (changes requested):
The main thing the PR Needs is to use IC instances vs types in the getter functions to be consistent with the other interfaces.

src/devices_models/devices/common/range_constraint.jl
sourabhdalvi: its used in the electric load file

src/devices_models/devices/common/range_constraint.jl
sourabhdalvi: I wasn't sure is that was a stale comment, because the constraint makes sense  but I'll comment it out again

src/core/optimization_container.jl
```diff
@@ -310,8 +310,8 @@ end
 function _make_system_expressions!(
     container::OptimizationContainer,
     bus_numbers::Vector{Int},
-    ::Type{StandardPTDFModel},
-)
+    ::Type{T},
+) where {T <: Union{PTDFPowerModel, StandardPTDFModel}}
```
jd-lara, line 314: @sourabhdalvi are you sure we need this here? I think that the CopperPlate is handled directly by PM

src/devices_models/devices/common/add_to_expression.jl
```diff
@@ -212,7 +212,7 @@ function add_to_expression!(
     U <: TimeSeriesParameter,
     V <: PSY.StaticInjection,
     W <: AbstractDeviceFormulation,
-    X <: StandardPTDFModel,
+    X <: Union{PTDFPowerModel, StandardPTDFModel},
```
jd-lara, line 215: Same here. I think that the `PTDFPowerModel` just needs the addition to the nodals expressions.

src/core/optimization_container.jl
sourabhdalvi: ok, I'll look into that again

jd-lara (approved):

PR #748, opened 2021-09-01
Updating Services and Fixing minor bugs

sourabhdalvi:
Created a new PR with cleaner commit history


src/devices_models/device_constructors/common/constructor_validations.jl
```diff
@@ -12,58 +12,53 @@ function validate_available_devices(
...
-            typeof(d) ∉ incompatible_device_types && PSY.get_available(d)
-        ]
-        if isempty(contributing_devices)
-            @warn "The contributing devices for service $(PSY.get_name(s)) is empty, consider removing the service from the system" _group =
-                :ConstructGroup
-        else
-            push!(services, s)
-        end
+
+    contributing_devices_ =
```
jd-lara, line 29: I wonder if we should be caching this calculation somewhere else. But we can improve this later.

src/devices_models/device_constructors/common/constructor_validations.jl
```diff
@@ -12,58 +12,53 @@ function validate_available_devices(
...
-    service::Type{S},
-    services::Vector{S},
+    model::ServiceModel{S, <:AbstractServiceFormulation},
     ::Vector{<:DataType},
     sys::PSY.System,
 ) where {S <: PSY.StaticReserveGroup}
-    services_ = PSY.get_components(S, sys)
-    if isempty(services_)
-        @warn "The data doesn't include services of type $(service), consider changing the service models" _group =
+    service = PSY.get_components(S, sys, get_service_name(model))
```
jd-lara, line 49: How is this working? the correct syntax is `get_component(S, sys, name)` maybe this isn't being tested

src/devices_models/devices/hydro_generation.jl
```diff
@@ -178,6 +179,14 @@ function get_min_max_limits(
     PSY.get_active_power_limits(x)
 end
 
+function get_min_max_limits(
+    x::PSY.HydroGen,
+    ::Type{<:ActivePowerVariableLimitsConstraint},
+    ::Type{HydroDispatchRunOfRiver},
+)
+    return (min = 0.0, max = PSY.get_active_power_limits(x).max)
```
jd-lara, line 187: you can use `get_max_active_power` directly

src/devices_models/devices/storage.jl
```diff
@@ -211,51 +211,98 @@ function add_constraints!(
...
-        up_info = DeviceRangeConstraintInfo(name, PSY.get_output_active_power_limits(d))
-        down_info = DeviceRangeConstraintInfo(name, PSY.get_input_active_power_limits(d))
-        energy_info = ReserveRangeConstraintInfo(
-            name,
-            PSY.get_state_of_charge_limits(d),
-            PSY.get_efficiency(d),
-            T,
+        expression_up = JuMP.AffExpr(0.0)
+        expression_dn = JuMP.AffExpr(0.0)
+        # TODO: add reserve variables
```
jd-lara, line 246: Is this comment still valid or is it a stale comment?

src/devices_models/devices/storage.jl
```diff
@@ -211,51 +211,98 @@ function add_constraints!(
...
+        names,
+        time_steps,
+        meta = "dn",
     )
 
+    for d in devices, t in time_steps
+        name = PSY.get_name(d)
+        expression_up = JuMP.AffExpr(0.0)
+        expression_dn = JuMP.AffExpr(0.0)
+        # TODO: add reserve variables
```
jd-lara, line 294: same here

src/services_models/reserves.jl
```diff
@@ -31,7 +33,15 @@ function service_requirement_constraint!(
...
-    constraint = get_constraint(container, RequirementConstraint(), SR)
+    add_cons_container!(
+        container,
+        RequirementConstraint(),
+        SR,
+        [name],
+        time_steps;
+        meta = name,
+    )
+    constraint = get_constraint(container, RequirementConstraint(), SR, name)
```
jd-lara, line 44: you don't need to retrieve the constraint after creation you can just do `constraint = add_cons_container(...)`

src/services_models/reserves.jl
```diff
@@ -44,13 +54,12 @@ function service_requirement_constraint!(
         container = get_parameter(
             container,
             RequirementTimeSeriesParameter(PSY.Deterministic, "requirement"),
```
jd-lara, line 56: Parameters no longer use this syntax. Are you sure this function is being tested?

src/services_models/reserves.jl
```diff
@@ -44,13 +54,12 @@ function service_requirement_constraint!(
...
-            SR,
+            SR;
+            meta = name,
         )
         param = get_parameter_array(container)
         multiplier = get_multiplier_array(container)
         for t in time_steps
-            param[name, t] = add_parameter(optimization_container.JuMPmodel, ts_vector[t])
-            multiplier[name, t] = requirement
             if use_slacks
```
jd-lara, line 63: The slacks also need to be refactored properly like the network slacks.

src/services_models/reserves.jl
```diff
@@ -63,9 +72,14 @@ function service_requirement_constraint!(
         end
     else
         for t in time_steps
+            if use_slacks
+                resource_expression = sum(reserve_variable[:, t]) + slack_vars[t]
```
jd-lara, line 76: do 
`sum(reserve_variable[:, t] for t in time_periods)` the summation of variables in JuMP is more performant with the index specified.

src/services_models/reserves.jl
```diff
@@ -81,7 +95,15 @@ function service_requirement_constraint!(
...
-    constraint = get_constraint(container, RequirementConstraint(), SR)
+    add_cons_container!(
+        container,
+        RequirementConstraint(),
+        SR,
+        [name],
+        time_steps;
+        meta = name,
+    )
+    constraint = get_constraint(container, RequirementConstraint(), SR, name)
```
jd-lara, line 106: same here, no need to create the container and call it after

src/services_models/reserves.jl
```diff
@@ -122,7 +144,15 @@ function service_requirement_constraint!(
...
```
jd-lara, line 155: also here

src/services_models/services_constructor.jl
```diff
@@ -1,3 +1,32 @@
+function initialize_timeseries_names(
+    ::Type{<:PSY.Reserve},
+    ::Type{T},
+) where {T <: Union{RangeReserve, RampReserve}}
+    return Dict{Type{<:TimeSeriesParameter}, String}(
+        RequirementTimeSeriesParameter => "requirement",
+    )
+end
+
+function initialize_timeseries_names(
```
jd-lara, line 10: do we need these?

src/services_models/services_constructor.jl
```diff
@@ -1,3 +1,32 @@
...
+    ::Type{<:PSY.Service},
+    ::Type{<:AbstractServiceFormulation},
+)
+    return Dict{Type{<:TimeSeriesParameter}, String}()
+end
+
+function initialize_attributes(::Type{<:PSY.Service}, ::Type{<:AbstractServiceFormulation})
```
jd-lara, line 17: and this?

src/services_models/services_constructor.jl
```diff
@@ -15,96 +44,182 @@ end
...
-    function _construct_valid_services!(service_model::ServiceModel)
-        @debug "Building $(get_component_type(service_model)) with $(get_formulation(service_model)) formulation"
-        services = get_component_type(service_model)[]
-        if validate_services!(
-            get_component_type(service_model),
-            services,
-            incompatible_device_types,
+    groupservice = nothing
+
+    for (key, service_model) in services_template
```
jd-lara, line 56: do we need this type check at run time?

src/services_models/services_constructor.jl
```diff
@@ -15,96 +44,182 @@ end
...
+        stage,
+        services_template[groupservice],
+        devices_template,
+        incompatible_device_types,
+    )
+    return
+end
 
-    groupservice = nothing
+function construct_services!(
```
jd-lara, line 82: We should be passing the model into the construct services call now. We don't need the template. I am confused about this function use.

src/devices_models/devices/storage.jl
jd-lara: But in storage we have a formulation for this

src/devices_models/device_constructors/common/constructor_validations.jl
sourabhdalvi: that's a typo should be

src/devices_models/devices/storage.jl
sourabhdalvi: still valid

src/services_models/reserves.jl
sourabhdalvi: that would require to create expression for the reserve balance constraint, and then implement a `add_to_expression` right ?

src/services_models/reserves.jl
sourabhdalvi: here we only want to sum all reserve variable for a single time period but I get your point

src/services_models/services_constructor.jl
sourabhdalvi: I think its only used for the ORDC type reserve, I had to add this since there is no default implementation `initialize_timeseries_names`

src/services_models/services_constructor.jl
sourabhdalvi: This default implementation `initialize_attributes`, again I can put the dict directly in the ServiceModel constructor but that will not allow us to dispatch on different service and formulation types when we starting using attributes

src/services_models/services_constructor.jl
sourabhdalvi: I just followed the existing code logic to build the group reserves last, we could make the `ServicesModelContainer` a OrderedDict to make sure the group reserve is always the last.

src/services_models/services_constructor.jl
sourabhdalvi: the template is used to figure out the incompatible devices, this is existing code logic I followed

src/devices_models/device_constructors/common/constructor_validations.jl
sourabhdalvi: I'll add a TODO here so we can address this later

jd-lara (approved):

PR #750, opened 2021-09-01
emulation model v3

jd-lara:
Opening this PR as a draft first to discuss the move of the formerly called ProblemStore to the model internal.


daniel-thom (commented):
Nice consolidation. I have one ask: consistent use of store_model vs model_store.
model_store_params.jl
StoreModelParams
model_store_init

src/operation/model_store_params.jl
```diff
@@ -0,0 +1,56 @@
...
+    return StoreModelRequirements(
+        Dict{ConstraintKey, Dict{String, Any}}(),
+        Dict{ParameterKey, Dict{String, Any}}(),
+        Dict{VariableKey, Dict{String, Any}}(),
+    )
+end
+
+struct StoreModelParams
+    num_executions::Int
+    horizon::Int
```
jd-lara, line 17: @daniel-thom we could do away with horizon since it is already in the container.

jd-lara: > Nice consolidation. I have one ask: consistent use of store_model vs model_store.
> model_store_params.jl
> StoreModelParams
> model_store_init

Fixed in  2312d91

daniel-thom (approved):

PR #751, opened 2021-09-02
Refactoring service construction 

sourabhdalvi:


src/devices_models/devices/AC_branches.jl
```diff
@@ -39,6 +39,20 @@ get_variable_binary(
 ) = false
 
 get_variable_multiplier(_, ::Type{<:PSY.ACBranch}, _) = NaN
+
+function initialize_timeseries_names(
```
jd-lara, line 43: I don't understand why we have time series here. No model in branches uses time series.

src/core/service_model.jl
```diff
@@ -85,6 +88,13 @@ get_duals(m::ServiceModel) = m.duals
 get_time_series_names(m::ServiceModel) = m.time_series_names
 get_attributes(m::ServiceModel) = m.attributes
 get_attribute(m::ServiceModel, key::String) = get(m.attributes, key, nothing)
+get_contributing_devices_map(m::ServiceModel) = m.contributing_devices_map
+get_contributing_devices_map(m::ServiceModel, key) =
+    get(m.contributing_devices_map, key, nothing)
+get_contributing_devices(m::ServiceModel) =
```
jd-lara, line 94: I don't understand this function. Are you trying to get all the contributing devices?

src/devices_models/devices/electric_loads.jl
```diff
@@ -33,6 +33,23 @@ get_multiplier_value(::TimeSeriesParameter, d::PSY.ElectricLoad, ::AbstractContr
 
 #! format: on
 
+function initialize_timeseries_names(
```
jd-lara, line 36: @daniel-thom what do you think about these names to set-up the default values of the time series and attributes of the models. Should we re-name this methods to something more clear?

src/services_models/services_constructor.jl
```diff
@@ -41,6 +32,54 @@ function get_incompatible_devices(devices_template::Dict)
...
+    types = unique(typeof.(contributing_devices))
+    device_list = Dict()
+    for t in types
+        _devices = filter(x -> typeof(x) == t, contributing_devices)
+        device_list[t] = _devices
+    end
+    return device_list
+end
+
+function populate_contributing_devices!(
```
jd-lara, line 45: this function might need some additional care. I don't fully understand what it is doing.

src/services_models/services_constructor.jl
```diff
@@ -41,6 +32,54 @@ function get_incompatible_devices(devices_template::Dict)
...
+    container::OptimizationContainer,
```
jd-lara, line 46: where are we using the container here?

src/devices_models/devices/electric_loads.jl
daniel-thom: I would probably call it `get_time_series_name_definitions`.

Note that we use `time_series` (two words).

jd-lara: How about `get_time_series_default_names`

src/devices_models/devices/electric_loads.jl
```diff
@@ -33,6 +33,23 @@ get_multiplier_value(::TimeSeriesParameter, d::PSY.ElectricLoad, ::AbstractContr
...
+    ::Type{<:PSY.ElectricLoad},
+    ::Type{<:Union{FixedOutput, AbstractLoadFormulation}},
+)
+    return Dict{Type{<:TimeSeriesParameter}, String}(
+        ActivePowerTimeSeriesParameter => "max_active_power",
+        ReactivePowerTimeSeriesParameter => "max_active_power",
+    )
+end
+
+function initialize_attributes(
```
jd-lara, line 46: `get_default_attributes`

src/core/service_model.jl
jd-lara: Iterators.Flatten(values(m.contributing_devices_map))

src/services_models/services_constructor.jl
```diff
@@ -41,6 +32,54 @@ function get_incompatible_devices(devices_template::Dict)
     return incompatible_device_types
 end
 
+function sort_contributing_devices(contributing_devices)
```
jd-lara, line 35: `map_by_type_contributing_devices`

src/core/service_model.jl
sourabhdalvi: The downstream implementation of add_variables only take a `Vector{PSY.Component}` or  `IS.FlattenIteratorWrapper{PSY.Component}`. So leaving it as is for now

jd-lara: Can you leave a comment in the code please

src/devices_models/devices/thermal_generation.jl
```diff
@@ -28,6 +28,7 @@ get_variable_initial_value(::ActivePowerVariable, d::PSY.ThermalGen, ::AbstractT
 get_variable_initial_value(::PowerAboveMinimumVariable, d::PSY.ThermalGen, ::AbstractCompactUnitCommitment) = max(0.0, PSY.get_active_power(d) - PSY.get_active_power_limits(d).min)
 
 get_variable_lower_bound(::ActivePowerVariable, d::PSY.ThermalGen, ::AbstractThermalFormulation) = PSY.get_active_power_limits(d).min
+get_variable_lower_bound(::ActivePowerVariable, d::PSY.ThermalGen, ::AbstractThermalUnitCommitment) = 0.0
```
sourabhdalvi, line 31: So in the previous implementation of range constraints, we would grab the variable and remove its bounds if we added a binary variable based lower bound. Now with the possibility of having either variables or expressions, its better to set the bound to 0.0 here and changing it later

src/devices_models/devices/hydro_generation.jl
```diff
@@ -19,6 +19,7 @@ get_variable_binary(::ActivePowerVariable, ::Type{<:PSY.HydroGen}, ::AbstractHyd
 get_variable_initial_value(::ActivePowerVariable, d::PSY.HydroGen, ::AbstractHydroFormulation) = PSY.get_active_power(d)
 
 get_variable_lower_bound(::ActivePowerVariable, d::PSY.HydroGen, ::AbstractHydroFormulation) = PSY.get_active_power_limits(d).min
+get_variable_lower_bound(::ActivePowerVariable, d::PSY.HydroGen, ::AbstractHydroUnitCommitment) = 0.0
```
sourabhdalvi, line 22: same thing as thermal commitment range constraints

jd-lara: ok

src/devices_models/devices/thermal_generation.jl
jd-lara: ok

jd-lara:
Just tests locally and all tests passed

PR #753, opened 2021-09-04
Refactoring to use expressions for adding reserve variables to range constraints

sourabhdalvi:
Still need to figure out the correct reserve constraint for HydroPumpedStorage devices (probably something similar to BatteryAncillaryServices)


src/core/optimization_container.jl
```diff
@@ -881,9 +880,9 @@ function _add_expression_container!(
     sparse = false,
 )
     if sparse
-        expr_container = sparse_container_spec(JuMP.AbstractJuMPScalar, axs...)
+        expr_container = sparse_container_spec(JuMP.GenericAffExpr, axs...)
```
jd-lara, line 883: You need to be careful with this change since it can break when we need these to store `PGAE`. If you are unsure, let's just put Any or something super generic. This will be problematic until we can use `ParametricOptInterface`

src/devices_models/device_constructors/hydrogeneration_constructor.jl
```diff
@@ -105,6 +105,30 @@ function construct_device!(
     )
 
     add_parameters!(container, ActivePowerTimeSeriesParameter, devices, model)
+
+    if has_service_model(model)
+        time_steps = get_time_steps(container)
+        names = PSY.get_name.(devices)
+        for meta in [LOWER_BOUND, UPPER_BOUND]
+            add_expression_container!(
```
jd-lara, line 113: Why are you adding the container here? We have to use `add_to_expression!` methods to handle this. The addition of the container is done inside of the `add_to_expression!` call.

src/devices_models/device_constructors/hydrogeneration_constructor.jl
```diff
@@ -121,10 +145,15 @@ function construct_device!(
     devices = get_available_components(H, sys)
 
     # Constraints
+    if has_service_model(model)
+        variable = ActivePowerRangeExpression
```
jd-lara, line 149: we should rename this from variable to something more generic.

jd-lara (changes requested):
The patter of using 

```
         names = PSY.get_name.(devices)
         for meta in [LOWER_BOUND, UPPER_BOUND]
             add_expression_container!(
                 container,
                 ActivePowerRangeExpression(),
                 H,
                 names,
                 time_steps;
                 meta = meta,
             )
```

in the constructor calls is incorrect.

This should be handled at the `add_to_expression!` level.

src/core/optimization_container.jl
sourabhdalvi: I think the super generic type that would work will be `JuMP.AbstractJuMPScalar`, So I'll leave as is

jd-lara (changes requested):
We shouldn't use the meta field in the device constructors.

jd-lara (approved):

PR #754, opened 2021-09-07
Add support for EmulationModel results

daniel-thom:


src/PowerSimulations.jl
```diff
@@ -146,8 +146,13 @@ export get_problem_results
 export get_system
 export get_system!
 export set_system!
+export list_dual_keys
```
daniel-thom, line 149: I added exports of `list.*keys` because of their potential use in PowerGraphics. But, I'm not convinced that they are required.

src/core/auxiliary_variables.jl
```diff
@@ -1,3 +1,4 @@
+# TODO DT: Should tis be called AuxVariableKey?
```
daniel-thom, line 1: Note this.

src/operation/emulation_model.jl
```diff
@@ -221,7 +221,9 @@ function model_store_params_init!(model::EmulationModel)
 end
 
 function model_store_init!(model::EmulationModel)
-    initialize_model_storage!(
+    # TODO DT: style of function names: verb_noun or noun_verb?
```
daniel-thom, line 224: Potential naming consistency issue

src/operation/emulation_model.jl
```diff
@@ -235,28 +237,18 @@ function build_pre_step!(model::EmulationModel)
             @info "EmulationProblem status not BuildStatus.EMPTY. Resetting"
             reset!(model)
         end
-        # Initial time are set here because the information is specified in the
```
daniel-thom, line 238: I moved this code to the section where we perform other container initialization.

src/operation/emulation_model.jl
```diff
@@ -235,28 +237,18 @@ function build_pre_step!(model::EmulationModel)
...
-        get_optimization_container(model).built_for_recurrent_solves = true
-        @info "Initializing ModelStoreParams"
-        model_store_params_init!(model)
         set_status!(model, BuildStatus.IN_PROGRESS)
     end
     return
 end
 
+# TODO DT: should this be called build_impl!
+# Note that run! calls run_impl!
```
daniel-thom, line 246: Potential naming consistency issue

src/operation/emulation_model.jl
```diff
@@ -235,28 +237,18 @@ function build_pre_step!(model::EmulationModel)
...
 function _build!(model::EmulationModel{<:EmulationProblem}, serialize::Bool)
     TimerOutputs.@timeit BUILD_PROBLEMS_TIMER "Problem $(get_name(model))" begin
         try
             build_pre_step!(model)
+            # TODO DT: Why is this called problem build?
```
daniel-thom, line 251: Potential naming consistency issue

src/operation/emulation_model.jl
```diff
@@ -463,173 +453,109 @@ Default solve method for an Emulation model used inside of a Simulation. Solves
...
-    #             file_type = exports[:file_type]
-    #             df = axis_array_to_dataframe(variable)
-    #             time_col = range(timestamp, length = horizon, step = resolution)
-    #             DataFrames.insertcols!(df, 1, :DateTime => time_col)
-    #             export_result(file_type, exports_path, key, timestamp, df)
-    #         end
-    #     end
+function _write_model_aux_variable_results!(store, container, execution)
+    for (key, variable) in get_aux_variables(container)
+        write_result!(store, STORE_CONTAINER_AUX_VARIABLES, key, execution, variable)
```
daniel-thom, line 557: Can you confirm that aux_variables will always be stored independently of variables? We have some inconsistency.

src/operation/in_memory_model_store.jl
```diff
@@ -112,22 +97,26 @@ Base.flush(store::InMemoryModelStore) = nothing
...
         store_container = getfield(store.data, type)
-        for (key, container) in field_containers
-            container_axes = axes(container)
+        for (key, field_container) in field_containers
+            container_axes = axes(field_container)
             @debug "Adding $(encode_key_as_string(key)) to InMemoryModelStore"
             if length(container_axes) == 2
-                column_names = string.(axes(container)[1])
+                if type == STORE_CONTAINER_PARAMETERS
+                    column_names = string.(get_parameter_array(field_container).axes[1])
```
daniel-thom, line 116: Let me know if there is a better way to handle this.

src/operation/in_memory_model_store.jl
```diff
@@ -145,37 +134,41 @@ function initialize_model_storage!(
...
+    execution,
     array,
     columns = nothing,
 )
-    container = getfield(store.data[model_name], container_type)
-    container[name][timestamp] = axis_array_to_dataframe(array, columns)
+    container = getfield(store.data, container_type)
+    df = axis_array_to_dataframe(array, columns)
+
+    # TODO DT: PERF: names are not in the same order.
```
daniel-thom, line 153: Is there a way to ensure that we get the names in the same order at initialization and every call here?

src/operation/problem_results.jl
```diff
@@ -35,7 +45,7 @@ function ProblemResults(model::DecisionModel)
     duals = read_duals(container)
     parameters = read_parameters(container)
     timestamps = get_timestamps(model)
-    optimizer_stats = OptimizerStats(model)
+    optimizer_stats = to_dataframe(OptimizerStats(model))
```
daniel-thom, line 48: In order to have type consistency this struct built from a DecisionModel will have a DataFrame with one row. If it's built from an EmulationModel then it will have one row per execution. We might need to add something else to the struct to indicate how it was built.

test/test_model_emulation.jl
```diff
@@ -1,3 +1,5 @@
+import DataFrames: DataFrame
```
daniel-thom, line 1: Maybe we should do this in runtests.jl

src/PowerSimulations.jl
jd-lara: These would be useful also for exploration, we should keep them in the meantime.

src/core/auxiliary_variables.jl
jd-lara: Yes, noted for future massive renames

src/operation/emulation_model.jl
jd-lara: I have no preference, but yeah needs to be made consistent

src/operation/emulation_model.jl
jd-lara: can you do the rename please?

src/operation/emulation_model.jl
jd-lara: we need 2 methods generally speaking whether the build is called on its own or part of the simulation construction step. The reason for this was the ability to capture in the same logger the processes. We can review if needed.

src/operation/emulation_model.jl
```diff
@@ -392,14 +395,17 @@ function run_impl(
...
+
     try
         prog_bar =
             ProgressMeter.Progress(internal.executions; enabled = enable_progress_bar)
         for execution in 1:(internal.executions)
-            # timed_log = get_solve_timed_log(model)
-            # _, timed_log[:timed_solve_time], timed_log[:solve_bytes_alloc], timed_log[:sec_in_gc] = @timed
-            one_step_solve!(model)
-            #write_results(model)
+            timed_log = get_solve_timed_log(model)
```
jd-lara, line 403: we probably need to make this consistent later with the decision model since the timer now captures the solver's time and also the `aux_variable` and duals `calculation` we can extract the "pure" solver time using JuMP.

src/operation/emulation_model.jl
```diff
@@ -463,173 +453,109 @@ Default solve method for an Emulation model used inside of a Simulation. Solves
...
     end
 
     return solve_status
 end
 
-function write_results!(model::EmulationModel)
-    # substitute with getters
-    store = model.store
-    container = get_container(model)
+list_aux_variable_keys(x::EmulationModel) =
```
jd-lara, line 473: the list methods should be moved to `operation_model_interface` since they should potentially work on DecisionModels too

src/operation/emulation_model.jl
jd-lara: This is something we need to decide on. I prefer them to be stored separately for clarity

src/operation/emulation_model.jl
```diff
@@ -463,173 +453,109 @@ Default solve method for an Emulation model used inside of a Simulation. Solves
...
-    #             export_result(file_type, exports_path, key, timestamp, df)
-    #         end
-    #     end
+function _write_model_aux_variable_results!(store, container, execution)
+    for (key, variable) in get_aux_variables(container)
+        write_result!(store, STORE_CONTAINER_AUX_VARIABLES, key, execution, variable)
+    end
 end
+
+read_optimizer_stats(model::EmulationModel) = read_optimizer_stats(get_store(model))
```
jd-lara, line 561: should also be in the interface file.

src/operation/in_memory_model_store.jl
jd-lara: I don't have a better idea at the moment except enforcing that the axes are always strings.

src/operation/in_memory_model_store.jl
jd-lara: I think the names of data frames are always in the same order.

src/operation/problem_results.jl
jd-lara: We could add something to problem results to indicate that it was a DecisionModel or en EmulationModel.

src/simulation/hdf_simulation_store.jl
```diff
@@ -28,6 +28,7 @@ mutable struct ProblemDatasets
     duals::Dict{ConstraintKey, Dataset}
     parameters::Dict{ParameterKey, Dataset}
     variables::Dict{VariableKey, Dataset}
+    # TODO v015: aux_variables?
```
jd-lara, line 31: We should discuss this to weigh the cost vs benefit

src/operation/emulation_model.jl
daniel-thom: I will make it verb_noun

src/operation/emulation_model.jl
daniel-thom: Change to `build_problem!`

src/operation/problem_results.jl
daniel-thom: Record the type as a string in the struct.

test/test_model_emulation.jl
daniel-thom: Move to runtests.jl

src/operation/emulation_model.jl
```diff
@@ -356,19 +355,22 @@ function calculate_dual_variables!(model::EmulationModel)
...
     model_status = JuMP.primal_status(jump_model)
     if model_status != MOI.FEASIBLE_POINT::MOI.ResultStatusCode
         error("Optimizer returned $model_status")
-    else
-        calculate_aux_variables!(model)
-        calculate_dual_variables!(model)
     end
+
+    _, log[:timed_calculate_aux_variables] = @timed calculate_aux_variables!(model)
+    _, log[:timed_calculate_dual_variables] = @timed calculate_dual_variables!(model)
```
daniel-thom, line 372: I added these, but I don't think anything is logging them.

src/operation/operation_model_interface.jl
```diff
@@ -110,3 +106,38 @@ function _pre_solve_model_checks(model::OperationModel, optimizer)
...
+    return read_results(get_store(model), STORE_CONTAINER_PARAMETERS, key)
+end
+
+function read_aux_variable(model::OperationModel, key::AuxVarKey)
+    return read_results(get_store(model), STORE_CONTAINER_AUX_VARIABLES, key)
+end
+
+function read_variable(model::OperationModel, key::VariableKey)
+    return read_results(get_store(model), STORE_CONTAINER_VARIABLES, key)
+end
```
daniel-thom, line 141: Note that I moved these in addition the ones you requested.

src/operation/emulation_model.jl
jd-lara: For these tests, there aren't any. We need to extend the test cases to account for these.

jd-lara (approved):

PR #755, opened 2021-09-14
Check for invalid components at build time

daniel-thom:
This is dependent on https://github.com/NREL-SIIP/PowerSystems.jl/pull/811 and so CI will fail until there is a new PSY tag.


jd-lara:
@daniel-thom after PSY 1.12 is released can you please rebase this branch to trigger the tests?

PR #757, opened 2021-09-16
Refactor OperationProblem results handling

daniel-thom:
1. DecisionModel and EmulationModel can be deserialized and re-run as long as the optimizer and JuMP model are provided.
2. ProblemResults can be constructed from either DecisionModel or EmulationModel.
3. ProblemResults can be serialized in binary form. The PSY.System can be restored upon deserialization if the file is stored in the same location as the serialized model.
4. Change the solve!/run! functions to support automatic build, results serialization, and results export.
5. Consolidated some common functions between DecisionModel and EmulationModel.

### DIFF
--- a/src/utils/dataframes_utils.jl
+++ b/src/utils/dataframes_utils.jl
@@ -75,7 +75,11 @@ function axis_array_to_dataframe(
             result[t] = _jump_value(input_array[t])
         end
 
-        @assert columns !== nothing
+        # TODO v015: This is a hack to workaround lack of support for this scenario.
+        # @assert columns !== nothing
+        if columns === nothing
+            columns = ["System"]
+        end
         return DataFrames.DataFrame(columns[1] => result)
     elseif length(axes(input_array)) == 2
         result = Array{Float64, length(input_array.axes)}(
--- a/src/PowerSimulations.jl
+++ b/src/PowerSimulations.jl
@@ -113,6 +113,7 @@ export build!
 ## Op Model Exports
 export get_initial_conditions
 export serialize_problem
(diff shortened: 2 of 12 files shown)


daniel-thom:
I noticed that neither `solve!` nor `run!` enable logging. So, we don't capture any log events or timing stats in a log file. We also don't record the optimizer that was used. Should we enable this? We would need to do it in a way that allows for the simulation to control logging when the models are run that way.

daniel-thom:
@jd-lara The most recent change to force serialization of ProblemResults for DecisionProblem causes a failure on line 123 of test_model_decision.jl because the columns passed to axis_array_to_dataframe is nothing. Do you know what is wrong?

daniel-thom:
> I noticed that neither `solve!` nor `run!` enable logging. So, we don't capture any log events or timing stats in a log file. We also don't record the optimizer that was used. Should we enable this? We would need to do it in a way that allows for the simulation to control logging when the models are run that way.

I ran out of time to implement this. We'll need to implement it later.

src/utils/dataframes_utils.jl
```diff
@@ -75,7 +75,11 @@ function axis_array_to_dataframe(
             result[t] = _jump_value(input_array[t])
         end
 
-        @assert columns !== nothing
+        # TODO v015: This is a hack to workaround lack of support for this scenario.
```
jd-lara, line 78: noted. we need a better fix to this.

jd-lara (approved):
This looks great, we might need to refine some details later with the addition of the initialization.

PR #758, opened 2021-09-16
Bug Fix in ramp constraint

sourabhdalvi:


jd-lara:
Do we need to propagate to v015?

jd-lara (approved):

sourabhdalvi:
> Do we need to propagate to v015?

Yes, I'll include that in my non-spinning reserve PR

jd-lara:
Can you add a test to check this please?

jd-lara:
@sourabhdalvi we need to study the tests failures before merging. Please a look carefully for this fixes.

sourabhdalvi:
> Can you add a test to check this please?

So adding the start variable for the 1st period was causing the ramping constraint to be relaxed (some units turn on in the 1st period) and removing that cause it to be infeasible in ED (not enough power in balance constraint) but if I remove the IntegralLimitFF which was enforcing (sum(p_h <= 0.0) in the same period it works

test/test_simulation_execute.jl
```diff
@@ -129,10 +129,6 @@ end
                 binary_source_problem = PSI.ON,
                 affected_variables = [PSI.ACTIVE_POWER],
             ),
-            ("ED", :devices, :HydroEnergyReservoir) => IntegralLimitFF(
```
jd-lara, line 132: Why are we removing these? I don't understand.

sourabhdalvi: Like I said the ramp was relaxed so the problem was feasible, with that enforced the problem becomes infeasible in the 2nd period, because IntegralLimitFF is forcing hydro output  to be zero in the 2nd period and you don't have enough to meet the balance constraint.

sourabhdalvi: Just realized I can add balance slacks to make the test pass,

jd-lara: or better, you can adjust the data so it isn't infeasible and still does the correct thing. The solution of removing test coverage to make the tests pass isn't a good idea.

sourabhdalvi: ok, I'll change the data

PR #759, opened 2021-09-17
Adding Non-spinning reserves

sourabhdalvi:


src/devices_models/devices/common/add_variable.jl
```diff
@@ -19,7 +19,11 @@ function add_variables!(
     service::U,
     contributing_devices::Union{Vector{V}, IS.FlattenIteratorWrapper{V}},
     formulation::AbstractReservesFormulation,
-) where {T <: VariableType, U <: PSY.Reserve, V <: PSY.Component}
+) where {
+    T <: VariableType,
+    U <: Union{PSY.Reserve, PSY.ReserveNonSpinning},
```
jd-lara, line 24: This shouldn't be necessary. Isn't `PSY.ReserveNonSpinning <: PSY.Reserve`

src/services_models/reserves.jl
```diff
@@ -150,23 +166,21 @@ function cost_function!(
...
+    contributing_devices::U,
+    model::ServiceModel{SR, StepwiseCostReserve},
+) where {
+    SR <: PSY.ReserveDemandCurve,
+    U <: Union{Vector{D}, IS.FlattenIteratorWrapper{D}},
+} where {D <: PSY.Component}
     initial_time = get_initial_time(container)
     @debug initial_time
     time_steps = get_time_steps(container)
     name = PSY.get_name(service)
```
jd-lara, line 182:
```suggestion
   names = [PSY.get_name(s) for s in services]
```

Please avoid patterns like creating single element arrays like in line 183. If a user uses a multi-area model this will fail.

test/test_services_constructor.jl
```diff
@@ -96,6 +96,23 @@ end
     moi_tests(model, false, 1008, 0, 480, 216, 192, true)
 end
 
+@testset "Test Reserves from Thermal Standard UC with NonSpinningReserve" begin
```
jd-lara, line 99: Please add a test in the `DecisionModel` test file that solves a problem with non-spinning reserves and also checks for the values.

src/devices_models/devices/common/add_variable.jl
sourabhdalvi: Nope, both are sub types of `<: Service`

jd-lara: Then we need to fix PSY first

src/services_models/reserves.jl
sourabhdalvi: I'm not sure if we actually support multi-area models? Here we just have one service so the change you recommend will not work unless I do make a single element array.
`names = [PSY.get_name(s) for s in [service]]`

jd-lara (approved):

PR #760, opened 2021-09-20
Add logging of run/solve functions

daniel-thom:


src/core/definitions.jl
```diff
@@ -55,7 +56,7 @@ const LOWER_BOUND = "lb"
 
 # File Names definitions
 const PROBLEM_SERIALIZATION_FILENAME = "operation_problem.bin"
-const PROBLEM_BUILD_LOG_FILENAME = "operation_problem_build.log"
```
daniel-thom, line 58: The log now contains build and solve activities.

src/operation/operation_model_interface.jl
```diff
@@ -109,23 +109,31 @@ function build_impl!(model::OperationModel, serialize::Bool)
...
+end
+
+function _pre_solve_model_checks(model::OperationModel, optimizer)
     jump_model = get_jump_model(model)
     if optimizer !== nothing
         JuMP.set_optimizer(jump_model, optimizer)
     end
+
+    optimizer_name = JuMP.solver_name(jump_model)
+    @info "Solving $(typeof(model)) with optimizer = $optimizer_name"
```
daniel-thom, line 135: We could potentially call `JuMP.backend(jump_model)` to capture more information.

jd-lara: Yes, this is correct. We could add that info this PR to capture solver version also.

jd-lara (approved):
This looks great. I think we should add tests in the decision model and emulation model test files to make sure we capture the solver properties correctly. I will approve and leave the testing as optional if @daniel-thom your think that's better.

test/test_model_emulation.jl
```diff
@@ -132,4 +132,22 @@ end
...
+    # Deserialize with different optimizer attributes.
+    #optimizer = JuMP.optimizer_with_attributes(
+    #    OSQP.Optimizer,
+    #    "verbose" => true,
+    #    "max_iter" => 50000,
+    #)
+    #@test_logs (:warn, r"Original solver used attribute") match_mode = :any EmulationModel(
+    #    path,
+    #    optimizer,
+    #)
```
daniel-thom, line 146: This requested use case doesn't work. PSI code creates optimizer attributes outside of JuMP, not within it. I don't see a way to list the custom solver attributes that have been set (from JuMP). If you guys know of a way to do that, I could check them. Otherwise, I'm stuck.

jd-lara: @daniel-thom you are correct, see this: 

image

daniel-thom: My takeaway is that I should delete the code related to optimizer attributes and only log the name. Yours?

jd-lara: We can still capture the attributes but only if the optimizer passed is an `optimizer_with_attributes` object. Otherwise we can't. We can do something like: 

if no attributes -> assume that the solver config was the default and otherwise capture the attributes.

jd-lara: @daniel-thom we keep a copy of the optimizer in the settings: 

```Julia
struct Settings
    horizon::Base.RefValue{Int}
    time_series_cache_size::Int
    warm_start::Base.RefValue{Bool}
    initial_time::Base.RefValue{Dates.DateTime}
    optimizer::Union{Nothing, MOI.OptimizerWithAttributes}
    direct_mode_optimizer::Bool
    optimizer_log_print::Bool
    system_to_file::Bool
    export_pwl_vars::Bool
    allow_fails::Bool
    ext::Dict{String, Any}
end
```

On capturing the version of the solver we are kind of out of luck since MOI doesn't have a solver version attribute like a name that we can directly use. It shouldn't be too hard to contribute that into the JuMP-dev ecosystem since many wrappers (most of the ones I checked) support a call to return the solver version. 

if a user uses a custom configuration, those parameters will be captured there. We can just do 

```
if !isempty(settings.optimizer.params)
    atttributes = deepcopy(settings.optimizer.params)
else
   attributes = "Solver used with default parameters" 
end
```

daniel-thom: 1. Isn't that a private attribute subject to change? Shouldn't we only use public/exported methods?
2. params is defined as `params::Vector{Pair{AbstractOptimizerAttribute,Any}}`. Is this guaranteed to be serializable with the Serialization library?
3. I didn't realize that we save a copy of the optimizer in Settings. The user can change it in the call to `run!` or `solve!` but the value in Settings is not updated. Is that a bug? Does this mean that we should detect a change in the optimizer in `run!` and then re-serialize the problem?

src/operation/decision_model.jl
```diff
@@ -351,14 +351,49 @@ results = solve!(OpModel, output_dir = "output")
...
+                status = solve_impl(model; kwargs...)
+                set_run_status!(model, status)
+            end
+            if status == RunStatus.SUCCESSFUL
+                if serialize
+                    TimerOutputs.@timeit RUN_OPERATION_MODEL_TIMER "Serialize" begin
+                        optimizer = get(kwargs, :optimizer, nothing)
+                        serialize_problem(model, optimizer = optimizer)
+                        serialize_optimization_model(model)
+                    end
```
daniel-thom, line 385: @jd-lara @claytonpbarrows Note this change. This is the best way to ensure that we serialize the correct optimizer attributes.

And it only runs when the solve is successful. Is that correct? (Same applies for EmulationModel)

PR #762, opened 2021-09-22
Cb/branch constraints

claytonpbarrows:


src/devices_models/devices/DC_branches.jl
```diff
@@ -76,52 +77,52 @@ function branch_rate_constraints!(
...
+        )
+        max_rate = min(
+            PSY.get_active_power_limits_from(d).max,
+            PSY.get_active_power_limits_to(d).max,
+        )
+        constraint[PSY.get_name(d), t] = JuMP.@constraint(
+            container.JuMPmodel,
+            min_rate <= var[PSY.get_name(d), t] <= max_rate
+        )
+        # Needs refactoring. This add to expression model doesn't work anymore
```
jd-lara, line 117: To do this correctly you need to follow a pattern similar to the one in Services. @sourabhdalvi can probably help to get this done in the correct way.

jd-lara (approved):
Looks like a great improvement!

PR #763, opened 2021-09-24
Propagating initialization changes

sourabhdalvi:


src/core/settings.jl
```diff
@@ -1,4 +1,4 @@
-struct Settings
+mutable struct Settings
```
jd-lara, line 1: Why is this mutable now? This shouldn't have switched to mutable.

src/devices_models/devices/common/add_to_expression.jl
```diff
@@ -15,16 +15,34 @@ function add_expressions!(
     return
 end
 
+function add_expressions!(
+    container::OptimizationContainer,
+    ::Type{T},
+    devices::U,
+    model::ServiceModel{V, W};
+    meta = CONTAINER_KEY_EMPTY_META,
```
jd-lara, line 23: You shouldn't need to pass meta here. The meta field should be determined inside of the function calling `add_expression_container!`

src/devices_models/devices/thermal_generation.jl
```diff
@@ -301,6 +349,7 @@ function add_constraints!(
         names,
         time_steps,
         meta = "off",
+        sparse = true,
```
jd-lara, line 352: Why is this a sparse container?

src/devices_models/devices/thermal_generation.jl
```diff
@@ -325,6 +374,10 @@ function add_constraints!(
             (limits.max - limits.min) * varstatus[name, t] -
             max(limits.max - startup_shutdown_limits.startup, 0) * varon[name, t]
         )
+
+        exp_lb = JuMP.AffExpr(0.0, varp[name, t] => 1.0)
+        con_lb[name, t] = JuMP.@constraint(container.JuMPmodel, exp_lb >= 0)
```
jd-lara, line 379: Use floats to avoid type conversions please.

src/initial_conditions/initialization.jl
```diff
@@ -23,7 +23,7 @@ end
 function build_initialization_problem!(model::T) where {T <: OperationModel}
     model.internal.ic_model_container = deepcopy(get_optimization_container(model))
     ic_settings = model.internal.ic_model_container.settings
-    set_horizon!(ic_settings, 1)
+    # TODO: add interface to allow user to change the horizon
```
jd-lara, line 26:
```suggestion
    # TODO: add an interface to allow user to configure initialization problem
```

src/initial_conditions/update_initial_conditions.jl
```diff
@@ -50,6 +50,58 @@ function update_initial_conditions(
...
+        set_ic_quantity!(ic, values[index, get_component_name(ic)])
+    end
+    return
+end
+
+function update_initial_conditions(
+    ic_vector::Vector{T},
+    store::InMemoryModelStore,
+    ::Dates.Period,
+) where {T <: InitialCondition{InitialEnergyLevelDown, PJ.ParameterRef}}
```
jd-lara, line 96: All of these constraints should be able to take `InitialConditions` with `Float64` or `ParameterRef` there seems to be something off here.

src/operation/operation_model_interface.jl
```diff
@@ -145,6 +145,11 @@ function build_impl!(model::OperationModel)
 end
 
 function build_if_not_already_built!(model; kwargs...)
```
jd-lara, line 147: @daniel-thom here is where we need to fix the flow to handle the initialization solver set-up.

jd-lara (changes requested):
We need to resolve the build requirements before merging or at least making a plan to handle it on a separate PR.

src/devices_models/devices/thermal_generation.jl
sourabhdalvi: that constraint is not created for the last time step, so the container had a undef. I should have changed the axes to be `time steps[1:end-1]` then we can avoid the usage of spare. sorry my bad.

src/initial_conditions/update_initial_conditions.jl
sourabhdalvi: I'll make it `Union{Float64, PJ.ParameterRef}`

src/operation/model_numerical_analysis_utils.jl
```diff
@@ -106,10 +106,19 @@ function get_constraint_numerical_bounds(model::OperationModel)
     end
     bounds = ConstraintBounds()
     for (const_key, constriant_array) in get_constraints(get_optimization_container(model))
-        for idx in Iterators.product(constriant_array.axes...)
-            con_obj = JuMP.constraint_object(constriant_array[idx...])
-            update_coefficient_bounds(bounds, con_obj, (const_key, idx))
-            update_rhs_bounds(bounds, con_obj, (const_key, idx))
+        if typeof(constriant_array) <: JuMP.Containers.SparseAxisArray
```
jd-lara, line 109: 1. Add a TODO to handle this at compile and not at run time
2. 1
```suggestion
        if isa(constriant_array, JuMP.Containers.SparseAxisArray)
```

jd-lara (approved):
I added a few fixes needed. Please address them and merge when the tests pass.

PR #764, opened 2021-09-29
Add recorder for OperationModel

daniel-thom:


src/initial_conditions/update_initial_conditions.jl
```diff
@@ -1,4 +1,4 @@
-function update_initial_conditions(
+function update_initial_conditions!(
```
daniel-thom, line 1: Not sure if no `!` was intentional, but it felt wrong to me. Tell me if I should revert these changes.

src/initial_conditions/update_initial_conditions.jl
```diff
@@ -130,9 +130,16 @@ function update_initial_conditions!(
...
     interval = get_interval(model.internal.store_parameters)
     ini_conditions_vector = get_initial_condition(container, key)
-    update_initial_conditions(ini_conditions_vector, model.store, interval)
+    timestamp = get_initial_time(model) + get_execution_count(model) * get_resolution(model)
+    update_initial_conditions!(ini_conditions_vector, model.store, interval)
+    for (i, initial_condition) in enumerate(ini_conditions_vector)
+        IS.@record :simulation InitialConditionUpdateEvent(
+            timestamp,
+            initial_condition,
+            1,  # TODO DT
```
daniel-thom, line 142: We probably need this because the event is shared with simulations.

src/utils/recorder_events.jl
```diff
@@ -59,24 +59,62 @@ end
...
+    common::IS.RecorderEventCommon
+    category::String
+    simulation_time::Dates.DateTime
+    parameter_type::String
+    device_name::String
+    previous_value::Float64
+    val::Float64
+    problem_number::Int
+    source::Int
+end
```
daniel-thom, line 95: This is not correct and TBD. Let me know what fields should be recorded for Parameter Updates. The implementation is as of now unclear.

src/operation/decision_model.jl
```diff
@@ -480,7 +488,7 @@ function _write_model_dual_results!(
             key,
             timestamp,
             constraint,
-            [encode_key(key)],  # TODO DT: this doesn't seem right
+            [encode_key(key)],  # TODO DT: is this what the columns should be?
```
daniel-thom, line 491: This code is only called in the Simulation execution path. The original code used the name (Symbol) of the dual. Is this going to be correct?

src/utils/recorder_events.jl
```diff
@@ -48,35 +48,59 @@ end
...
         problem_number,
     )
 end
 
+struct ParameterUpdateEvent <: IS.AbstractRecorderEvent
+    common::IS.RecorderEventCommon
+    execution_timestamp::Dates.DateTime
+    parameter_type::String
+    device_type::String
+    name::String
```
daniel-thom, line 83: This field name may need to change. It is only valid if all ParameterAttributes subtypes will define `name`. We could call it something generic like `meta` or `tag`.

src/operation/decision_model.jl
jd-lara: It shouldn't be needed anymore since the duals are not related to the constraints that have their own column names. I think we need to test this in an emulation test that asks for duals.

src/utils/recorder_events.jl
jd-lara: `timeseries_name`?

jd-lara (approved):

src/operation/decision_model.jl
daniel-thom: If you're saying that that might expose a problem in the EmulationModel/InMemoryModelStore code, I agree.

src/utils/recorder_events.jl
daniel-thom: Note that I changed the field from `name` to `tag`. The base recorder event already uses `name` for the event type.

Using `time_series_name` implies that the event is only for TimeSeriesParameters, but that isn't right case, right? Won't we soon add events for other parameter types? I would be OK changing this event to be TimeSeriesParameterUpdateEvent. A consequence of doing that is that we would not be able to print all parameter update events in the same table.

PR #765, opened 2021-10-01
Minor Changes

sourabhdalvi:
- Changing formulation names 
    - ThermalDispatch -> ThermalBasicDispatch,  
    - ThermalRampLimited -> ThermalStandardDispatch
 - Bug fix in cost function
 - Fixing typos in Renewable and Storage 
- [ ]  Fixing issue when service name contain an underscore

### DIFF
--- a/docs/src/get_test_data.jl
+++ b/docs/src/get_test_data.jl
@@ -16,7 +16,7 @@ system = build_c_sys5_re(; add_reserves = true)
 solver = optimizer_with_attributes(Cbc.Optimizer)
 
 devices = Dict{Symbol, DeviceModel}(
-    :Generators => DeviceModel(ThermalStandard, ThermalDispatch),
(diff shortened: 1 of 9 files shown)


jd-lara (approved):

src/core/optimization_container.jl
```diff
@@ -626,6 +626,7 @@ function get_variable(
     ::Type{U},
     meta::String = CONTAINER_KEY_EMPTY_META,
 ) where {T <: VariableType, U <: Union{PSY.Component, PSY.System}}
+    meta = remove_delimiter!(meta)
```
jd-lara, line 629: The call to this function should be when we pass the name to the `add_container` call, not here.

sourabhdalvi: yup, I should have caught that

PR #772, opened 2021-10-20
Bug fix in compact thermal formulations

sourabhdalvi:


sourabhdalvi:
there is still the issue of cost conversion for Market bid cost, @jd-lara  do you have any suggestion on how we should handle the cost conversion here ? 
https://github.com/NREL-SIIP/PowerSimulations.jl/blob/1004194d03e24a9c085402f28c4f9e672c7d53bf/src/devices_models/devices/common/cost_functions.jl#L547-L552

PR #779, opened 2021-11-02
Fixing hybrid tests

sourabhdalvi:


src/core/optimization_container.jl
```diff
@@ -583,7 +583,7 @@ function _add_variable_container!(
     axs...,
 ) where {T <: VariableType, U <: Union{PSY.Component, PSY.System}}
     if sparse
-        var_container = sparse_container_spec(JuMP.VariableRef, axs...)
+        var_container = sparse_container_spec(Float64, axs...)
```
jd-lara, line 586:
```suggestion
       # We initialize sparse containers with Floats64, not ideal and introduces type instability, because JuMP.Containers.SparseAxisArrays can't be initialized with undef
        var_container = sparse_container_spec(Float64, axs...)
```

src/utils/jump_utils.jl
```diff
@@ -93,10 +93,7 @@ function container_spec(::Type{Float64}, axs...)
 end
 
 """ Returns the correct container spec for the selected type of JuMP Model"""
-function sparse_container_spec(
-    ::Type{T},
-    axs...,
-) where {T <: Union{JuMP.AbstractJuMPScalar, Float64}}
+function sparse_container_spec(::Type{T}, axs...) where {T <: JuMP.AbstractJuMPScalar}
     indexes = Base.Iterators.product(axs...)
     contents = Dict{eltype(indexes), Any}(indexes .=> zero(T))
```
jd-lara, line 98:
```suggestion
    contents = Dict{eltype(indexes), Any}(indexes .=> 0.0)
```

jd-lara: Substitute the 0.0 in line 104 with `zero(T)`

PR #780, opened 2021-11-03
sim buildv3

jd-lara:
Opening PR for some initial reactions from @daniel-thom. Currently, the build process repeats what we had before building just the decision models. 

@daniel-thom should we also initialize the simulation store to the build call? I couldn't locate the place where `SimulationStoreProblemRequirements` gets instantiated.  I think this was lost somewhere in the refactor.


src/simulation/simulation_internal.jl
```diff
@@ -16,25 +15,33 @@ mutable struct SimulationInternal
     recorders::Vector{Symbol}
     console_level::Base.CoreLogging.LogLevel
     file_level::Base.CoreLogging.LogLevel
+    #cache_size_mib::Int
```
jd-lara, line 18: @daniel-thom I think these parameters should be defined here to be used later in the build process.

src/simulation/simulation.jl
```diff
@@ -196,49 +205,99 @@ function _check_folder(sim::Simulation)
...
+        _build_decision_models!(sim)
+        _build_emulation_model(sim)
+    end
+
+    TimerOutputs.@timeit BUILD_PROBLEMS_TIMER "Initialize Simulation State" begin
+        _initialize_simulation_state(sim)
     end
+
+
+    # _initialize_problem_storage!(sim, cache_size_mib, min_cache_flush_size_mib)
```
jd-lara, line 299: @daniel-thom The storage should be instantiated here. The code is pretty similar to what we will be doing for the `simulation_state`

src/simulation/simulation.jl
```diff
@@ -112,39 +108,35 @@ set_simulation_build_status!(sim::Simulation, status::BuildStatus) =
...
             if !(element + system_interval == model_initial_times[model_number][ix + 1])
-                throw(IS.ConflictingInputsError("The sequence of forecasts is invalid"))
+                throw(
+                    IS.ConflictingInputsError(
+                        "The sequence of forecasts in the model's systems are invalid",
+                    ),
+                )
             end
         end
         if !(sim_ini_time === nothing) &&
```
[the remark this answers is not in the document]
jd-lara: fixed via 842b8d1

src/simulation/simulation.jl
```diff
@@ -196,49 +205,98 @@ function _check_folder(sim::Simulation)
...
     TimerOutputs.@timeit BUILD_PROBLEMS_TIMER "Check Steps" begin
         _check_steps(sim, problem_initial_times)
     end
+
     TimerOutputs.@timeit BUILD_PROBLEMS_TIMER "Build Problems" begin
-        _build_problems!(sim, serialize)
-        # Make EmulationModel
-        # Make SimulationState here
+        _build_decision_models!(sim)
+        _build_emulation_model(sim)
```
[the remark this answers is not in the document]
jd-lara: 4c571ed

src/simulation/simulation.jl
```diff
@@ -196,49 +205,98 @@ function _check_folder(sim::Simulation)
...
+
+function _build_emulation_model(sim::Simulation)
+    model = get_emulation_model(get_models(sim))
+
+    if model === nothing
+        return
+    end
+
+    try
+        @show get_execution_count(model)
```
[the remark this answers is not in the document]
jd-lara: 4c571ed

src/simulation/simulation.jl
```diff
@@ -196,49 +205,98 @@ function _check_folder(sim::Simulation)
...
         set_initial_time!(model, initial_time)
         output_dir = joinpath(get_models_dir(sim))
         set_output_dir!(model, output_dir)
-        initialize_simulation_info!(model, problem_chronology)
-        problem_build_status = _build!(model, serialize)
-        if problem_build_status != BuildStatus.BUILT
-            error("Problem $(name) failed to build successfully")
+        try
+            TimerOutputs.@timeit BUILD_PROBLEMS_TIMER "Problem $(get_name(model))" begin
+                # Temporary while are able to switch from PJ to POI
```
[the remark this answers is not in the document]
jd-lara: 9d52c24

src/simulation/simulation.jl
jd-lara: We discussed this over the phone and decided to just instantiate all the store params and not the store itself here.

src/simulation/simulation_internal.jl
jd-lara: Save this values here so they are set at build time for consistency.

jd-lara:
@daniel-thom I made the adjustments to the cache and simulation store in c8a6537. I tried to keep most of the changes into a single commit to facilitate review.

src/simulation/simulation.jl
```diff
@@ -25,23 +25,19 @@ mutable struct Simulation
...
             if model.internal.simulation_info.sequence_uuid != sequence.uuid
+                model_name = get_name(model)
                 throw(
                     IS.ConflictingInputsError(
                         "The model definition for $model_name doesn't correspond to the simulation sequence",
                     ),
                 )
             end
         end
+        if models.emulation_model !== nothing
```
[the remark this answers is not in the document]
jd-lara: I don't understand. EmulationModel is defined as: 

```Julia
mutable struct EmulationModel{M <: EmulationProblem} <: OperationModel
    name::Symbol
    template::ProblemTemplate
    sys::PSY.System
    internal::ModelInternal
    store::InMemoryModelStore # might be extended to other stores for simulation
    ext::Dict{String, Any}
```

and the uuid is assigned to the model internal in the same place. This line of code was incorrect and was fixed in 35aff46

src/simulation/simulation.jl
```diff
@@ -196,49 +205,98 @@ function _check_folder(sim::Simulation)
...
+                # Temporary while are able to switch from PJ to POI
+                container = get_optimization_container(model)
+                container.built_for_recurrent_solves = true
+                build_impl!(model)
+            end
+            sim.internal.date_ref[model_number] = initial_time
+            set_status!(model, BuildStatus.BUILT)
+        catch e
+            set_status!(model, BuildStatus.FAILED)
+            rethrow(e)
```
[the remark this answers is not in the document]
jd-lara: I followed the patter that was there before, I changed it everywhere.

src/simulation/simulation.jl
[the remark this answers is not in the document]
daniel-thom: Yeah, my comment was based on the incorrect line. Looks good now.

src/simulation/hdf_simulation_store.jl
```diff
@@ -2,6 +2,7 @@ const HDF_FILENAME = "simulation_store.h5"
 const HDF_SIMULATION_ROOT_PATH = "simulation"
 const OPTIMIZER_STATS_PATH = "optimizer_stats"
 
+# Each row corresponds to a time stamp
```
daniel-thom, line 5: This probably isn't helpful because it has 3 dimensions.

src/simulation/hdf_simulation_store.jl
```diff
@@ -50,7 +51,7 @@ mutable struct HdfSimulationStore <: SimulationStore
     # The key is the problem name.
     optimizer_stats_datasets::Dict{Symbol, HDF5.Dataset}
     optimizer_stats_write_index::Dict{Symbol, Int}
-    cache::ResultCache
+    cache::OptimizationOutputCache
```
daniel-thom, line 54: We are using the terms `read_result`, `write_result`, `OutputCache`, `OptimizationOutputCache`, `OutputCacheKey`.

`result` and `output` are referring to the same thing, and so we could use the same term. Either is OK, but I think `result` is more specific.

Take this example:
```
add_result!(output_cache, ...)
```
could be
```
add_result!(result_cache, ...)
# or
add_output!(output_cache, ...)
```
 What do you think?

src/simulation/simulation.jl
```diff
@@ -196,49 +213,99 @@ function _check_folder(sim::Simulation)
...
         sim.internal.date_ref[model_number] = initial_time
+        set_status!(model, BuildStatus.BUILT)
+    catch
+        set_status!(model, BuildStatus.FAILED)
+        rethrow()
     end
     return
 end
 
+function _initialize_simulation_state(sim::Simulation) end
```
daniel-thom, line 261: Is this a placeholder for future work? Is that listed somewhere?

src/simulation/simulation.jl
daniel-thom: Should these lines be deleted?

src/simulation/simulation_store_params.jl
```diff
@@ -3,7 +3,7 @@ struct SimulationStoreParams
     step_resolution::Dates.Period
     num_steps::Int
     # The key order is the problem execution order.
-    models::OrderedDict{Symbol, ModelStoreParams}
+    problems::OrderedDict{Symbol, ModelStoreParams}
```
daniel-thom, line 6: If the type is ModelStoreParams, isn't `models` a better term?

src/simulation/hdf_simulation_store.jl
jd-lara: We need to distinguish the functions with the word `result` that imply an interface to the simulation results (post-processed) and the outputs of the optimizer. If we change the names from result to output we would be implying that the interface to the store is for that purpose and not direct result gathering.

src/simulation/simulation.jl
jd-lara: yes, this is a place holder at the moment.

src/simulation/simulation_store_params.jl
jd-lara: I switched to problems since the key is the problem name.

src/simulation/simulation.jl
jd-lara: Yeah, this will be substituted with the correct function call.

src/simulation/hdf_simulation_store.jl
jd-lara: resolved in 
4fbead9

PR #782, opened 2021-11-15
sim state dev2

jd-lara:
This is another step in the build of simulation that instantiates the state. I realized that without running a full simulation is difficult to know beforehand all the details of what is needed, so this is just another not final step.


src/core/optimization_container.jl
```diff
@@ -584,7 +584,7 @@ function _add_variable_container!(
 ) where {T <: VariableType, U <: Union{PSY.Component, PSY.System}}
     if sparse
         var_container = sparse_container_spec(Float64, axs...)
-        # We initialize sparse containers with Floats64, not ideal and introduces type instability, 
+        # We initialize sparse containers with Floats64, not ideal and introduces type instability,
```
daniel-thom, line 587: Float64

src/simulation/simulation_state.jl
```diff
@@ -1,32 +1,61 @@
...
+get_system_state(s::SimulationState) = s.system_state
+
+function _initialize_model_states!(
+    states::StateInfo,
+    model::OperationModel,
+    simulation_step::Dates.Period,
+)
+    container = get_optimization_container(model)
+    model_resolution = get_resolution(model)
+    value_counts = Int(simulation_step / model_resolution)
```
daniel-thom, line 32: Couldn't this produce an InexactError? Use `÷` instead?

jd-lara: In theory, this possible error should have been caught earlier. The issue is that if it isn't an integer doing `1:value_counts` will produce Floats and be inconsistent with the indexing.

src/core/optimization_container.jl
jd-lara: solved via eb7f474

src/simulation/simulation_state.jl
daniel-thom: `÷` will always produce an Int. Benefits of this:
- Concise code
- Future readers will never have to wonder whether there is a bug.

I haven't always done this because I didn't realize Julia offered it.

jd-lara: I had no idea, fixed via 84ad95d

daniel-thom (approved):

PR #786, opened 2021-11-17
Adds serialization/deserialization of initial conditions to OperationModels

daniel-thom:


test/test_model_decision.jl
```diff
@@ -556,3 +556,75 @@ end
...
 end
+
+@testset "Test serialization of InitialConditionsData" begin
+    sys = PSB.build_system(PSITestSystems, "c_sys5")
+    template = get_thermal_standard_uc_template()
+    set_service_model!(
+        template,
+        ServiceModel(VariableReserve{ReserveUp}, RangeReserve, "test"),
+    )
+    optimizer = GLPK_optimizer
```
jd-lara, line 567: we have the optimizers defined in `runtests.jl`

test/test_model_emulation.jl
```diff
@@ -478,3 +478,55 @@ end
...
+
+@testset "Test serialization of InitialConditionsData" begin
+    template = get_thermal_standard_uc_template()
+    sys = PSB.build_system(
+        PSITestSystems,
+        "c_sys5_pglib";
+        add_single_time_series = true,
+        force_build = true,
+    )
+    optimizer = Cbc_optimizer
```
jd-lara, line 490: same here, it is better to use the predefined optimizers so we can change them in a single place.

jd-lara (approved):

test/test_model_emulation.jl
daniel-thom: Maybe I'm missing you, but that is exactly what I tried to do here. Every `run` in this test is controlled by a single optimizer. You someone wants to use GLPK_optimizer instead of Cbc_optimizer, they can change it on this line.

PR #788, opened 2021-11-19
sim state part3

jd-lara:
Implements the simulation state update code and enables some tests of simulation execution. It also touches some of the recorders.


src/simulation/simulation_state.jl
```diff
@@ -1,44 +1,107 @@
...
-            column_names, _ = axes(value)
-            field_states[key] =
-                JuMP.Containers.DenseAxisArray{Float64}(undef, column_names, 1:value_counts)
+            value_counts = params[key][1] ÷ params[key][2]
+            if length(axes(value)) == 1
+                column_names = [string(encode_key(key))]
+            elseif length(axes(value)) == 2
+                column_names, _ = axes(value)
+            else
+                @warn("Multidimensional Array caching is not currently supported")
```
daniel-thom, line 85: Will everything still work in this case? Should it throw an error instead?

src/simulation/simulation_state.jl
```diff
@@ -48,15 +111,64 @@ function initialize_simulation_state!(
...
+    if simulation_time > end_of_step_timestamp
+        state_data_index = 1
+    else
+        # This seems to be a bug in indexin that requires an array when the types are Dates.DateTime
+        state_data_index = indexin([simulation_time], get_timestamps(state_data))[1]
+    end
+
+    offset = resolution_ratio - 1
+
+    # Not most optimal way to search. Most be improved before merging to master
```
daniel-thom, line 157: Probably need a better tag to find this TODO.

src/utils/recorder_events.jl
```diff
@@ -81,58 +81,59 @@ struct ParameterUpdateEvent <: IS.AbstractRecorderEvent
...
+    return StateUpdateEvent(
+        IS.RecorderEventCommon("StateUpdateEvent"),
+        execution_timestamp,
+        string(entry_type),
+        string(component_type),
+        string(model_name),
+    )
+end
+
+function StateUpdateEvent(
```
daniel-thom, line 127: Should we make a script that automatically generates files with textual tables of all recorder events? That could make it very easy to browse these. I could take care of it later.

src/simulation/simulation_state.jl
jd-lara: I put a warning since at the moment isn't clear when this ill happen

src/utils/recorder_events.jl
jd-lara: That's a good idea, we should do this.

PR #789, opened 2021-11-19
pm and jump update

jd-lara:
update of the dependencies.


jd-lara:
will fail until we get https://github.com/JuliaStochOpt/ParameterJuMP.jl/pull/86

daniel-thom (approved):

src/operation/operation_model_serialization.jl
```diff
@@ -2,15 +2,22 @@ const _SERIALIZED_MODEL_FILENAME = "model.bin"
...
 
 function OptimizerAttributes(model::OperationModel, optimizer::MOI.OptimizerWithAttributes)
-    name = JuMP.solver_name(get_jump_model(model))
+    jump_model = get_jump_model(model)
+    name = JuMP.solver_name(jump_model)
     # Note that this uses private field access to MOI.OptimizerWithAttributes because there
     # is no public method available.
     # This could break if MOI changes their implementation.
-    return OptimizerAttributes(name, optimizer.params)
+    if MOI.supports(JuMP.backend(jump_model), MOI.SolverVersion())
```
odow, line 15: This is an incorrect usage of `MOI.supports`.

You need a `try-catch` here instead. See https://github.com/jump-dev/MathOptInterface.jl/issues/1628 and related issues.

jd-lara: Thanks for the note, we are doing this in several places actually. Will have to re-implement as suggested.

PR #790, opened 2021-11-20
updates to hybrids and initialization 

sourabhdalvi:


src/devices_models/devices/thermal_generation.jl
```diff
@@ -91,19 +91,40 @@ initial_condition_variable(::InitialTimeDurationOff, d::PSY.ThermalGen, ::Abstra
 
 #! format: on
 function get_initial_conditions_device_model(
-    model::DeviceModel{T, D},
+    model::OperationModel,
+    ::DeviceModel{T, D},
 ) where {T <: PSY.ThermalGen, D <: AbstractThermalDispatchFormulation}
-    return DeviceModel(T, ThermalBasicDispatch)
+    optimizer_model = model.internal.container.JuMPmodel.moi_backend.optimizer.model
```
jd-lara, line 97: we already have a function to check if the model is MILP

PR #797, opened 2021-12-07
ic update

jd-lara:
I am opening this first as a draft PR now that the initial conditions code is implemented. I will add tests and conduct other fixes through the review process.


src/simulation/simulation.jl
```diff
@@ -639,19 +639,17 @@ function _execute!(
...
-                    elseif !get_allow_fails(settings) && (status != RunStatus.SUCCESSFUL)
-                        throw(
-                            ErrorException(
-                                "Simulation Failed in problem $(model_name). Returned $(status)",
-                            ),
-                        )
-                    else
-                        @assert status == RunStatus.SUCCESSFUL
-                    end
+                    # TODO: This block of code is currently not in the execution path after a failed solve
```
jd-lara, line 645: @daniel-thom The current way we have the try-catch and loggers, etc make it such that if the simulation fails it never uses this block of code we need to address that.

sourabhdalvi:
You need to delete this line here, that's adding the Pmin as a constant to the expression
https://github.com/NREL-SIIP/PowerSimulations.jl/blob/c182b13a5ee777ef9e966993be2972a832d61c47/src/devices_models/devices/common/add_to_expression.jl#L590

src/simulation/simulation.jl
daniel-thom: The simplest solution is to wrap this in a try-catch. I'm assuming that you want to keep the current throw logic in solve.

src/simulation/simulation.jl
```diff
@@ -656,89 +700,67 @@ function _execute!(
...
 
-        model_req[model_name] = reqs
+    if (get_simulation_build_status(sim) != BuildStatus.BUILT) ||
+       (get_simulation_status(sim) != RunStatus.READY)
+        error("Simulation status is invalid, you need to rebuild the simulation")
+    end
+    try
+        open_store(store_type, get_store_dir(sim), "w") do store
+            set_simulation_store!(sim, store)
+            # TODO: return file name for hash calculation instead of hard code
```
daniel-thom, line 736: Is this comment still relevant?

src/simulation/simulation_state.jl
```diff
@@ -112,24 +88,46 @@ function _initialize_model_states!(
                             length = value_counts,
                         ),
                     ),
+                    params[key][2],
```
daniel-thom, line 91: The `2` make this hard to read because I don't know what this Tuple is. Can it be a NamedTuple instead? Then you would use a field name here.

src/simulation/simulation_state.jl
```diff
@@ -166,7 +160,7 @@ function update_state_data!(
 
     offset = resolution_ratio - 1
     result_time_index = axes(store_data)[1]
-
+    state_data.last_recorded_row = state_data_index
```
daniel-thom, line 163: Use a setter function here per my earlier comment.

src/simulation/simulation.jl
jd-lara: We need to be careful not to wrap the time critical code in try-catch

src/simulation/simulation.jl
jd-lara: I think this has been there for a while, do you recall what this is about? it seems related to the set of the store file

src/initial_conditions/update_initial_conditions.jl
```diff
@@ -1,3 +1,143 @@
...
+end
+
+function _update_initial_conditions!(
+    ic_vector::Vector{T},
+    state::StateInfo,
+    ::Dates.Period,
+) where {T <: InitialCondition{DevicePower, S}} where {S <: Union{Float64, PJ.ParameterRef}}
+    for ic in ic_vector
+        status_data = get_state_data(state, OnVariable(), get_component_type(ic))
+        status_val = get_state_values(status_data)[1, get_component_name(ic)]
```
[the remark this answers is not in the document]
jd-lara: I think this syntax could be confusing because it will return the value of the state and not the current value of the ic. I made several changes to the syntax that hopefully help clean it up.

src/simulation/simulation.jl
```diff
@@ -452,11 +520,48 @@ function _update_simulation_state!(sim::Simulation, model::DecisionModel)
...
+        decision_field = getfield(decision_state, field)
+        for (key, data) in decision_field
+            if get_last_update_timestamp(decision_field[key]) <= simulation_time
+                get_state_values(system_state_field[key])[1, :] .=
+                    DataFrames.values(get_state_value(data, simulation_time))
+            elseif get_last_update_timestamp(decision_field[key]) > simulation_time
+                error("Something went really wrong. Please report this error.")
+            else
+                @assert false
+            end
```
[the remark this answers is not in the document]
jd-lara: about 3. The issue is that in data frames you can't write into the row easily. What we do here probably will work better with a setter function for the system state, but then we would have to make setter functions for the decision states too. I will leave a TODO for now.

jd-lara (commented):
@daniel-thom I addressed almost all comments except for the naming ones. I think we should talk about the names on Monday. I agree the current names for the state-related structs are confusing.

src/core/state_data.jl
```diff
@@ -0,0 +1,165 @@
...
+end
+
+function get_last_recorded_value(s::StateData)
+    if get_last_recorded_row(s) == 0
+        error("The State hasn't been written yet")
+    end
+    return _get_values(s)[get_last_recorded_row(s), :]
+end
+
+function set_last_recorded_row(s::StateData, val::Int)
```
daniel-thom, line 49:
```suggestion
function set_last_recorded_row!(s::StateData, val::Int)
```

src/core/state_data.jl
```diff
@@ -0,0 +1,165 @@
...
+
+function get_state_data(state::StateInfo, key::AuxVarKey)
+    return state.aux_variables[key]
+end
+
+function get_state_data(state::StateInfo, key::ConstraintKey)
+    return state.duals[key]
+end
+
+function set_state_data(state::StateInfo, key::VariableKey, val::StateData)
```
daniel-thom, line 84:
```suggestion
function set_state_data!(state::StateInfo, key::VariableKey, val::StateData)
```

src/core/state_data.jl
```diff
@@ -0,0 +1,165 @@
...
+    state.variables[key] = val
+    return
+end
+
+function set_state_data(state::StateInfo, key::AuxVarKey, val::StateData)
```
daniel-thom, line 89: ditto

src/simulation/simulation_state.jl
```diff
@@ -52,23 +28,28 @@ function set_current_time!(s::SimulationState, val::Dates.DateTime)
 end
 
 get_decision_states(s::SimulationState) = s.decision_states
-get_system_state(s::SimulationState) = s.system_state
+get_system_states(s::SimulationState) = s.system_state
```
daniel-thom, line 31: Should `SimulationState.system_state` be changed to `SimulationState.system_states`?

jd-lara:
@daniel-thom all the discussed changes are now implemented.

src/core/value_states.jl
```diff
@@ -0,0 +1,165 @@
...
+        return UNSET_INI_TIME
+    end
+    return get_timestamps(s)[get_last_recorded_row(s)]
+end
+
+function _get_state_value(s::ValueState, date::Dates.DateTime)
+    if _get_last_updated_timestamp(s) == date
+        s_index = get_last_recorded_row(s)
+    else
+        s_index = findlast(get_timestamps(s) .<= date)
```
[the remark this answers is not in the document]
jd-lara: sorry, I reversed the change in this file inadvertently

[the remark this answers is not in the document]
jd-lara: solved via 5330d20

daniel-thom (approved):

PR #801, opened 2021-12-18
General code clean up and bug fixes

sourabhdalvi:
Some of the changes in this branch are not completely implemented


jd-lara (commented):
Note to @jdlara before merging. This PR needs the fix to update the end-of-step timestamp as discussed with @sourabhdalvi in Slack.

PR #802, opened 2021-12-20
Initial condition update bug fix 

sourabhdalvi:


src/core/update_initial_conditions.jl
```diff
@@ -16,9 +16,11 @@ function calculate_ic_quantity(
     last_status = time_cache[:status]
     var_status = isapprox(var_value, 0.0, atol = ABSOLUTE_TOLERANCE) ? 0.0 : 1.0
     @debug last_status, var_status, abs(last_status - var_status)
-    @assert abs(last_status - var_status) < ABSOLUTE_TOLERANCE
+    # Commenting out this as when going from UC -> 1st ED solve,
```
jd-lara, line 19: This might require a different fix to make sure we are catching the correct index.

PR #803, opened 2021-12-21
Refactor store to support in-memory simulation data

daniel-thom:


src/operation/decision_model.jl
```diff
@@ -518,7 +518,7 @@ function write_model_dual_results!(
         if cols == get_time_steps(container)
             cols = ["System"]
         end
-        write_result!(store, STORE_CONTAINER_DUALS, key, timestamp, dual, cols)
+        write_result!(store, key, timestamp, dual, cols)
```
daniel-thom, line 521: I'd like to remove this from the interface. We can get it from the key.

src/operation/decision_model_optimizer_results.jl
```diff
@@ -0,0 +1,84 @@
+"""
```
daniel-thom, line 1: Since this struct is used within an InMemoryModelStore and InMemorySimulationStore, I think it merits its own file.

src/operation/emulation_model_optimizer_results.jl
```diff
@@ -0,0 +1,71 @@
...
+    return copy(container[key], copycols = true)
+end
+
+function write_optimizer_stats!(
+    results::EmulationModelOptimizerResults,
+    stats::OptimizerStats,
+    execution::Int,
+)
+    # TODO DT: This trips in one test. Should we enforce this rule?
+    #@assert !(execution in keys(results.optimizer_stats))
```
daniel-thom, line 57: Let me know what you think about this.

src/operation/decision_model_optimizer_results.jl
```diff
@@ -0,0 +1,84 @@
...
+        }(),
+        Dict{
+            AuxVarKey,
+            Dict{ExpressionKey, OrderedDict{Dates.DateTime, DataFrames.DataFrame}},
+        }(),
+        OrderedDict{Dates.DateTime, OptimizerStats}(),
+    )
+end
+
+function write_result!(
```
jd-lara, line 39: @daniel-thom should we support an interface that doesn't require passing the field and we choose it based on the `OptimizationContainerKey`?

src/operation/in_memory_model_store.jl
```diff
@@ -282,3 +179,7 @@ function get_parameter_value(
 ) where {T <: ParameterType, U <: Union{PSY.Component, PSY.System}}
     return store.data.parameters[ParameterKey(T, U)]
 end
+
+# TODO DT: Jose, do we need this for DecisionModel?
```
daniel-thom, line 183: Note this question

src/simulation/in_memory_simulation_store.jl
```diff
@@ -64,88 +59,97 @@ function write_optimizer_stats!(
...
+            for (name, reqs) in getfield(problem_reqs[problem], type)
+                container = getfield(store.dm_data[problem], type)
                 container[name] = OrderedDict{Dates.DateTime, DataFrames.DataFrame}()
+                @debug "Added $type $name in $problem" _group = LOG_GROUP_SIMULATION_STORE
             end
         end
 
-        store.optimizer_stats[model_name] = OrderedDict{Dates.DateTime, OptimizerStats}()
-        @debug "Initialized optimizer_stats_datasets $model_name"
+        # TODO EmulationModel: how do we differentiate DM and EM?
```
daniel-thom, line 109: Is this something we can fix now?

src/simulation/simulation_store_common.jl
```diff
@@ -0,0 +1,180 @@
+# Keep these in sync with the Symbols in src/core/definitions.
+get_store_container_type(::AuxVarKey) = STORE_CONTAINER_AUX_VARIABLES
+get_store_container_type(::ConstraintKey) = STORE_CONTAINER_DUALS
+get_store_container_type(::ExpressionKey) = STORE_CONTAINER_EXPRESSIONS
+get_store_container_type(::ParameterKey) = STORE_CONTAINER_PARAMETERS
+get_store_container_type(::VariableKey) = STORE_CONTAINER_VARIABLES
+
+function write_model_dual_results!(
```
daniel-thom, line 8: I moved this code here because it is common to both subtypes of SimulationStore.

src/operation/decision_model_optimizer_results.jl
jd-lara: We could add the interface that chooses the correct field based on the key type in the abstract

src/simulation/simulation_store_common.jl
```diff
@@ -0,0 +1,180 @@
...
+
+    for (key, constraint) in get_duals(container)
+        write_result!(
+            store,
+            model_name,
+            STORE_CONTAINER_DUALS,
+            key,
+            timestamp,
+            constraint,
+            [encode_key(key)],  # TODO DT: is this what the columns should be?
```
daniel-thom, line 29: This old comment applies to a current problem. We need a rule here and a common way to do it.

src/operation/emulation_model_optimizer_results.jl
```diff
@@ -0,0 +1,71 @@
...
+    # Return a copy because callers may mutate it.
+    return copy(container[key], copycols = true)
+end
+
+function write_optimizer_stats!(
+    results::EmulationModelOptimizerResults,
+    stats::OptimizerStats,
+    execution::Int,
+)
+    # TODO DT: This trips in one test. Should we enforce this rule?
```
jd-lara, line 56: We should enable it and then confirm why it happens.

test/test_simulation_execute.jl
```diff
@@ -15,11 +15,17 @@
     )
     build_out = build!(sim_single)
     @test build_out == PSI.BuildStatus.BUILT
-    execute_out = execute!(sim_single)
+    execute_out = execute!(sim_single, in_memory = in_memory)
     @test execute_out == PSI.RunStatus.SUCCESSFUL
 end
 
-@testset "2-Stage Decision Models with FeedForwards" begin
+@testset "Single stage sequential tests" begin
```
daniel-thom, line 22: This pattern doesn't add too much time and it is valuable.

src/operation/decision_model_optimizer_results.jl
daniel-thom: Refer to the functions in abstract_model_optimizer_results. I handled it there.

src/operation/operation_model_interface.jl
```diff
@@ -297,23 +297,23 @@ function _list_names(model::OperationModel, container_type)
 end
 
 function read_dual(model::OperationModel, key::ConstraintKey)
```
jd-lara, line 299: Are these functions meant to be used for the export? or by the user?

src/operation/in_memory_model_store.jl
```diff
@@ -1,74 +1,12 @@
...
 Stores simulation data in memory
 """
 mutable struct InMemoryModelStore{T <: AbstractModelOptimizerResults}
     data::T
-    optimizer_stats::OrderedDict{Int, OptimizerStats}
 end
 
 function InMemoryModelStore(::Type{T}) where {T <: AbstractModelOptimizerResults}
-    return InMemoryModelStore(T(), OrderedDict{Int, OptimizerStats}())
+    return InMemoryModelStore(T())
```
daniel-thom, line 9: This file is now almost useless. If we moved the initialize function to each of the subtypes then each model can simply create the desired type.

src/operation/operation_model_interface.jl
daniel-thom: Anything that takes a key should be internal only. Why do you ask?

src/simulation/in_memory_simulation_store.jl
jd-lara: I was thinking about using `get_decision_model_results` and `get_emulation_model_results` to emphasize the difference. Honing down the concept on users is critical.

src/simulation/simulation_store_common.jl
jd-lara: agreed, we should discuss how to make this consistent for one dimensional variables.

test/test_simulation_execute.jl
jd-lara: yeah, trimming down testing is important.

src/operation/operation_model_interface.jl
jd-lara: Just to see what you were thinking about the use of the functions.

src/operation/in_memory_model_store.jl
jd-lara: we can do that if you prefer. Probably the design was over complicated.

src/simulation/simulation_store_common.jl
jd-lara: ok.

src/simulation/in_memory_simulation_store.jl
daniel-thom: You might have missed my question. In this function we iterate over the problems in `problem_reqs`. It is currently assuming that every problem is a DecisionModel. If there is an EmulationModel, how do we tell?

jd-lara: > You might have missed my question. In this function we iterate over the problems in `problem_reqs`. It is currently assuming that every problem is a DecisionModel. If there is an EmulationModel, how do we tell?

TBD, the trick here is that for 1.0 we will support only one EM and if there is none, this is populated from the `SystemState`

daniel-thom: The HDF5 datasets need to get created before the simulation starts. How will that happen?

src/operation/decision_model.jl
jd-lara: ok

src/operation/emulation_model_optimizer_results.jl
jd-lara: We should add the @assert and then fix the test to see why is failing.

src/simulation/in_memory_simulation_store.jl
jd-lara: > The HDF5 datasets need to get created before the simulation starts. How will that happen?

We know how many executions will happen the tables will have `n_executions` number of rows.

src/operation/in_memory_model_store.jl
jd-lara: No, last recorded row is not needed for DecisionModel. In the Emulation model, it is useful because we pre-allocate the DataFrame.

src/operation/emulation_model_optimizer_results.jl
daniel-thom: We weren't clearing the model stores. Fixed.

jd-lara (approved):

PR #804, opened 2021-12-22
Minor updates

sourabhdalvi:
* Refactoring hybrid range constraint code
* Updating methods to save sparse axis arrays in the right order
* Updating the indexing od piece linear cost variables 
* Adding hybrid test simulation


src/devices_models/devices/common/cost_functions.jl
```diff
@@ -270,15 +270,14 @@ function pwl_gencost_sos!(
     for i in 1:length(cost_data)
         pwlvars[i] = JuMP.@variable(
             container.JuMPmodel,
-            # TODO: Redefine the base_name for PWL vars.
-            # base_name = "{$(variable)}_{sos}",
+            base_name = "{$(variable)}_{sos_$(i)}",
```
jd-lara, line 273: Did you test how does this look?

jd-lara (commented):
Check the printing of the sos please

jd-lara (changes requested):
@sourabhdalvi can you make a test with a two-stage model Hybrid please?

src/devices_models/devices/common/cost_functions.jl
sourabhdalvi: this is how the variable name looks like , `ActivePowerVariable_ThermalStandard_{Alta, 2}_{pwl_3}` `ActivePowerVariable_ThermalStandard_{Alta, 2}_{sos_2}`

src/utils/jump_utils.jl
```diff
@@ -66,16 +66,27 @@ function to_array(array::DenseAxisArray{<:Number})
 end
 
 function to_array(array::SparseAxisArray)
```
jd-lara, line 68: @daniel-thom there are some coming changes to this function to support Hybrid Systems

src/utils/jump_utils.jl
```diff
@@ -66,16 +66,27 @@ function to_array(array::DenseAxisArray{<:Number})
...
+    final_column_names = Vector{Symbol}()
     for (ix, col) in enumerate(columns)
-        res = values(filter(v -> first(v)[[1, 3]] == col, array.data))
-        tmp_data[col] = jump_value.(res)
+        filtered_vars =
+            collect(filter(v -> (first(v)[[1, 2]] == col) && (last(v) != 0), array.data))
+        sorted_vars = sort(filtered_vars, by = x -> x[1][3])
+        vars = last.(sorted_vars)
+        if !isempty(vars)
+            tmp_data[Symbol(col...)] = PSI._jump_value.(vars)
```
jd-lara, line 79: @sourabhdalvi you don't need the `PSI.` here. I overlooked it.

src/utils/jump_utils.jl
jd-lara, line 68:
```suggestion
function to_array(array::SparseAxisArray)
    columns = Set()
    timesteps = Set{Int}()
    for k in keys(array.data)
        push!(columns, (k[1], k[2]))
        push!(timesteps, k[3])
    end

    data = Array{Float64, 2}(
        undef,
        length(timesteps),
        length(columns),
    )

    for (ix, col) in enumerate(columns), t in timesteps
        data[t, ix] = array.data[(col..., t)]
    end

    return data
end
```

PR #805, opened 2021-12-22
Cb/log msg errors

claytonpbarrows:


src/devices_models/device_constructors/load_constructor.jl
```diff
@@ -279,7 +278,7 @@ function construct_device!(
     add_to_expression!(
         container,
         ReactivePowerBalance,
-        ReactivePowerTimeSeriesParameter,
+        ActivePowerTimeSeriesParameter,
```
jd-lara, line 281: Add a note here please about why we do this. 

```suggestion
        # The base assumption is that reactive power follows the same curve as active power
        ActivePowerTimeSeriesParameter,
```

src/devices_models/device_constructors/load_constructor.jl
```diff
@@ -345,7 +343,7 @@ function construct_device!(
...
```
jd-lara, line 346: Add a note here please about why we do this. 

```suggestion
        # The base assumption is that reactive power follows the same curve as active power
        ActivePowerTimeSeriesParameter,
```

src/parameters/update_parameters.jl
```diff
@@ -40,6 +40,11 @@ function update_parameter_values!(
     horizon = get_time_steps(get_optimization_container(model))[end]
     components = get_available_components(V, get_system(model))
     for component in components
+        name = PSY.get_name(component)
+        @debug "Updating parameter value with time series" name,
```
jd-lara, line 44: there is a recorder entry for this. It is better to avoid @debug statements in hot loops like this. Alternatively, we can just add a comment to uncomment the line for debugging.

src/parameters/update_parameters.jl
```diff
@@ -156,6 +160,7 @@ function update_parameter_values!(
     # if the keys have strings in the meta fields
     parameter_array = get_parameter_array(optimization_container, key)
     parameter_attributes = get_parameter_attributes(optimization_container, key)
+    @debug "Updating parameter" key
```
jd-lara, line 163: same here

jd-lara (changes requested):
The most important change is to avoid the `@debug` statements in hot loops. In the past it resulted in extremely large log files if the configuration is set to `@debug` in the file_model

jd-lara (approved):

jd-lara:
@claytonpbarrows it seems this change is making the AC-PF tests fail.

claytonpbarrows:
Yeah... I can take a look later today.

test/Project.toml
```diff
@@ -19,6 +19,7 @@ PowerSimulations = "e690365d-45e2-57bb-ac84-44ba829e73c4"
 PowerSystemCaseBuilder = "f00506e0-b84f-492a-93c2-c0a9afc4364e"
 PowerSystems = "bcd98974-b02a-5e2f-9ee0-a103f5c450dd"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+SCIP = "82193955-e24f-5292-bf16-6f2c5261a85f"
```
jd-lara, line 22: remove SCIP from toml please to avoid Windowd build errors in CI.

src/core/optimization_container.jl
```diff
@@ -856,12 +856,13 @@ function add_param_container!(
     name::String,
     axs...;
     meta = CONTAINER_KEY_EMPTY_META,
+    time_series_multiplier_id = 1,
```
jd-lara, line 859: we should default to `1` when making the attributes and modify the attribute at the `add_param_container!` call site.

src/core/parameters.jl
```diff
@@ -27,10 +27,20 @@ struct NoAttributes end
...
+    ::Type{T},
+    name::String,
+    multiplier_id::Int = 1,
+) where {T <: PSY.TimeSeriesData}
+    return TimeSeriesAttributes{T}(name, multiplier_id)
 end
 
 get_time_series_type(::TimeSeriesAttributes{T}) where {T <: PSY.TimeSeriesData} = T
 get_time_series_name(attr::TimeSeriesAttributes) = attr.name
+get_time_series_multiplier_id(attr::TimeSeriesAttributes) = attr.multiplier_id
```
jd-lara, line 43: add a setter function for this attribute

src/operation/time_series_interface.jl
```diff
@@ -59,6 +59,7 @@ function get_time_series_values!(
     initial_time,
     horizon::Int;
     ignore_scaling_factors = true,
+    multiplier_id = "",
```
daniel-thom, line 62: Is there a conflict with int vs string?

jd-lara, line 62: If the id is and `Int` why is it a `String` here.

src/parameters/add_parameters.jl
```diff
@@ -114,7 +125,10 @@ function add_parameters!(
     if !(ts_type <: Union{PSY.AbstractDeterministic, PSY.StaticTimeSeries})
         error("add_parameters! for TimeSeriesParameter is not compatible with $ts_type")
     end
-    ts_name = get_time_series_names(model)[T]
+    ts_names = get_time_series_names(model)
+    ts_name = ts_names[T]
+    time_series_multiplier_id =
+        findfirst(x -> x == T, [first(p) for p in ts_names if last(p) == ts_name])
```
claytonpbarrows, line 131: @daniel-thom and @jd-lara is this OK?
It will return 1 for everything where there is only one occurrence of a time series name in the dict. If there are more than one entries referencing a time series name, it will return different values for each parameter.

src/core/optimization_container.jl
claytonpbarrows: so you're saying that this function should be?

```
function add_param_container!(
    ::T,
    ::Type{U},
    ::Type{V},
    name::String,
    axs...;
    meta = CONTAINER_KEY_EMPTY_META,
) where {T <: TimeSeriesParameter, U <: PSY.Component, V <: PSY.TimeSeriesData}
    param_key = ParameterKey(T, U, meta)
    if isabstracttype(V)
        error("$V can't be abstract: $param_key")
    end
    attributes = TimeSeriesAttributes(V, name)
    return _add_param_container!(container, param_key, attributes, axs...)
end
```

jd-lara: Yes, it should be unchanged and then we manipulate the attribute at the call site.

src/parameters/add_parameters.jl
jd-lara: Can you describe what you are looking to achieve? I don't immediately understand the logic.

daniel-thom: In principle, nice solution. I didn't know that ts_names was a dict, and so I was tripped up at first. Is it more clear when written as
```
[k for (k, v) in ts_names if v == ts_name]
```
Secondly, since the logic is non-trivial and you repeat it, I think it would be good to put it in a function.

Per Jose's comment, perhaps the name of the function could serve as a comment.

src/parameters/add_parameters.jl
```diff
@@ -65,10 +65,21 @@ function add_parameters!(
     end
     time_steps = get_time_steps(container)
     names = [PSY.get_name(d) for d in devices]
-    ts_name = get_time_series_names(model)[T]
-    @debug "adding" T name ts_type _group = LOG_GROUP_OPTIMIZATION_CONTAINER
-    parameter_container =
-        add_param_container!(container, T(), D, ts_type, ts_name, names, time_steps)
+    ts_names = get_time_series_names(model)
+    ts_name = ts_names[T]
```
jd-lara, line 69: you probably want a conditionat like 

```julia
if length(ts_name) > 1
   do something 
else
  ts_name = ts_name[1]
end

Comment (claytonpbarrows) [reply src/parameters/add_parameters.jl]: ec9c38d  should address this

File: src/parameters/add_parameters.jl [hunk 11]
```diff
@@ -1,3 +1,17 @@
...
+Function to create a unique index of time series names for each device model. For example,
+if two parameters each reference the same time series name, this function will return a
+different value for each parameter entry
+"""
+function create_time_series_multiplier_index(
+    model,
+    ::Type{T},
+) where {T <: TimeSeriesParameter}
+    ts_names = get_time_series_names(model)
+    ts_name = ts_names[T]
```
jd-lara, line 11: you can run a check here for the size of `ts_name` and only run line 12 is size > 1

PR #806, opened 2021-12-22
updates for milp duals and conflict

jd-lara:
@daniel-thom I am opening the PR because there might be some merge between your PR and this one when it comes to reading the variables and expressions to write them into the store.


src/core/optimization_container.jl
```diff
@@ -1271,8 +1291,14 @@ function _calculate_dual_variables_continous_model!(
 end
 
 function _process_duals(container::OptimizationContainer, lp_optimizer)
```
jd-lara, line 1293: @odow this is the implementation we followed to retrieve the "duals" of the MILP UC models. In case you have time to check it out please.

src/core/optimization_container.jl
```diff
@@ -32,20 +32,36 @@ get_container_key(x::OptimizationContainerMetadata, key) = x.container_key_looku
...
+
+function PrimalValuesCache()
+    return PrimalValuesCache(
+        Dict{VariableKey, AbstractArray}(),
+        Dict{ExpressionKey, AbstractArray}(),
+    )
+end
+
+function Base.isempty(pvc::PrimalValuesCache)
+    return isempty(pvc.variables_cache) || isempty(pvc.expressions_cache)
```
daniel-thom, line 48: Should this be `&&`?

jd-lara: yeah, my bad.

daniel-thom (approved):

PR #808, opened 2021-12-22
fix matrix 

jd-lara:


src/simulation/hdf_simulation_store.jl
```diff
@@ -699,16 +699,23 @@ function _read_length(::Type{OptimizerStats}, store::HdfSimulationStore)
...
-    if ndims(array) == 2
-        dataset[:, 1, row_range] = array
-    elseif ndims(array) == 3
-        dataset[:, :, row_range] = array
-        #elseif ndims(array) == 4
-        #    dataset[:, :, :, row_range] = array
-    else
-        error("ndims not supported: $(ndims(array))")
-    end
+function _write_dataset!(dataset, array::Array{Float64, 2}, row_range::UnitRange{Int64})
```
daniel-thom, line 702: Does row_range need to be strictly typed?

jd-lara: make sure that it is continuous and Int so that can be used for indexing.

daniel-thom: I have found the following:
- Sometimes there is another type that can be passed and will work. The strict typing will cause an unnecessary failure. Not sure if that applies here.
- If you use duck typing here and the user passes the wrong type, Julia will fail with a meaningful error.
- Adding the type in this case doesn't help for dispatch or documentation.

src/simulation/hdf_simulation_store.jl
```diff
@@ -699,16 +699,23 @@ function _read_length(::Type{OptimizerStats}, store::HdfSimulationStore)
...
     @debug "wrote dataset" dataset row_range
+    return
+end
+
+function _write_dataset!(
+    dataset,
+    ::Array{Float64, N},
+    row_range::UnitRange{Int64},
+) where {N}
+    error("ndims not supported: $N")
```
daniel-thom, line 719: ndims? Why not leave this unimplemented?

src/simulation/hdf_simulation_store.jl
jd-lara: I am fine removing it. But Julia won't throw a meaningful error in this case if the user passes a non-contiguous vector. It will actually execute.

src/simulation/hdf_simulation_store.jl
jd-lara: I thought It would be better to Throw an error, I am fine removing it

daniel-thom: That is what I started doing. We had a team discussion and decided that the most "julian" thing to do is not implemented the fall-through method.

jd-lara: ok, removing it

daniel-thom (approved):

PR #812, opened 2021-12-27
sort columns

jd-lara:


src/utils/jump_utils.jl
```diff
@@ -106,13 +112,14 @@ function _to_matrix(
 end
 
 function to_matrix(array::SparseAxisArray{T, N, K}) where {T, N, K <: NTuple{N, Any}}
-    columns = Set(k[1:(N - 1)] for k in keys(array.data))
+    # Don't use _get_column_names to avoid allocation of the columns vector
+    columns = sort!(Set(k[1:(N - 1)] for k in keys(array.data)))
```
daniel-thom, line 116: Are you sure that you can `sort!(Set)`?

jd-lara: I shouldn't have sorted here

jd-lara: Fixed via 5de53a9

src/utils/jump_utils.jl
```diff
@@ -70,16 +70,22 @@ end
...
-    arr::DenseAxisArray{T, 2, K},
+    array::DenseAxisArray{T, 2, K},
 ) where {T, K <: NTuple{2, Any}}
-    return axes(arr)[1]
+    return axes(array)[1]
+end
+
+function _get_column_names(
+    arr::SparseAxisArray{T, N, K},
+)::Vector{String} where {T, N, K <: NTuple{N, Any}}
```
daniel-thom, line 80: Do you need `::Vector{String}`? Isn't `encode_tuple_to_column` guaranteed to produce strings?

jd-lara: I added it because `@code_warntype` wasn't detecting the return type. I could remove it.

jd-lara: fixed via 23c1f56

src/utils/jump_utils.jl
```diff
@@ -106,13 +112,14 @@ function _to_matrix(
 end
 
 function to_matrix(array::SparseAxisArray{T, N, K}) where {T, N, K <: NTuple{N, Any}}
+    # Don't use _get_column_names to avoid allocation of the columns vector
```
daniel-thom, line 115: It doesn't matter for correctness. I'm wondering if you want the columns sorted so that the results are deterministic for users.

jd-lara: we then would have to call `sort!(unique!())` instead of set.

jd-lara: 226de8d

daniel-thom (approved):

PR #817, opened 2022-01-12
WIP: Restore functionality for reading realized results

daniel-thom:
Please see if this is on the right track for realized variable results. I restored some deleted code and then rearranged some things to make it work.


src/simulation/simulation_problem_results.jl
```diff
@@ -529,148 +493,200 @@ function _read_optimizer_stats(res::SimulationProblemResults, store::SimulationS
...
+)
+    meta = RealizedMeta(res, initial_time = initial_time, len = len)
+    timestamps = _process_timestamps(res, meta.initial_time, meta.count)
+    result_values = _read_variables(res, variables, timestamps, nothing)
+    timestamps = get_realized_timestamps(
+        res,
+        initial_time = initial_time,
+        len = meta.count * meta.interval_len,
+    )
+    return get_realization(result_values, meta, timestamps)
```
daniel-thom, line 636: Is this actually correct?

jd-lara: What does `get_realization` do?  I don't recall

claytonpbarrows: `get_realization` is the code that stitches the final result for each timestamp in a range into a contiguous time series

claytonpbarrows (commented):
This seems correct to me. I think this re-enables everything that PG relies on.

jd-lara (approved):
approval to merge into other WIP branch

PR #818, opened 2022-01-14
store refactors

jd-lara:


src/core/dataset.jl
```diff
@@ -0,0 +1,148 @@
...
+    return s.values[get_last_recorded_row(s), :]
+end
+
+get_data_resolution(s::DataFrameDataset) = s.resolution
+
+function get_end_of_step_timestamp(s::AbstractDataset)
+    return get_timestamps(s)[s.end_of_step_index]
+end
+
+get_timestamps(s::DataFrameDataset) = s.timestamps
```
daniel-thom, line 73: Since this can't be part of the external interface, I suggest removing the function and always use dot access.

src/core/dataset.jl
```diff
@@ -0,0 +1,148 @@
+abstract type AbstractDataset end
```
daniel-thom, line 1: My main comment is that these types aren't quite polymorphic. They are not going to return the same types. For example, `get_values`. It is returning a `DataFrame` currently. The HDF5Dataset can't do that. Here is a call site:
```
get_values(container[key])[index, :] = df_row
```
You could invert the logic to this:
```
set_values!(container, key, index, df_row)
```
The same function and inputs would work for both both types. The HDF5 could would transform the dataframe row as necessary. The dataframe/dataset would not need to be returned to the caller.

I suggest documenting all of the external interfaces right here. By "external" I mean external to the file where callers don't care whether they have dataframes or hdf5 datasets.

src/core/dataset.jl
```diff
@@ -0,0 +1,148 @@
...
+    write_index::Int
+    last_recorded_row::Int
+    update_timestamp::Dates.DateTime
+end
+
+function get_column_names(::OptimizationContainerKey, s::HDF5Dataset)
+    return s.column_dataset[:]
+end
+
+Base.length(s::AbstractDataset) = size(s.values)[1]
```
daniel-thom, line 117: I suggest moving these methods at the top of the file next to the abstract type.

src/core/dataset_container.jl
```diff
@@ -0,0 +1,179 @@
...
+
+function get_dataset(container::DatasetContainer, key::OptimizationContainerKey)
+    datasets = getfield(container, get_store_container_type(key))
+    return datasets[key]
+end
+
+function set_dataset!(
+    container::DatasetContainer{T},
+    key::OptimizationContainerKey,
+    val::T,
```
daniel-thom, line 64: The `T` doesn't look important. Can you remove it?
```
val::AbstractDataset
```

src/operation/emulation_model_store.jl
```diff
@@ -133,8 +138,8 @@ function write_result!(
     update_timestamp::Dates.DateTime,
     df_row::DataFrames.DataFrameRow,
 )
-    container = getfield(store, get_store_container_type(key))
-    container[key][index, :] = df_row
+    container = getfield(store.data_container, get_store_container_type(key))
+    get_values(container[key])[index, :] = df_row
```
daniel-thom, line 147: Would it be more clear to invert the logic like
```
set_values!(container, key, index, df_row)
```

src/core/dataset_container.jl
```diff
@@ -0,0 +1,179 @@
...
+        for val in values(field_dict)
+            empty!(val)
+        end
+    end
+    @debug "Emptied the store" _group = LOG_GROUP_SIMULATION_STORE
+    return
+end
+
+function get_duals_values(container::DatasetContainer)
+    return container.duals
```
daniel-thom, line 31: These functions are going to return different types based on the underlying type. Is that ok?

src/core/dataset_container.jl
```diff
@@ -0,0 +1,179 @@
...
+
+function get_dataset(
+    container::DatasetContainer,
+    ::T,
+    ::Type{U},
+) where {T <: ExpressionType, U <: Union{PSY.Component, PSY.System}}
+    return get_dataset(container, ExpressionKey(T, U))
+end
+
+function get_dataset_values(container::DatasetContainer, key::OptimizationContainerKey)
```
daniel-thom, line 116: Same comment about these functions and different return types.

src/core/dataset.jl
```diff
@@ -0,0 +1,148 @@
...
+
+mutable struct HDF5Dataset <: AbstractDataset
+    values::HDF5.Dataset
+    column_dataset::HDF5.Dataset
+    write_index::Int
+    last_recorded_row::Int
+    update_timestamp::Dates.DateTime
+end
+
+function get_column_names(::OptimizationContainerKey, s::HDF5Dataset)
```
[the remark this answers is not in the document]
jd-lara: Yeah, it is part of the interface for the containers. It was like this before.

src/core/dataset_container.jl
jd-lara: We can probably remove these methods in favor of others.

src/core/dataset_container.jl
jd-lara: This function is more problematic, I don't have a good alternative in this case unless we always return an array.

daniel-thom: We could always return either a dataframe or an array.

src/core/dataset_container.jl
jd-lara: it does, as it forces that the parametric type of the container matches the type of value being written

src/operation/emulation_model_store.jl
jd-lara: fixed via 927d781e96adf851a8ed98243142c4a4f2e64289

src/core/dataset.jl
jd-lara: fixed in part via 927d781e96adf851a8ed98243142c4a4f2e64289, still need to implement the HDF version of this

src/core/dataset_container.jl
daniel-thom: That is technically not necessary because the addition to `datasets` will fail if the type is wrong. But I guess I don't have a problem with it.

src/core/dataset.jl
jd-lara: I am still figuring out if I will need the timestamps here or not.

src/core/dataset_container.jl
jd-lara: I would need to refactor `_get_emulation_store_requirements` to be able to remove these methods.

src/core/dataset_container.jl
jd-lara: Still pending what the requirements are until we have the HDF data set implemented.

jd-lara:
@daniel-thom I flagged the PR as ready for review since the tests are passing and the code is working. Still requires implementing the caching of the emulation_model data. I think we should leave that for another PR and merge all the pending PR's into dev.

src/core/dataset_container.jl
jd-lara: I ran a @code_warntype and it seems not to matter. I think we can leave it like that since this is not in the critical path

src/simulation/hdf_simulation_store.jl
```diff
@@ -451,6 +468,21 @@ end
 """
 Write an emulation model result for a timestamp to the store.
```
daniel-thom, line 469: Is this docstring still correct? My understanding is that the timestamp is the simulation time, not necessarily the timestamp for the result which correlates to the index.

src/simulation/hdf_simulation_store.jl
```diff
@@ -734,18 +767,55 @@ function _read_data_columns(
...
+    index::EmulationModelIndexType,
+)
+    if is_cached(store.cache, model_name, key, index)
+        # data = read_result(store.cache, model_name, key, index)
+        #columns = _get_em_dataset(store, model_name, key).column_dataset[:]
+    else
+        data, columns = _read_result(store, model_name, key, index)
+    end
+
+    return data, columns
```
daniel-thom, line 783: This looks like a bug - will crash if is_cached is true.

src/simulation/hdf_simulation_store.jl
```diff
@@ -734,18 +767,55 @@ function _read_data_columns(
...
+end
+
 function _read_length(::Type{OptimizerStats}, store::HdfSimulationStore)
     dataset = _get_dataset(OptimizerStats, store)
     return HDF5.read(HDF5.attributes(dataset), "columns")
 end
 
-function _write_dataset!(dataset, array::Array{Float64, 2}, row_range::UnitRange{Int64})
+function write_dataset!(
```
daniel-thom, line 791: Previously, this function started with an underscore because it was private to this file. It was not intended to be used externally because it writes to the HDF5 file. The external interface needs to account for the cache. Has any of that changed?

src/core/dataset_container.jl
daniel-thom: I recommend defining this only for the dataframe version - `DatasetContainer{DataFrameDataset}`. We don't ever want to expose the values attribute of `HDF5Dataset` because of the caching layer in `Hdf5SimulationStore`.

src/utils/datetime_utils.jl
```diff
@@ -1,7 +1,10 @@
 """
 calculates the index in the time series corresponding to the data. Assumes that the dates vector is sorted.
 """
-function find_timestamp_index(dates::Vector{Dates.DateTime}, date::Dates.DateTime)::Int
+function find_timestamp_index(
+    dates::Union{Vector{Dates.DateTime}, StepRange{Dates.DateTime, <:Dates.Period}},
+    date::Dates.DateTime,
+)::Int
```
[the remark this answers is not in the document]
jd-lara: not anymore with the change to Millisecond. I removed it.

src/simulation/hdf_simulation_store.jl
jd-lara: I changed the name to use it in `set_value!()` in the dataset interface

src/simulation/hdf_simulation_store.jl
jd-lara: Right now it is short circuited. I will comment the code in the mean time.

jd-lara (commented):
I addressed almost all of the changes except one that merits more discussion.

src/simulation/hdf_simulation_store.jl
jd-lara: @daniel-thom fixed via e75c5af

test/test_simulation_results.jl
```diff
@@ -433,7 +432,9 @@ function test_simulation_results(file_path::String, export_path; in_memory = fal
 end
 
 @testset "Test simulation results" begin
-    for in_memory in (true, false)
+    # TODO: hdf5 store is not currently working
```
daniel-thom, line 435: Does this work now?

jd-lara: yeah, it works. I should remove this comment

daniel-thom (approved):

PR #827, opened 2022-01-27
restore agc

jd-lara:
The values in the tests changed because in the new implementation we don't have the hack that created the system balance during the device constraint building.

### DIFF
--- a/src/services_models/services_constructor.jl
+++ b/src/services_models/services_constructor.jl
@@ -297,7 +297,7 @@ function construct_service!(
     services = PSY.get_components(S, sys)
     agc_areas = PSY.get_area.(services)
     areas = PSY.get_components(PSY.Area, sys)
-    if !isemtpy(setdiff(areas, agc_areas))
+    if !isempty(setdiff(areas, agc_areas))
         throw(
             IS.ConflictingInputsError(
                 "All area most have an AGC service assigned in order to model the System's Frequency regulation",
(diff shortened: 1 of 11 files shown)


src/services_models/services_constructor.jl
```diff
@@ -294,13 +294,16 @@ function construct_service!(
...
-    name = get_service_name(model)
-    service = PSY.get_component(S, sys, name)
-    agc_area = PSY.get_area(service)
+    services = PSY.get_components(S, sys)
+    agc_areas = PSY.get_area.(services)
     areas = PSY.get_components(PSY.Area, sys)
     for area in areas
-        if area != agc_area
-            #    throw(IS.ConflictingInputsError("All area most have an AGC service assigned in order to model the System's Frequency regulation"))
+        if area ∉ agc_areas
```
daniel-thom, line 301: In case you care, a setdiff would be more performant.

src/services_models/services_constructor.jl
```diff
@@ -311,11 +314,30 @@ function construct_service!(
...
+        container,
+        EmergencyDown,
+        AdditionalDeltaActivePowerDownVariable,
+        areas,
+        model,
+    )
+
+    add_to_expression!(container, RawACE, SteadyStateFrequencyDeviation, services, model)
+
+    # add_feedforward_arguments!(container, model, service)
```
daniel-thom, line 340: Is this supposed to be commented out? Deleted?

jd-lara: Fixed

src/services_models/services_constructor.jl
jd-lara: Fixed

daniel-thom (approved):

PR #829, opened 2022-02-01
ignore not assigned entries

jd-lara:
In some cases, we skip constraints when the data would make it an irrelevant constraint. This can cause trouble when checking the numerical bounds.


jd-lara:
@scdhulipala this PR fixes the issue you found try it.

daniel-thom (approved):

PR #830, opened 2022-02-01
Cb/aux vars and expressions

claytonpbarrows:
One remaining item worth flagging that I haven't addressed here is the `write_to_CSV`. (edit: Done)


src/operation/problem_results.jl
```diff
@@ -259,6 +245,21 @@ function read_dual(res::ProblemResults, args...)
     return read_dual(res, key)
 end
 
+function read_aux_variable(res::ProblemResults, key::AuxVarKey)
```
jd-lara, line 248: Looks like this function needs a test.

src/operation/problem_results.jl
```diff
@@ -296,7 +297,9 @@ function set_system!(res::ProblemResults, system::PSY.System)
 end
 
 # TODO:
-# - Handle PER-UNIT conversion of variables according to type
+# - ensure proper PER-UNIT conversion of variables according to type (i.e. use `read_XXX` functions)
```
jd-lara, line 300: I think this is ready. This the conversion trait

src/operation/problem_results.jl
```diff
@@ -573,3 +560,101 @@ function read_realized_duals_with_keys(
...
+"""
+function read_realized_aux_variables(res::ProblemResults; kwargs...)
+    return read_realized_aux_variables(
+        res,
+        collect(keys(res.aux_variable_values));
+        kwargs...,
+    )
+end
+
+function read_realized_aux_variables(res::ProblemResults, aux_variables; kwargs...)
```
jd-lara, line 582: These functions require testing

jd-lara (changes requested):
The PR generally speaking looks good but the coverage is low. We need tests.

jd-lara (approved):
Looks great to me.

PR #832, opened 2022-02-11
Updating hybrid time series methods

sourabhdalvi:


src/parameters/update_parameters.jl
```diff
@@ -56,6 +56,37 @@ function update_parameter_values!(
...
+    ::DatasetContainer{DataFrameDataset},
+) where {
+    T <: Union{PJ.ParameterRef, Float64},
+    U <: PSY.AbstractDeterministic,
+    V <: PSY.HybridSystem,
+}
+    initial_forecast_time = get_current_time(model) # Function not well defined for DecisionModels
+    horizon = get_time_steps(get_optimization_container(model))[end]
+    components = get_available_components(V, get_system(model))
+    for component in components
```
jd-lara, line 73: we probably need to loop over the subcomponents here instead of just the Hybrid.

jd-lara (approved):
LGTM, we need to confirm that this PR achieves the same results as the previous one before tagging v015

PR #843, opened 2022-03-10
More cost function fixes

sourabhdalvi:


src/devices_models/devices/common/add_to_expression.jl
```diff
@@ -196,6 +196,34 @@ function add_to_expression!(
...
+}
+    variable = get_variable(container, U(), V)
+    expression = get_expression(container, T(), X)
+    for d in devices, t in get_time_steps(container)
+        name = PSY.get_name(d)
+        bus_number = PSY.get_number(PSY.get_bus(d))
+        _add_to_jump_expression!(
+            expression[bus_number, t],
+            variable[name, t],
+            get_variable_multiplier(U(), d, W()),
```
sourabhdalvi, line 221: @jd-lara I had to add few method like this one for Compact thermal formulation, because here to know the multiplier for On we need the device to for retrieving the Pmin. 
The other methods were using this 
https://github.com/NREL-SIIP/PowerSimulations.jl/blob/9d4f952a25d27b0366cc3e659f894fb6e0443e78/src/devices_models/devices/thermal_generation.jl#L9
but this is the one we need here 
https://github.com/NREL-SIIP/PowerSimulations.jl/blob/9d4f952a25d27b0366cc3e659f894fb6e0443e78/src/devices_models/devices/thermal_generation.jl#L10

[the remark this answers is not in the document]
jd-lara: This seems reasonable.

test/test_device_thermal_generation_constructors.jl
```diff
@@ -801,5 +801,5 @@ end
     c_sys5 = PSB.build_system(PSITestSystems, "c_sys5_pglib")
     model = DecisionModel(MockOperationProblem, DCPPowerModel, c_sys5)
     mock_construct_device!(model, device_model; built_for_recurrent_solves=true)
-    moi_tests(model, true, 240, 0, 192, 48, 48, false)
+    moi_tests(model, true, 240, 0, 192, 48, 96, false)
```
jd-lara, line 804: @sourabhdalvi why are these tests changing value?

test/test_device_thermal_generation_constructors.jl
```diff
@@ -617,7 +617,7 @@ end
         initialize_model=false,
     )
     @test build!(UC; output_dir=mktempdir(cleanup=true)) == PSI.BuildStatus.BUILT
-    moi_tests(UC, false, 32, 0, 8, 4, 10, true)
+    moi_tests(UC, false, 32, 0, 8, 4, 14, true)
```
jd-lara, line 620: Here too, why did the test resulted in more constraints?

src/devices_models/devices/common/add_to_expression.jl
```diff
@@ -283,6 +311,33 @@ function add_to_expression!(
     return
 end
 
+function add_to_expression!(
```
jd-lara, line 314: same here.

src/devices_models/devices/common/objective_functions.jl
```diff
@@ -525,15 +525,15 @@ function _add_pwl_term!(
         # First element of the return is the average cost at P_min.
         # Shouldn't be passed for convexity check
         is_convex = _slope_convexity_check(slopes[2:end])
-        break_points = PSY.get_breakpoint_upperbounds(data) ./ base_power
+        break_points = map(x -> last(x), data) ./ base_power
```
jd-lara, line 528: how is this different than the other breakpoints

test/test_device_thermal_generation_constructors.jl
```diff
@@ -634,7 +634,7 @@ end
     moi_tests(UC, false, 32, 0, 8, 4, 14, true)
     # Cbc can have reliability issues with SoS. The objective function target in the this
     # test was calculated with CPLEX do not change if Cbc gets a bad result
-    psi_checksolve_test(UC, [MOI.OPTIMAL], 1286.0, 10.0)
+    psi_checksolve_test(UC, [MOI.OPTIMAL], 8500.0, 10.0)
```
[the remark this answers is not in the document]
sourabhdalvi: The thing you ran in CPLEX was wrong formulation, so that results doesn't make sense and I constructed these tests myself so the value hasn't been changes since. this is the same values as before in v0.14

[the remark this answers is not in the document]
jd-lara: ok, I ran the test with CPLEX before the PR with the cost function changes. The previous value is the current value in master. We had disabled it because of the segmentation faults in Cbc.

src/devices_models/devices/common/add_to_expression.jl
```diff
@@ -196,6 +196,34 @@ function add_to_expression!(
...
```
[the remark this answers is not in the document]
sourabhdalvi: it would be a union of AbstractCompactUnitCommitment and ThermalCompactDispatch, but no other thermal formulation tries to add On variable in the demand balance equations. So I just left it at AbstractThermalFormulation

src/devices_models/devices/common/objective_functions.jl
```diff
@@ -525,15 +525,15 @@ function _add_pwl_term!(
...
+        if has_container_key(container, OnStatusParameter, T)
```
jd-lara, line 529: This block of code container a relatively complex logic, we should abstract it into a function.

jd-lara:
All the concerns were resolved over the phone.

Issue #834, opened 2022-02-19
Error: sys cache is full but there are no clean entries to pop


ajsauter:
As requested, please find my error report below.

**Describe the bug**
Toward the end of my most recent simulation run, I started to see this recurring error message just before the progress bar for the simulation would reappear:

`Barrier solved model in 44 iterations and 12.27 seconds (7.46 work units)
Optimal objective 2.75945856e+07

Crossover log...

    1217 DPushes remaining with DInf 0.0000000e+00                13s
       0 DPushes remaining with DInf 0.0000000e+00                15s

    4457 PPushes remaining with PInf 5.0101744e-05                15s
    2870 PPushes remaining with PInf 5.0101733e-05                15s
       0 PPushes remaining with PInf 5.0101744e-05                16s

  Push phase complete: Pinf 5.0101744e-05, Dinf 1.5249151e-05     16s

Iteration    Objective       Primal Inf.    Dual Inf.      Time
    5677    2.7594586e+07   5.010174e-05   0.000000e+00     16s
    5678    2.7594586e+07   0.000000e+00   0.000000e+00     18s

Solved with barrier
Solved in 5678 iterations and 31.94 seconds (15.01 work units)
Optimal objective  2.759458626e+07

User-callback calls 15520812, time in user-callback 6.44 sec
┌ Error: sys cache is full but there are no clean entries to pop: PowerSimulations.OptimizationResultCacheKey(:UC, PowerSimulations.ParameterKey{ActivePowerTimeSeriesParameter, RenewableDispatch}(""))
└ @ PowerSimulations C:\Users\A.J. Sauter\.julia\packages\PowerSimulations\wNZsa\src\simulation\model_output_cache.jl:68
┌ Error: sys cache is full but there are no clean entries to pop: PowerSimulations.OptimizationResultCacheKey(:UC, PowerSimulations.ParameterKey{RequirementTimeSeriesParameter, VariableReserve{ReserveUp}}("REG_UP"))
└ @ PowerSimulations C:\Users\A.J. Sauter\.julia\packages\PowerSimulations\wNZsa\src\simulation\model_output_cache.jl:68
┌ Error: sys cache is full but there are no clean entries to pop: PowerSimulations.OptimizationResultCacheKey(:UC, PowerSimulations.ParameterKey{RequirementTimeSeriesParameter, VariableReserve{ReserveUp}}("SPIN"))
[...]

PR #847, opened 2022-03-18
Fix handling of cache-full condition

daniel-thom:
This PR fixes issue #834. This changes the code to detect an upper cache-full limit and then discard clean entries (data that has already been flushed to file) down to a lower limit. I also simplified some of the existing code.


src/simulation/hdf_simulation_store.jl
```diff
@@ -119,6 +116,7 @@ end
 function Base.close(store::HdfSimulationStore)
     flush(store)
     HDF5.close(store.file)
+    empty!(store.cache)
```
daniel-thom, line 119: I added this change because I think we should free the user's memory. Let me know if disagree. We have a different mechanism to cache results in SimulationResults.

src/simulation/hdf_simulation_store.jl
```diff
@@ -110,7 +108,6 @@ function open_store(
         return func(store)
     finally
         if store !== nothing
-            flush(store)
```
daniel-thom, line 113: This was already being called in close.

src/simulation/in_memory_simulation_store.jl
```diff
@@ -172,7 +172,7 @@ function read_result(
     key::OptimizationContainerKey,
     index::EmulationModelIndexType,
 )
-    return read_results(get_em_data(store), key, index)
+    return read_results(get_em_data(store), key, index=index)
```
daniel-thom, line 175: This is unrelated to the cache issue, but was clearly wrong. If you fixed this in a different way in your branch such that this will cause a merge conflict, I can remove it.

src/simulation/simulation.jl
```diff
@@ -843,9 +843,9 @@ function execute!(sim::Simulation; kwargs...)
         error("Simulation status is invalid, you need to rebuild the simulation")
     end
     try
-        open_store(store_type, get_store_dir(sim), "w") do store
-            set_simulation_store!(sim, store)
-            Logging.with_logger(logger) do
+        Logging.with_logger(logger) do
```
daniel-thom, line 846: This was causing some log statements in the store to get dropped.

src/simulation/hdf_simulation_store.jl
jd-lara: I am ok with this. The cache shouldn't be populated after closing.

src/simulation/in_memory_simulation_store.jl
jd-lara: This is ok, I haven't run into this issue.

src/simulation/simulation.jl
jd-lara: ok, I am wondering why?

test/test_simulation_store.jl
```diff
@@ -139,21 +143,23 @@ function _verify_data(expected, store, model, name, time, columns)
...
     )
     variables = Dict(
         PSI.VariableKey(ActivePowerVariable, ThermalStandard) =>
             Dict("cache_priority" => CachePriority.HIGH, "keep_in_cache" => true),
+        PSI.VariableKey(ActivePowerVariable, RenewableDispatch) =>
+            Dict("cache_priority" => CachePriority.HIGH, "keep_in_cache" => true),
         PSI.VariableKey(ActivePowerVariable, InterruptibleLoad) =>
             Dict("cache_priority" => CachePriority.LOW, "keep_in_cache" => false),
+        PSI.VariableKey(ActivePowerVariable, RenewableFix) =>
+            Dict("cache_priority" => CachePriority.LOW, "keep_in_cache" => false),
```
jd-lara, line 162: Are we even using the CachePriority at the moment? Is it worth to remove that in this branch?

jd-lara (commented):
The PR looks good to me. As a suggestion, I think that this is the right time to remove the PriorityCaching if its an idea we aren't using.

src/simulation/simulation.jl
daniel-thom: The `with_logger` causes the simulation logger to be the global logger for its entire code block. The way it was left a gap: any log messages sent by the simulation store in its `close` execution were not covered in that code block, and so were left out of the simulation log file.

test/test_simulation_store.jl
daniel-thom: Sounds good. If we ever need it, we can easily add it back.

jd-lara (approved):
This is great, thanks for fixing the cache implementation.

PR #848, opened 2022-03-19
print improvements

jd-lara:


src/utils/printing.jl
```diff
@@ -1,82 +1,353 @@
...
+function Base.show(io::IO, container::OptimizationContainer)
+    show(io, get_jump_model(container))
+end
+
+function Base.show(io::IO, ::MIME"text/plain", input::Union{ServiceModel, DeviceModel})
+    _show_method(io, input, :auto)
+end
+
+function Base.show(io::IO, ::MIME"text/html", input::Union{ServiceModel, DeviceModel})
+    _show_method(io, input, :html)
```
daniel-thom, line 10: Anytime we set the backend to `:html` we need to set `standalone=false` in order to prevent PrettyTables from creating a full HTML document. The default value may be changing in the next release, but we should be explicit. It can cause issues for the package Dheepak is using for documentation.

jd-lara: I tried to pass the standalone kwarg and pretty tables threw an error about kwarg not valid

daniel-thom: I forgot to mention that it requires a newer version of PrettyTables. You probably have 1.1 or 1.2. We bumped the version in PSY and IS to
```
PrettyTables = "^1.3"
```

Project.toml
```diff
@@ -41,6 +42,7 @@ MathOptInterface = "1"
 ParameterJuMP = "~v0.4"
 PowerModels = "~0.19"
 PowerSystems = "1.15"
+PrettyTables = "1"
```
daniel-thom, line 45: I think this needs to be `"^1.3"`.

daniel-thom (approved):

PR #849, opened 2022-03-20
Cb/results fixes

claytonpbarrows:


jd-lara (approved):

jd-lara:
Will merge without tests due to the limited availability of free MILP solvers that support duals.

PR #853, opened 2022-03-24
Minor Bug fix in Aux variable calculation

sourabhdalvi:


jd-lara (commented):
Can you check we are not making the same mistake in other places?

sourabhdalvi:
> Can you check we are not making the same mistake in other places?

we do use `get_components` for bus and reserves, should I change that as well ?

jd-lara:
> > Can you check we are not making the same mistake in other places?
> 
> we do use `get_components` for bus and reserves, should I change that as well ?

yes

jd-lara:
tests are expected to fail since there isn't a tag of PJ and PowerModels

PR #854, opened 2022-03-31
correct production cost for compact

jd-lara:


sourabhdalvi (approved):
LGTM, just add the NoMin fix as well before you merge

PR #855, opened 2022-04-06
Cb/obj fcn bugs

claytonpbarrows:
- fixes a bug with pwl objective functions
- adds plural read results functions

### DIFF
--- a/src/devices_models/devices/common/objective_functions.jl
+++ b/src/devices_models/devices/common/objective_functions.jl
@@ -756,8 +756,12 @@ function _add_pwl_term!(
     ::U,
     ::V,
 ) where {T <: PSY.ThermalGen, U <: VariableType, V <: ThermalDispatchNoMin}
+    multiplier = objective_function_multiplier(U(), V())
+    resolution = get_resolution(container)
+    dt = Dates.value(Dates.Second(resolution)) / SECONDS_IN_HOUR
     component_name = PSY.get_name(component)
     @debug "PWL cost function detected for device $(component_name) using $V"
+    base_power = get_base_power(container)
     slopes = PSY.get_slopes(data)
     if any(slopes .< 0) || !_slope_convexity_check(slopes[2:end])
         throw(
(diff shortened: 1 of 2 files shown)


src/devices_models/devices/common/objective_functions.jl
```diff
@@ -756,8 +756,12 @@ function _add_pwl_term!(
     ::U,
     ::V,
 ) where {T <: PSY.ThermalGen, U <: VariableType, V <: ThermalDispatchNoMin}
+    multiplier = objective_function_multiplier(U(), V())
```
jd-lara, line 759: Was this an oversight?

jd-lara (approved):
LGMT, if you can clarify the change to the cost function I would appreciate it.

src/devices_models/devices/common/objective_functions.jl
claytonpbarrows: yeah it is used down below, but was never defined in this function. There were several variables in this situation.

src/operation/problem_results.jl
```diff
@@ -409,6 +409,13 @@ function read_variables(
     return Dict(encode_key_as_string(k) => v for (k, v) in result_values)
 end
 
+"""
+Return the values for all variables.
+"""
+function read_variables(res::IS.Results)
```
daniel-thom, line 415: Consistency concerns:
1. We have this function for ProblemResults. It accepts kwargs for parameters like start_time. We don't have it for the others. We could delete this set of functions.
```
function read_variables(res::ProblemResults; kwargs...)
```
2. We could add `kwargs...` to all the new functions and forward them. We would need to test all combinations. It could be messy. It's probably easier to say that these functions return all values for all variables. I would probably delete the functions mentioned in (1) but am not certain.

jd-lara: I think is best to remove the methods and see what happens, some of those functions seem too generic and can be confusing.

src/devices_models/devices/common/objective_functions.jl
sourabhdalvi: Thanks for making these fixes, I completely forgot about them after I made the changes to SIIPExamples

PR #856, opened 2022-04-08
Remove read_variables from ProblemResults

daniel-thom:
This leaves only one set of plural read_X methods. Those methods do not accept kwargs.


daniel-thom:
The remaining methods accept `IS.Results`. Isn't essentially type piracy from IS? Didn't we used to have a PSI subtype of IS.Results?

jd-lara (approved):

PR #864, opened 2022-04-23
ff ramps fix

jd-lara:
This PR addresses an issue with semicontinuous FF when there are ramp constraints.

### DIFF
--- a/src/devices_models/devices/common/rateofchange_constraints.jl
+++ b/src/devices_models/devices/common/rateofchange_constraints.jl
@@ -97,7 +97,7 @@ function add_linear_ramp_constraints!(
         )
         con_down[name, 1] = JuMP.@constraint(
             container.JuMPmodel,
-            ic_power - expr_dn[name, 1] <= ramp_limits.down * minutes_per_period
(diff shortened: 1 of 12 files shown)


src/devices_models/devices/common/rateofchange_constraints.jl
```diff
@@ -97,7 +97,7 @@ function add_linear_ramp_constraints!(
         )
         con_down[name, 1] = JuMP.@constraint(
             container.JuMPmodel,
-            ic_power - expr_dn[name, 1] <= ramp_limits.down * minutes_per_period
+            ic_power - expr_dn[name, 1] >= -1 * ramp_limits.down * minutes_per_period
```
sourabhdalvi, line 100: can you remind me why are we doing this ? This would basically mean  ramp >= expr_dn - ic_power (which looks like a ramp up constraint )

sourabhdalvi (approved):
Mostly looks good to me, other than the ramp constraint

src/devices_models/devices/common/rateofchange_constraints.jl
jd-lara: because the expressions are all ```var + min*param``` so they we need to account of the distribution of the minus sign

PR #869, opened 2022-07-13
Jd/use poi v2

jd-lara:
This is a PR with an initial implementation of POI as a substitute for ParameterJuMP. 

Critically, it requires using `check_numerical_bounds=false` since the option in POI to save the original constraints is set to false. Since this update is motivated by scalability concerns with ParameterJuMP, the option not to save the original constraints is required but breaks the bounds checking. 

More performance testing is necessary before merging. 

This PR is also going to require some way of evaluating the equivalent to `JuMP.value(parameter)` otherwise it is going to be a massively disruptive change in the processing of the optimizer output.


jd-lara:
Progress in this PR is blocked by https://github.com/jump-dev/ParametricOptInterface.jl/issues/119

jd-lara:
progress is currently blocked by https://github.com/jump-dev/ParametricOptInterface.jl/pull/120/files

jd-lara:
The branch was merged into another version and close the PR since there are still blocking issues.

PR #874, opened 2022-08-09
Add ability to split simulations into partitions and run them in parallel

daniel-thom:
This PR does introduce one behavioral change that affects existing simulations. In the current code if you repeatedly run a simulation, you will get a directory structure like this:
```
output/my_simulation/1
output/my_simulation/2
output/my_simulation/3
```
This is changed to
```
output/my_simulation
output/my_simulation-2
output/my_simulation-3
```
The reason for the change is to reduce the number of directories that get created in a partitioned simulation. Ideally, we will have the same directory structure regardless of the execution method.

Unimplemented features:
1. This code leaves all files from partitioned simulations. We could remove the simulation_store.h5 files. I'm leaving these in until we're confident that everything is working correctly. This could be a configuration flag.


jd-lara:
@daniel-thom any idea why the windows CI is failing?

daniel-thom:
> @daniel-thom any idea why the windows CI is failing?

I was running three subprocesses in parallel. It was likely overwhelming the virtual machine.

PR #889, opened 2022-12-06
WIP: test using fixed variables instead of ParameterJuMP

odow:
This PR is still a WIP. It's untested etc. Trying some things locally, but I thought I'd push this here so you can see the direction.

It is likely that the overhead of the extra decision variables is worse than ParameterJuMP or POI, but we should at least test and benchmark this approach before considering more complicated ones.

x-ref https://github.com/NREL-SIIP/PowerSimulations.jl/issues/800#issuecomment-1340076963


odow:
@jd-lara try this out on a larger example. It seems to work for the one you sent me. I don't see much of a difference in the timings (at all). Is there a parameter I need to set to use PJ?

jd-lara:
@odow thanks for the fix. I merged into the poi branch since that one had a lot of the tests fixed to remove ParameterJuMP.

Issue #886, opened 2022-11-29
implement an update model for models that need to be re-built. `get_requires_rebuild(model) == true`


jd-lara:
This feature needs to be implemented as follows 

```Julia
function update_model!(model::OperationModel, sim::Simulation)
    if get_requires_rebuild(model)
        # initial_conditions_data =  make_initial_conditions_data()
        # Reset and flush of optimization container()
        # Set initial conditions in container
        # build_impl!(model.container, template, system)
    else
        update_model!(model, get_simulation_state(sim), get_ini_cond_chronology(sim))
    end
    return
end
```

Items: 

- [ ] Implement a make initial conditions data object the returns an `InitialConditionsData` populated correctly with the values for the parameters and variables. 
- [ ] Implement a method to flush the optimization problem from the optimization container. Hint: use `clear!` from Dicts. 
- [ ] Modify the implementation of the FeedForwards to check if the parameters for the FeedForwards exist in `InitialConditionsData` to initialize the values of the parameters for the FeedForwards.

PR #895, opened 2022-12-08
Fixing bug in creating initial conditions for Energy variables

sourabhdalvi:
The code would always use system data for any type of energy variables, now its use initial condition data if available.
This will be useful for enabling simulations with model re-build feature


jd-lara (approved):

jd-lara (commented):
@sourabhdalvi this is a fix for the rebuild but not a systematic one. The code path you are changing here is the build one not the update one.

claytonpbarrows:
#886

jd-lara:
@sourabhdalvi I think that your branch is fixing a slightly different problem which is the fact that this field should be initialized from the initial conditions data if it is available. But during the simulation the initial conditions are always updated from the store so this code path never gets called. So, unless you get a formulation that solves for the initial energy this code path will always use the default data source.

jd-lara:
After reviewing the effects of this PR, it was reversed in 3af8e2c48b733d97b758ac21b3f8133bea1fa346 since it ignores the initial condition in the data.

PR #903, opened 2022-12-21
Enable simulation with rebuild

### LINKED ISSUE #886: implement an update model for models that need to be re-built. `get_requires_rebuild(model) == true`

jd-lara:
This is work in progress for enabling rebuild and close #886. 

Some notes for the pending items. 

1. if the model runs with rebuild parameters can't be VariableRefs otherwise the associated JuMP models will be flushed and the variable will be "orphan". 
2. Parameters for the FF's need to be kept too.


src/operation/operation_model_interface.jl
```diff
@@ -15,12 +15,12 @@ get_name(model::OperationModel) = model.name
 get_store(model::OperationModel) = model.store
 is_synchronized(model::OperationModel) = is_synchronized(get_optimization_container(model))
 
-function get_requires_rebuild(model::OperationModel)
+function get_rebuild_model(model::OperationModel)
     sim_info = get_internal(model).simulation_info
     if sim_info === nothing
-        error("Model not part of a simulation")
+        error("Model not part of a simulation")ß
```
daniel-thom, line 21: Is the character at the end a typo?

jd-lara: I have no idea how that character arrived there

src/parameters/add_parameters.jl
```diff
@@ -41,6 +64,9 @@ function add_parameters!(
     U <: Union{Vector{D}, IS.FlattenIteratorWrapper{D}},
     W <: AbstractDeviceFormulation,
 } where {D <: PSY.HybridSystem}
+    if get_rebuild_model(get_settings(container)) && has_container_key(container, T, D)
```
daniel-thom, line 67: Is there documentation that explains this to model developers?

daniel-thom: And is there a way to automate this so that it doesn't have to live everywhere?

jd-lara: not yet :\

PR #905, opened 2022-12-29
ps power control

### LINKED ISSUE #894: Phase Shifters control in DC power cases

jd-lara:
closes #894 

This PR adds a formulation compatible with StandardPTDF and DCPPowerFlow for the phase control in phase shifters. Note that there is no need to implement PDF matrices here.


jd-lara:
This PR needs some additional testing. The new HVDCDispatch formulation won't work with AC Models which a regression

PR #918, opened 2023-01-25
Adding an initial branch flow slacks implementation for PTDF

sourabhdalvi:


src/devices_models/devices/AC_branches.jl
```diff
@@ -173,6 +197,36 @@ function add_constraints!(
...
+    U <: VariableType,
+    V <: PSY.ACBranch,
+    W <: AbstractBranchFormulation,
+    X <: PM.AbstractPowerModel,
+}
+    array = get_variable(container, U(), V)
+    if get_use_slacks(model)
+        lower_bound_slack = get_variable(container, BoundSlackLowerBound(), V)
+        upper_bound_slack = get_variable(container, BoundSlackUpperBound(), V)
+        array_upper = array .- upper_bound_slack
```
jd-lara, line 217: we should unroll this to make it clear. The `.+` and `.-` are very inefficient.

PR #924, opened 2023-02-03
Optimize handling of shared time series arrays

### LINKED ISSUE #904: reuse timeseries UUIDs in parameter creation
In large power system cases the time series arrays are shared between multiple objects. However, in power simulation we create an individual parameter per load and per renewable generator to update to the same value. 

We want to implement a map in PSI that is able to avoid re-creating the parameter (VariableRef) for each object unnecessarily. However, to avoid breaking results processing and other utilities the parameter container arrays should still use the original object as the index.

daniel-thom:
closes #904


src/services_models/reserves.jl
```diff
@@ -105,19 +105,17 @@ function add_constraints!(
     requirement = PSY.get_requirement(service)
     jump_model = get_jump_model(container)
     if parameters
-        container =
+        param_container =
             get_parameter(container, RequirementTimeSeriesParameter(), SR, service_name)
-        param = get_parameter_array(container)
+        param = get_parameter_column_refs(param_container, service_name)
```
daniel-thom, line 110: Among other things, something is broken here or in the way this container is initialized.

jd-lara:
needs additional QA/QC and a couple of syntactic sugar improvements before merging

jd-lara (approved):

PR #945, opened 2023-03-16
Jd/support virtual ptdf

jd-lara:
This PR allows for the use of virtual PTDF in the PTDF model 

cc. @alefcastelli


src/core/network_model.jl
```diff
@@ -28,7 +28,7 @@ thermal_gens = NetworkModel(StandardPTDFModel, ptdf = ptdf_array),
 """
 mutable struct NetworkModel{T <: PM.AbstractPowerModel}
     use_slacks::Bool
-    PTDF_matrix::Union{Nothing, PTDF}
+    PTDF_matrix::Union{Nothing, PTDF, VirtualPTDF}
```
daniel-thom, line 31: If PTDF and VirtualPTDF share the same supertype, can you put that type here? That way, if you add another subtype later, you won't have to change this definition.

jd-lara: resolved

daniel-thom (approved):

PR #949, opened 2023-03-25
Jd/multi area coopperplate

jd-lara:
Requires https://github.com/NREL-SIIP/PowerNetworkMatrices.jl/pull/22 to pass tests


jd-lara:
This PR ended up requiring some additional refactoring of the model constructors

PR #953, opened 2023-03-30
use variable bounds in pwl compact check

jd-lara:


rodrigomha (approved):
LGTM as long test passes.

It seems that for each HybridSystem will warn that does not accept compact form?

PR #954, opened 2023-04-14
Jd/fix value ff fixes

jd-lara:
This PR addresses several shortcomings when simulating systems that might have variables with different time_step length and also fixes some outstanding limitations in the use of feed forwards for models without variables that share keys.


src/utils/jump_utils.jl
```diff
@@ -220,12 +220,20 @@ function _calc_dimensions(
     # 2. Read data for one execution for all devices.
     # This will ensure that data on disk is contiguous in both cases.
     if length(ax) == 1
-        dims = (horizon, 1, num_rows)
+        if length(ax[1]) != horizon
+            @debug "$(encode_key_as_string(key)) has length $(length(ax[1])). Different than horizon $horizon."
```
daniel-thom, line 224: Are debug messages appropriate here?

jd-lara: I can change it to an Info, but I assumed the user would understand the behavior.

PR #955, opened 2023-04-14
[WIP] Update docs API

rodrigomha:
Restructure of the Docs API to manually specify which Types and Methods will be used.

- [x] Device Models
- [x] Decision Models
- [x] Emulation Models
- [x] Simulation Models
- [x] Variables
- [x] Parameters
- [x] Constraints


rodrigomha:
@jd-lara @claytonpbarrows 
You can check the WIP here: https://nrel-siip.github.io/PowerSimulations.jl/previews/PR955/api/PowerSimulations/

src/operation/decision_model.jl
```diff
@@ -280,7 +280,19 @@ end
...
 """
-Implementation of build for any DecisionProblem
+Build the Decision Model based on the specified DecisionProblem.
+
+# Arguments
+
+  - `model::DecisionModel{<:DecisionProblem}`: DecisionModel object
+  - `output_dir::String`: Output directory for results
+
+# Accepted Key Words
```
daniel-thom, line 290: I don't think that there should be a separate section called "Accepted Key Words." You might check the latest docstring guidelines, but I suspect this should be under "Arguments".

rodrigomha: It seems that you are right given this: https://docs.julialang.org/en/v1/manual/documentation/

```
For simple functions, it is often clearer to mention the role of the arguments directly in the description of the function's purpose. An argument list would only repeat information already provided elsewhere. However, providing an argument list can be a good idea for complex functions with many arguments (in particular keyword arguments). In that case, insert it after the general description of the function, under an # Arguments header, with one - bullet for each argument. The list should mention the types and default values (if any) of the arguments:
```

On another note, do we keep the main method at the top of the docstring or should we just keep the list of methods at the end provided by DocStringExtensions?

rodrigomha: @jd-lara @daniel-thom I prefer to keep the general method on top of the dosctring instead of listing all the methods at the end based on docextensions:

```
@template DEFAULT = """
                    $(TYPEDSIGNATURES)
                    $(DOCSTRING)
                    $(METHODLIST)
                    """
```

Something that happens with this is that it does not look clean, like this:

image

My vote is to remove the MethodList, and manually set-up the main method at the beginning. At least that is how they handle in JuMP

rodrigomha: Yeah, I removed the METHODLIST, it does not look good and is not being used in PSY. With that, all methods will autogenerate the typed signature.

However, for the types, such as DeviceModel or DecisionModel it will not work.

So what about this plan:
 - Remove METHODLIST
 - Use DocStringExtensions for typed signatures for methods (such as `build!` or `solve!`)
 - Manually create the signature for relevant Types (such as `DeviceModel` or `DecisionModel`).

What do you think?

src/core/constraints.jl
```diff
@@ -82,8 +82,8 @@ struct RegulationLimitsConstraint <: ConstraintType end
 struct RequirementConstraint <: ConstraintType end
 struct ReserveEnergyCoverageConstraint <: ConstraintType end
 struct ReservePowerConstraint <: ConstraintType end
-struct SACEPIDAreaConstraint <: ConstraintType end
-struct StartTypeConstraint <: ConstraintType end
+struct SACEPIDAreaConstraint <: ConstraintType end #not being used
```
jd-lara, line 85: This should be used in the AGC model.

src/operation/decision_model.jl
jd-lara: removing Method list is ok. We will end up with an explosion of methods otherwise

rodrigomha:
I marked this for review, but I still need to add docstrings for most variables constraints and parameters

src/core/device_model.jl
```diff
@@ -16,21 +16,33 @@ function _check_device_formulation(
...
--`::Type{B}`: Abstract Device Formulation
-
-# Accepted Key Words
-
-  - `feedforward::Array{<:AbstractAffectFeedforward}` : use to pass parameters between models
+  - `::Type{D} where D<:PSY.Device`: Power System Device Type
+  - `::Type{B} where B<:AbstractDeviceFormulation`: Abstract Device Formulation
+  - `feedforward::Array{<:AbstractAffectFeedforward} = Vector{AbstractAffectFeedforward}()` : use to pass parameters between models
+  - `use_slacks::Bool = false` : Add slacks to the device model
+  - `duals::Vector{DataType} = Vector{DataType}()`: use to pass constraint type to calculate the duals
```
jd-lara, line 38: Mention that the data type has to be a constraint type.

src/core/device_model.jl
```diff
@@ -16,21 +16,33 @@ function _check_device_formulation(
...
--`::Type{D}`: Power System Device Type
--`::Type{B}`: Abstract Device Formulation
-
-# Accepted Key Words
-
-  - `feedforward::Array{<:AbstractAffectFeedforward}` : use to pass parameters between models
+  - `::Type{D} where D<:PSY.Device`: Power System Device Type
+  - `::Type{B} where B<:AbstractDeviceFormulation`: Abstract Device Formulation
+  - `feedforward::Array{<:AbstractAffectFeedforward} = Vector{AbstractAffectFeedforward}()` : use to pass parameters between models
+  - `use_slacks::Bool = false` : Add slacks to the device model
```
jd-lara, line 37: Mention, the slack implementation is model dependent. Not all models have slack implementations

src/core/device_model.jl
```diff
@@ -16,21 +16,33 @@ function _check_device_formulation(
...
+  - `duals::Vector{DataType} = Vector{DataType}()`: use to pass constraint type to calculate the duals
+  - `time_series_names::Dict{Type{<:TimeSeriesParameter}, String} = get_default_time_series_names(D, B)` : use to specify time series names associated to the device`
+  - `attributes::Dict{String, Any} = get_default_attributes(D, B)` : use to specify attributes to the device
 
 # Example
-
-thermal_gens = DeviceModel(ThermalStandard, ThermalBasicUnitCommitment),
+```julia
```
jd-lara, line 43: Add example with time series

src/operation/decision_model.jl
```diff
@@ -4,47 +4,46 @@ Default PowerSimulations Operation Problem Type
...
+  - `name = nothing`: name of model, string or symbol; defaults to the type of template converted to a symbol.
+  - `optimizer::Union{Nothing,MOI.OptimizerWithAttributes} = nothing` : The optimizer does
+    not get serialized. Callers should pass whatever they passed to the original problem.
+  - `horizon::Int = UNSET_HORIZON`: Manually specify the length of the forecast Horizon
+  - `warm_start::Bool = true`: True will use the current operation point in the system to initialize variable values. False initializes all variables to zero. Default is true
+  - `system_to_file::Bool = true:`: True to create a copy of the system used in the model.
+  - `initialize_model::Bool = true`: Option to decide to initialize the model or not.
+  - `initialization_file::String = ""`: TODO
+  - `deserialize_initial_conditions::Bool = false`: Option to deserialize conditions
+  - `export_pwl_vars::Bool = false`: True to export all the pwl intermediate variables. It can slow down significantly the solve time.
```
jd-lara, line 30:
```suggestion
  - `export_pwl_vars::Bool = false`: True to export all the pwl intermediate variables. It can slow down significantly the build and solve time.
```

src/operation/decision_model.jl
```diff
@@ -4,47 +4,46 @@ Default PowerSimulations Operation Problem Type
...
-  - `model::DecisionModel`: The operation model containing the model type, built JuMP model, Power
-    Systems system.
+  - `name = nothing`: name of model, string or symbol; defaults to the type of template converted to a symbol.
+  - `optimizer::Union{Nothing,MOI.OptimizerWithAttributes} = nothing` : The optimizer does
+    not get serialized. Callers should pass whatever they passed to the original problem.
+  - `horizon::Int = UNSET_HORIZON`: Manually specify the length of the forecast Horizon
+  - `warm_start::Bool = true`: True will use the current operation point in the system to initialize variable values. False initializes all variables to zero. Default is true
+  - `system_to_file::Bool = true:`: True to create a copy of the system used in the model.
+  - `initialize_model::Bool = true`: Option to decide to initialize the model or not.
+  - `initialization_file::String = ""`: TODO
```
jd-lara, line 28: This allows to pass pre-existing initialization values to avoid the solution of an optimization problem to find feasible initial conditions.

src/operation/decision_model.jl
```diff
@@ -4,47 +4,46 @@ Default PowerSimulations Operation Problem Type
...
+  - `deserialize_initial_conditions::Bool = false`: Option to deserialize conditions
+  - `export_pwl_vars::Bool = false`: True to export all the pwl intermediate variables. It can slow down significantly the solve time.
+  - `allow_fails::Bool = false`: True to allow the simulation to continue even if the optimization step fails. Use with care.
+  - `optimizer_solve_log_print::Bool = false`: Uses JuMP.unset_silent() to print the optimizer's log. By default all solvers are set to MOI.Silent()
+  - `detailed_optimizer_stats::Bool = false`: True to save detailed optimizer stats log.
+  - `calculate_conflict::Bool = false`: True to use solver to calculate conflicts for infeasible problems. Only specific solvers are able to calculate conflicts.
+  - `direct_mode_optimizer::Bool = false`: True to use the solver in direct mode. Creates a [JuMP.direct_model](https://jump.dev/JuMP.jl/dev/reference/models/#JuMP.direct_model).
+  - `store_variable_names::Bool = false`: True to store variable names in optimization model.
```
jd-lara, line 36:
```suggestion
  - `store_variable_names::Bool = false`: True to store variable names in optimization model.
```
```suggestion
  - `store_variable_names::Bool = false`: True to store variable names in optimization model. Decreases the build times. 
```

src/operation/decision_model.jl
```diff
@@ -4,47 +4,46 @@ Default PowerSimulations Operation Problem Type
...
+  - `rebuild_model::Bool = false`: TODO
```
jd-lara, line 37:
```suggestion
  - `rebuild_model::Bool = false`: It will force the rebuild of the underlying JuMP model with each call to update the model. It increases solution times, use only if the model can't be updated in memory. 
```

src/operation/decision_model.jl
```diff
@@ -4,47 +4,46 @@ Default PowerSimulations Operation Problem Type
...
+  - `initial_time::Dates.DateTime = UNSET_INI_TIME`: Initial Time for the model solve.
```
jd-lara, line 38:
```suggestion
  - `initial_time::Dates.DateTime = UNSET_INI_TIME`: Initial Time for the model solve. Default takes the initial time in the PowerSystems.System time series
```

src/operation/decision_model.jl
```diff
@@ -136,41 +135,20 @@ function DecisionModel{M}(
...
-    kwargs...) where {M <: DecisionProblem}
-
 This builds the optimization problem of type M with the specific system and template
 
 # Arguments
 
   - `::Type{M} where M<:DecisionProblem`: The abstract operation model type
-  - `template::ProblemTemplate`: The model reference made up of transmission, devices,
-    branches, and services.
+  - `template::ProblemTemplate`: The model reference made up of transmission, devices, branches, and services.
```
jd-lara, line 143:
```suggestion
  - `template::ProblemTemplate`: The model reference comprised of transmission, devices, branches, and services.
```

daniel-thom: Most people prefer `made up of` over `comprised of`. And this needs an `is`.

src/operation/emulation_model.jl
```diff
@@ -504,10 +486,7 @@ Default solve method for an EmulationModel used inside of a Simulation. Solves p
   - `model::OperationModel`: operation model
   - `start_time::Dates.DateTime`: Initial Time of the simulation step in Simulation time.
   - `store::SimulationStore`: Simulation output store
-
-# Accepted Key Words
-
-  - `exports`: realtime export of output. Use wisely, it can have negative impacts in the simulation times
+  - `exports = nothing`: realtime export of output. Use wisely, it can have negative impacts in the simulation times
```
daniel-thom, line 489: Does this really slow things down significantly? I would expect data storage size to be a bigger issue.

src/simulation/simulation_sequence.jl
```diff
@@ -192,12 +192,49 @@ function _attach_feedforwards(models::SimulationModels, feedforwards)
...
+        models::SimulationModels,
+        feedforward::Dict{String, Vector{<:AbstractAffectFeedforward}}
+        ini_cond_chronology::InitialConditionChronology
+    )
+
+Construct the simulation sequence between decision and emulation models.
+
+# Arguments
+
+  - `models::SimulationModels`: Vector of decisions and emulation models.
```
daniel-thom, line 206: Why are you always adding whitespace before the dashes? I'm sure it doesn't matter, but I don't think that is standard practice.

src/simulation/simulation_sequence.jl
```diff
@@ -192,12 +192,49 @@ function _attach_feedforwards(models::SimulationModels, feedforwards)
...
+  - `feedforward = Dict{String, Vector{<:AbstractAffectFeedforward}}()`: Optional dictionary to specify how information
```
daniel-thom, line 207: Is this effort covering feedforwards? If so, this would be a good place to refer to documentation that describes it in detail.

src/operation/emulation_model.jl
```diff
@@ -4,44 +4,45 @@ Default PowerSimulations Emulation Problem Type
...
-  - `model::EmulationModel`: The Emulation model containing the model type, built JuMP model, Power
-    Systems system.
-
+  - `name = nothing`: name of model, string or symbol; defaults to the type of template converted to a symbol.
+  - `optimizer::Union{Nothing,MOI.OptimizerWithAttributes} = nothing` : The optimizer does
+    not get serialized. Callers should pass whatever they passed to the original problem.
+  - `warm_start::Bool = true`: True will use the current operation point in the system to initialize variable values. False initializes all variables to zero. Default is true
+  - `system_to_file::Bool = true:`: True to create a copy of the system used in the model.
+  - `initialize_model::Bool = true`: Option to decide to initialize the model or not.
+  - `initialization_file::String = ""`: TODO
```
jd-lara, line 27: see above

src/operation/decision_model.jl
```diff
@@ -359,15 +344,17 @@ keyword arguments to that function.
 # Arguments
 
   - `model::OperationModel = model`: operation model
-  - `optimizer::MOI.OptimizerWithAttributes`: The optimizer that is used to solve the model
-  - `export_problem_results::Bool`: If true, export ProblemResults DataFrames to CSV files.
-  - `serialize::Bool`: If true, serialize the model to a file to allow re-execution later.
+  - `export_problem_results::Bool = false`: If true, export ProblemResults DataFrames to CSV files.
```
jd-lara, line 347:
```suggestion
  - `export_problem_results::Bool = false`: If true, export ProblemResults DataFrames to CSV files. Reduces solution times. 
```

src/operation/emulation_model.jl
```diff
@@ -4,44 +4,45 @@ Default PowerSimulations Emulation Problem Type
...
+  - `deserialize_initial_conditions::Bool = false`: Option to deserialize conditions
+  - `export_pwl_vars::Bool = false`: True to export all the pwl intermediate variables. It can slow down significantly the solve time.
+  - `allow_fails::Bool = false`: True to allow the simulation to continue even if the optimization step fails. Use with care.
+  - `calculate_conflict::Bool = false`: True to use solver to calculate conflicts for infeasible problems. Only specific solvers are able to calculate conflicts.
+  - `optimizer_solve_log_print::Bool = false`: Uses JuMP.unset_silent() to print the optimizer's log. By default all solvers are set to MOI.Silent()
+  - `detailed_optimizer_stats::Bool = false`: True to save detailed optimizer stats log.
+  - `direct_mode_optimizer::Bool = false`: True to use the solver in direct mode. Creates a [JuMP.direct_model](https://jump.dev/JuMP.jl/dev/reference/models/#JuMP.direct_model).
+  - `store_variable_names::Bool = false`: True to store variable names in optimization model.
+  - `rebuild_model::Bool = false`: TODO
```
jd-lara, line 36: See above

src/operation/emulation_model.jl
jd-lara: It depends on the size of the writes.

src/simulation/simulation_sequence.jl
jd-lara: agreed. We should check that the FeedForwards have correct docstrings

src/simulation/simulation_sequence.jl
jd-lara: This is done by the formatter

jd-lara (approved):

PR #958, opened 2023-04-18
use model type for name

jd-lara:


src/operation/decision_model.jl
```diff
@@ -62,7 +62,7 @@ mutable struct DecisionModel{M <: DecisionProblem} <: OperationModel
         name = nothing,
     ) where {M <: DecisionProblem}
         if name === nothing
-            name = Symbol(typeof(template))
+            name = Symbol("$M")
```
daniel-thom, line 65: Nit: no need to convert to a string first. Second, you may want `nameof(M)` because that is guaranteed to strip the parent module. I'm not sure what `Symbol` does.

daniel-thom (approved):

PR #959, opened 2023-04-18
minor bug fix in hydro multiplier

sourabhdalvi:
As Jarred had reported some issue with the Reservoir Budget formulation which initial was built for HydroEnergyReservoir, this should make sure it works for HydroDispatch and uses the correct multiplier.

### DIFF
--- a/src/devices_models/devices/hydro_generation.jl
+++ b/src/devices_models/devices/hydro_generation.jl
@@ -77,7 +77,8 @@ get_variable_upper_bound(::EnergySurplusVariable, d::PSY.HydroGen, ::AbstractHyd
 get_variable_lower_bound(::EnergySurplusVariable, d::PSY.HydroGen, ::AbstractHydroFormulation) = - PSY.get_storage_capacity(d)
 get_variable_lower_bound(::EnergySurplusVariable, d::PSY.HydroPumpedStorage, ::AbstractHydroFormulation) = - PSY.get_storage_capacity(d).up
 ########################### Parameter related set functions ################################
-get_multiplier_value(::EnergyBudgetTimeSeriesParameter, d::PSY.HydroGen, ::AbstractHydroFormulation) = PSY.get_storage_capacity(d)
+get_multiplier_value(::EnergyBudgetTimeSeriesParameter, d::PSY.HydroGen, ::AbstractHydroFormulation) = PSY.get_max_active_power(d)
+get_multiplier_value(::EnergyBudgetTimeSeriesParameter, d::PSY.HydroEnergyReservoir, ::AbstractHydroFormulation) = PSY.get_storage_capacity(d)
 get_multiplier_value(::EnergyTargetTimeSeriesParameter, d::PSY.HydroGen, ::AbstractHydroFormulation) = PSY.get_storage_capacity(d)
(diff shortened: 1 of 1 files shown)


jd-lara:
I just saw this issue also and have an open PR fixing some things here too. Good catch

src/devices_models/devices/hydro_generation.jl
```diff
@@ -425,7 +414,7 @@ function add_constraints!(
     ::Type{EnergyBalanceConstraint},
     devices::IS.FlattenIteratorWrapper{V},
     model::DeviceModel{V, W},
-    ::NetworkModel{X},
```
jd-lara, line 428: Why did this change back?

src/devices_models/devices/hydro_generation.jl
```diff
@@ -351,12 +348,8 @@ function add_constraints!(
     U::Type{<:Union{VariableType, ExpressionType}},
     devices::IS.FlattenIteratorWrapper{V},
     model::DeviceModel{V, W},
-    ::NetworkModel{X},
```
jd-lara, line 354: These changes to the NetworkModel passing can't be merged

src/devices_models/devices/hydro_generation.jl
```diff
@@ -549,7 +536,7 @@ function add_constraints!(
     ::Type{EnergyCapacityDownConstraint},
     devices::IS.FlattenIteratorWrapper{V},
     model::DeviceModel{V, W},
-    ::NetworkModel{X},
```
jd-lara, line 552: Why this is this changing here again?

jd-lara (changes requested):
We can't change the passing of the network model or the code for the multinetworks will break .

src/devices_models/devices/hydro_generation.jl
```diff
@@ -708,34 +694,13 @@ function calculate_aux_variable_value!(
...
-    resolution = get_resolution(container)
-    fraction_of_hour = Dates.value(Dates.Minute(resolution)) / MINUTES_IN_HOUR
-    p_variable_results = get_variable(container, ActivePowerOutVariable(), T)
+    p_variable_results = get_variable(container, ActivePowerVariable(), T)
     aux_variable_container = get_aux_variable(container, EnergyOutput(), T)
     for d in devices, t in time_steps
         name = PSY.get_name(d)
-        aux_variable_container[name, t] =
-            jump_value(p_variable_results[name, t]) * fraction_of_hour
+        min = PSY.get_active_power_limits(d).min
```
jd-lara, line 702: where is this min being used?

src/devices_models/devices/hydro_generation.jl
```diff
@@ -574,31 +561,30 @@ function add_constraints!(
...
 
         for t in time_steps[2:end]
             constraint[name, t] = JuMP.@constraint(
                 container.JuMPmodel,
                 energy_var[name, t] ==
-                energy_var[name, t - 1] - param[t] * multiplier[name, t] +
+                energy_var[name, t - 1] - param[name, t] * multiplier[name, t] +
                 (
-                    powerout_var[name, t] - powerin_var[name, t] / efficiency +
+                    powerout_var[name, 1] - powerin_var[name, t] / efficiency +
```
jd-lara, line 587: Why is this hardcoded as `1`

PR #962, opened 2023-04-19
Jd/bump pnm

jd-lara:
Requires https://github.com/NREL-SIIP/PowerNetworkMatrices.jl/pull/36 to pass


daniel-thom (approved):

jd-lara:
@daniel-thom I realized that we need to make a PR in PNM to fix the failing test

PR #963, opened 2023-04-19
Jd/bug fix objective function parameters

jd-lara:


rodrigomha (approved):
LGTM but let's wait for the tests to pass

PR #972, opened 2023-05-05
add fix to handle cases with isolated buses

jd-lara:


rodrigomha (approved):
LGTM if the tests are passing

Issue #973, opened 2023-05-08
Isolated buses are not being filtered consistently


josephmckinsey:
**Describe the bug**

Isolated buses are not being filtered within the build process.

- `src/devices_models/devices/common/add_to_expression.jl` errors on SystemBalanceSlackUp

Here are all the places I know that are causing problems:

- `src/network_models/network_slack_variables.jl` calls PSY.get_components(PSY.Bus, sys) in several places. This should probably use subnetworks or something like get_available_components
- `src/network_models/powermodels_interface.jl:321` tries to add maps for isolated buses.
- `src/network_models/pm_translator.jl` creates a map with isolated buses. I haven't seen errors directly here yet, but it is probably related to the `powermodels_interface.jl`
 
**To Reproduce**

I've been using a larger system with 21 isolated buses, which I can't share. There is a day ahead problem template, which fails at the `build!` step.

```
    template = ProblemTemplate()
    set_network_model!(
        template,
        NetworkModel(
            DCPPowerModel,#StandardPTDFModel,
            PTDF_matrix = PNM.PTDF(sys),
            use_slacks = true),
    )
    set_device_model!(template, Line, StaticBranchUnbounded)
    set_device_model!(template, MonitoredLine, StaticBranch)
    set_device_model!(template, Transformer2W, StaticBranchUnbounded)
    set_device_model!(template, TapTransformer, StaticBranchUnbounded)
    set_device_model!(template, PhaseShiftingTransformer, StaticBranchUnbounded)
    set_device_model!(template, RenewableDispatch, RenewableFullDispatch)
    set_device_model!(template, RenewableFix, FixedOutput)
    set_device_model!(template, HydroDispatch, HydroDispatchRunOfRiver)
    set_device_model!(template, ThermalStandard, ThermalStandardUnitCommitment)
    set_device_model!(template, GenericBattery, BookKeeping)
    set_device_model!(template, StandardLoad, StaticPowerLoad)
```

**Expected behavior**

The behavior should be identical to when we were modeling isolated buses.

**Relevant Stack Traces**

```
┌ Error: Simulation build failed                                                                                                               
│   exception =
│    AssertionError: 1393 == 1372
│    Stacktrace:
│      [1] macro expansion
│        @ ~/.julia/packages/InfrastructureSystems/U1f0t/src/utils/assert_op.jl:34 [inlined]
[...]

PR #976, opened 2023-05-11
Jd/fix isolated buses

jd-lara:


src/network_models/pm_translator.jl
```diff
@@ -416,20 +424,15 @@ function get_buses_to_pm(buses::IS.FlattenIteratorWrapper{PSY.Bus})
...
-        PSY.BusTypes.ISOLATED => 4,
-        PSY.BusTypes.PQ => 1,
-        PSY.BusTypes.PV => 2,
-        PSY.BusTypes.REF => 3,
-        PSY.BusTypes.SLACK => 3,
-    )
-
     for bus in buses
+        if PSY.get_bustype(bus) == PSY.BusTypes.ISOLATED
+            continue
```
josephmckinsey, line 429: Any idea why this line isn't being covered by the tests? Seems like test_model_decision should cover it.

josephmckinsey (approved):
I tested it against the data in #973, and the bug is gone even in that complex case. I tested with a couple formulations too.

PR #987, opened 2023-06-27
read the most recent execution of a partition

sourabhdalvi:


daniel-thom:
The changes look good. Why did CI fail?

jd-lara:
@daniel-thom there is something in the new HiGHS release that is causing it to fail

daniel-thom (approved):

PR #988, opened 2023-07-11
Jd/custom model improvements

jd-lara:


docs/src/tutorials/adding_new_problem_model.md
```diff
@@ -0,0 +1,85 @@
...
+```julia
+my_model = DecisionModel{MyCustomDecisionProblem}(
+    sys;
+    name = "MyModel",
+    optimizer = optimizer_with_attributes(HiGHS.Optimizer),
+     optimizer_solve_log_print = true,
+)
+```
+
+#### Mandatory Method Overloads
```
daniel-thom, line 56: I would use `overload` in cases where you add a method that replaces an existing method. In the case of these custom types, the user must implement (or extend) these methods because they don’t exist.

src/operation/decision_model.jl
```diff
@@ -135,7 +135,7 @@ function DecisionModel{M}(
 end
 
 """
-This builds the optimization problem of type M with the specific system and template
+Builds the optimization problem of type M with the specific system and template
```
daniel-thom, line 138: Per Julia rules it should be `Build the optimization …`
https://docs.julialang.org/en/v1/manual/documentation/#Writing-Documentation

daniel-thom (changes requested):

docs/src/modeler_guide/definitions.md
```diff
@@ -10,12 +10,12 @@
...
 
 ## I
 
-* *Interval*:
+* *Interval*: The amount of time between updates to the decision problem. For instance, Day-Ahead problems usually have a 24-hour intervals and Real-Time problems have 5-minute intervals. Check the time [Time Series Data Section in PowerSystems.jl](https://nrel-sienna.github.io/PowerSystems.jl/stable/modeler_guide/time_series/)
 
 ## R
 
-* *Resolution*: The amount of time between timesteps in a simulation. For instance 1-hour or 5-minutes. In Julia these are defined using the syntax `Hour(1)` and `Minute(5)`
+* *Resolution*: The amount of time between timesteps in a simulation. For instance 1-hour or 5-minutes. In Julia these are defined using the syntax `Hour(1)` and `Minute(5)`. Check the time [Time Series Data Section in PowerSystems.jl](https://nrel-sienna.github.io/PowerSystems.jl/stable/modeler_guide/time_series/)
```
rodrigomha, line 21: I think it will be good to add an example after these definitions. When I was struggling with these definitions, an example would have been the most helpful. We could add an example in other page and refer directly in the same definition. Here we are adding the time series data section in the Docs, and is fine, but we should also provide a direct link to an example.

Here is an example we could use:

In a typical simulation pipeline, we want to connect daily (24-hours) day-ahead unit commitment problems, with multiple economic dispatch problems. Usually, our day-ahead unit commitment problem will have an hourly (1-hour) resolution, while the economic dispatch will have a 5-minute resolution. 

Depending on your problem, it is common to use a 2-day look-ahead for unit commitment problems, so in this case, the Day-Ahead problem will have: `resolution = Hour(1)` with `interval = Hour(24)` and `horizon = 48`. In the case of the economic dispatch problem, it is common to use a look-ahead of two hours. Thus, the Real-Time problem will have: `resolution = Minute(5)`, with `interval = Minute(5)` (we only store the first operating point) and `horizon = 24` (24 time steps of 5 minutes are 120 minutes, that is 2 hours).

jd-lara: This example seems more appropriate for the definition of a sequence

jd-lara:
This PR has some documentation enhancement but it isn't exhaustive. I will leave the additional documentation for a later  PR

PR #995, opened 2023-08-09
Fix bug in custom indexing for realized results

daniel-thom:
This bug was introduced in commit
21f6dff464b322f4785a426ef6dd215082ed0510. Passing custom values of start_time and len that cause the code to span arrays from two initial times caused an array indexing error.


jd-lara (approved):

josephmckinsey (approved):
This new code makes a lot more sense than the old row indexing. I couldn't test on my own data since it's still on PSI 0.20.4

PR #999, opened 2023-08-24
Removing remainder hydro code from PowerSimulations

pesap:
This  PR removes most of the code from Power Simulations that was migrated to HydroPowerSimulations. @jd-lara , I just comment most of the code in here, but can delete it  if needed.


src/PowerSimulations.jl
```diff
@@ -70,7 +70,7 @@ export DeviceLimitedRegulation
 export ReserveLimitedRegulation
 
 ###### Hydro #######
-export HydroDispatchRunOfRiver
+# export HydroDispatchRunOfRiver
```
jd-lara, line 73: We can delete both models in PSI.

jd-lara:
@pesap the PR looks good to me. We can just go ahead and remove the code. We will need to merge and tag HPS before the tests can pass here.

jd-lara (approved):

PR #1001, opened 2023-08-29
Jdrh/fix reserves

jd-lara:


scripts/create_rts_systems.jl
jd-lara, line 1: The scripts folder needs to be removed before merging

rodrigomha (approved):
This is working fine and tests are passing locally in my machine.

PR #1010, opened 2023-09-19
Jd/hvdc losses bugfix

jd-lara:


alefcastelli (approved):
codecov states that there are some new lines that are not covered by the tests but this can be fixed in the future

PR #1011, opened 2023-09-21
relax test

jd-lara:


jd-lara:
The introduction of linear solve vs matrix inversion method in PowerNetworkMatrices created some numerical differences in the PTDF values that seem to be enough to cause a larger difference in the cost of PTDF vs DCP. 

The optimization json file diffs are shown below. 

image


image

PR #1014, opened 2023-10-09
Jd/move energy limit ff out

jd-lara:


rodrigomha:
@pesap Are we using this EnergyLimitFF in HydroSystems too?

pesap:
> @pesap Are we using this EnergyLimitFF in HydroSystems too?

I dont think so. We use the EnergyTargetParameter.  But probably is something that we might use?

jd-lara:
The tests are expected to fail due to the test dependencies on HydroPowerSimulations.

PR #1028, opened 2023-12-13
fix uses of default attributes

### LINKED ISSUE #1020: Implement new way of retrieving default attributes
Currently, the approach for including special constraints is by enabling an attribute to be true. In a specific device with more than one attribute, the user usually want to specify a single attribute and use default for others. If the default attribute is:

``` 
function PSI.get_default_attributes(
    ::Type{Component},
    ::Type{T},
) where {T <: AbstractFormulation}
    return Dict{String, Any}(
        "attr1" => false,
        "attr2" => false,
        "attr3" => false,
        "attr4" => false,
    )
end
```
and the user wants to specify the device model to be:
```
device_model = DeviceModel(
    GenericComponent,
    GenericFormulation;
    attributes=Dict{String, Any}(
        "attr1" => true,
    ),
)
set_device_model!(template, device_model)
```
The `build` will fail when it tries to construct given that the dictionary does not have `attr2` defined, and is expecting to do:
`if PSI.get_attribute(model, "attr2")` is not there. This forces the user to define the `DeviceModel` with all the attributes when only wants to change one as:
```
device_model = DeviceModel(
    GenericComponent,
    GenericFormulation;
    attributes=Dict{String, Any}(
        "attr1" => true,
        "attr2" => false,
        "attr3" => false,
        "attr4" => false,
    ),
)
set_device_model!(template, device_model)
```

jd-lara:
closes #1020

### DIFF
--- a/src/core/device_model.jl
+++ b/src/core/device_model.jl
@@ -59,8 +59,13 @@ mutable struct DeviceModel{D <: PSY.Device, B <: AbstractDeviceFormulation}
         use_slacks = false,
         duals = Vector{DataType}(),
         time_series_names = get_default_time_series_names(D, B),
-        attributes = get_default_attributes(D, B),
+        attributes = Dict{String, Any}(),
     ) where {D <: PSY.Device, B <: AbstractDeviceFormulation}
+        attributes_ = get_default_attributes(D, B)
+        for (k, v) in attributes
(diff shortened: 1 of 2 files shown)


rodrigomha (approved):
This looks good. Let's wait for the test to pass and merge

PR #1029, opened 2023-12-13
Remove EnergyOutput Aux Vars

rodrigomha:


pesap:
@rodrigomha , Do we need to make a realease on HPS to pass this test?

jd-lara (approved):
Tests are expected to fail until a new release of HPS

PR #1034, opened 2024-01-03
bump power models

jd-lara:


daniel-thom (approved):

jd-lara:
The tests are failing due to the change in the constraint creation in PowerModels.jl https://github.com/lanl-ansi/PowerModels.jl/issues/819

PR #1040, opened 2024-01-09
Support in the loop PowerFlow evaluation 

jd-lara:
This is a WIP branch to integrate the evaluation of PowerFlows into a simulation loop. 

~GKS edit: depends on https://github.com/NREL-Sienna/PowerFlows.jl/pull/53~
GKS edit: temporarily depends on PowerFlows' `gks/finalize_export_naming` instead


src/core/network_model.jl
```diff
@@ -34,6 +34,7 @@ mutable struct NetworkModel{T <: PM.AbstractPowerModel}
     duals::Vector{DataType}
     radial_branches::PNM.RadialBranches
     reduce_radial_branches::Bool
+    powerflow_evaluation::Union{Nothing, PFS.PowerFlowEvaluationModel}
```
daniel-thom, line 37: `power_flow_evaluation`?

src/network_models/powerflow_evaluation.jl
```diff
@@ -0,0 +1,72 @@
...
+    branch_names::Vector{String},
+)
+    time_steps = get_time_steps(container)
+    for var_type in vars
+        for D in Set(branch_types)
+            add_aux_variable_container!(
+                container,
+                var_type,
+                D,
+                branch_names[branch_types == D],
```
daniel-thom, line 18: What’s happening here?

src/core/powerflow_data_wrapper.jl
```diff
@@ -0,0 +1,13 @@
+mutable struct PowerFlowEvaluationData{T <: PFS.PowerFlowData}
```
jd-lara, line 1: @daniel-thom I added this new struct to store other information required to inter operate the PFData object the optimization container.

src/network_models/powerflow_evaluation.jl
```diff
@@ -0,0 +1,207 @@
...
+    return
+end
+
+const ACTIVE_POWER_INJECTION_KEYS = [
+    ActivePowerVariable
+    PowerOutput
+    ActivePowerTimeSeriesParameter
+]
+
+function _make_injection_map!(container::OptimizationContainer, sys::PSY.System)
```
jd-lara, line 54: @daniel-thom @GabrielKS This loop deserves son good review. This was the best approach I could come up with to find the index in the PoweFlowData matrices that correspond to the values that need to be read from the optimization container. 

What I am doing here is using a const for the types of keys that I want to capture, an additional alternative is to use traits and implement `is_injection_key(var)` but I went with this instead for simplicity. 

The function has comments about what is doing but if there are questions please let me know. I am not 100% confident on this implementation being the best.

GabrielKS:
Requires https://github.com/NREL-Sienna/PowerFlows.jl/pull/48.

src/network_models/power_flow_evaluation.jl
```diff
@@ -0,0 +1,262 @@
...
+        push!(added_injection_types, comp_type)
+        for n in axes(array)[1]
+            name_bus_ix_map[n] = temp_component_bus_map[comp_type][n]
+        end
+
+        pf_data_opt_container_map[key] = name_bus_ix_map
+    end
+
+    for (key, array) in get_aux_variables(container)
+        if get_entry_type(key) ∉ ACTIVE_POWER_INJECTION_KEYS
```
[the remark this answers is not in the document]
GabrielKS: Done

GabrielKS:
@jd-lara I can't tag you for review on this since it's technically your PR, so this is me letting you know that this is ready for review. There's a few `TODO`s that I'd like to think about a bit more, but I think it'd be good to get your eyes on it now.

jd-lara:
One issue to consider is the following, in Models with unit commitment (hydro and thermal) if the unit is connected to a PV Bus bus it is essentially off the bus type needs to change to PQ to make the power flow problem feasible. This implies that we need an aux variable for the bus type and then enact that change both in the call to solve as well as in the exporter for any models of the subtype Commitment

src/core/power_flow_data_wrapper.jl
```diff
@@ -2,18 +2,20 @@ mutable struct PowerFlowEvaluationData{T <: PFS.PowerFlowContainer}
...
-    For `PowerFlowData`, values are `Dict{String, Int64}` mapping component name to matrix
-    index of bus; for `SystemPowerFlowContainer`, values are Dict{Union{String, Int64},
+    The Symbol is a category of data: `:active_power`, `:reactive_power`, etc. The
+    `OptimizationContainerKey` is a source of that data in the `OptimizationContainer`. For
+    `PowerFlowData`, leaf values are `Dict{String, Int64}` mapping component name to matrix
+    index of bus; for `SystemPowerFlowContainer`, leaf values are Dict{Union{String, Int64},
     Union{String, Int64}} mapping component name/bus number to component name/bus number.
     """
-    input_key_map::Dict{<:OptimizationContainerKey, <:Any}
+    input_key_map::Dict{Symbol, <:Dict{<:OptimizationContainerKey, <:Any}}
```
jd-lara, line 11: Why this change?

src/network_models/power_flow_evaluation.jl
```diff
@@ -73,35 +87,48 @@ function _make_pf_input_map!(
...
+        pf_data_opt_container_map = Dict{OptimizationContainerKey, map_type}()
+        @info "Adding input map to send $category to $(nameof(typeof(pf_data)))"
+        precedence = PF_INPUT_KEY_PRECEDENCES[category]
+        added_injection_types = DataType[]
+        # For each data source that is relevant to this category in order of precedence,
+        # loop over the component types where data exists at that source and record the
+        # association
+        for entry_type in precedence
+            for (key, val) in available_keys
+                (get_entry_type(key) === entry_type) || continue
```
jd-lara, line 111: make this if statement for readability

src/core/power_flow_data_wrapper.jl
GabrielKS: Separate dicts for active and reactive power, basically. There were other ways to do it but I think this is the most natural.

src/network_models/power_flow_evaluation.jl
GabrielKS: Done

GabrielKS:
This is finally ready for review! It has a ton of unreleased dependencies so it can't actually be released yet, but I think there is no outstanding work to be done on this PR. The three test files modified here pass when run with the branches of PF, PNM, and PSY I'm currently working on; more thorough testing should be run when those dependencies are finalized.

PR #1046, opened 2024-01-22
remove changes related to PTDFNetwork

### LINKED ISSUE #1044: Drop support for PTDFPowerModel
There are a bunch of weirdly implemented logic and methods in relation to supporting `PTDFPowerModel`. In the current state we have already addressed some of the original scaling issues that lead to adding this capability 3 years ago. 

RFC: @claytonpbarrows

jd-lara:
Closes #1044

### DIFF
--- a/src/PowerSimulations.jl
+++ b/src/PowerSimulations.jl
@@ -20,7 +20,6 @@ export SimulationPartitionResults
 # Network Relevant Exports
 export NetworkModel
 export StandardPTDFModel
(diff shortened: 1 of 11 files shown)


jd-lara (commented):
@claytonpbarrows should we consider renaming StandardPTDFModel to PTDFPowerModel after this PR? It would be more consistent with the rest of the semantics

claytonpbarrows (approved):

PR #1051, opened 2024-02-08
add xpress patch

jd-lara:
See https://github.com/jump-dev/Xpress.jl/issues/212 for reference to this. It seems to be the only solution that is under our control.


jd-lara:
@odow this is the best I came up with to fix the problem with Xpress

odow:
In SDDP we do solve, if infeasible, re-build and then re-solve. If still infeasible. Write to file.

See

 * https://github.com/odow/SDDP.jl/blob/66096558ae9ba652086d954b156b809361c519ee/src/algorithm.jl#L421-L423
 * https://github.com/odow/SDDP.jl/blob/66096558ae9ba652086d954b156b809361c519ee/src/algorithm.jl#L279-L301

daniel-thom (approved):

PR #1061, opened 2024-02-26
Variable Cost Refactor Part 2: PowerSimulations

GabrielKS:
The second part of a significant refactor of Sienna cost data structures (part 1 is [here](https://github.com/NREL-Sienna/PowerSystems.jl/pull/1056#issuecomment-1959057500)). The main focus here is to bring PowerSimulations.jl up to date with the elimination of `VariableCost` in favor of `FunctionData`. That entails these changes in PowerSimulations:

- Rewrite some documentation to refer to the new cost structs
- Replace all instances of `VariableCost`, and some instances of contents of `VariableCost`, with `FunctionData`

On the latter point, I am not completely sure if I made the switch from using `FunctionData` to using raw data (vectors of tuples) at the proper place; I'd appreciate review of that.

Tests pass:
Screenshot 2024-02-26 at 4 32 27 AM

### DIFF
--- a/src/parameters/update_parameters.jl
+++ b/src/parameters/update_parameters.jl
@@ -576,7 +576,7 @@ function _update_pwl_cost_expression(
     dt = Dates.value(Dates.Second(resolution)) / SECONDS_IN_HOUR
     gen_cost = JuMP.AffExpr(0.0)
     slopes = PSY.get_slopes(cost_data)
-    upb = PSY.get_x_lengths(cost_data)
+    upb = get_breakpoint_upper_bounds(cost_data)
     for i in 1:length(cost_data)
         JuMP.add_to_expression!(
             gen_cost,
--- a/src/parameters/update_parameters.jl
+++ b/src/parameters/update_parameters.jl
@@ -571,14 +571,14 @@ function _update_pwl_cost_expression(
     ::Type{T},
     component_name::String,
     time_period::Int,
-    cost_data::Vector{NTuple{2, Float64}},
+    cost_data::PSY.PiecewiseLinearPointData,
 ) where {T <: PSY.Component}
     pwl_var_container = get_variable(container, PieceWiseLinearCostVariable(), T)
     resolution = get_resolution(container)
     dt = Dates.value(Dates.Second(resolution)) / SECONDS_IN_HOUR
     gen_cost = JuMP.AffExpr(0.0)
     slopes = PSY.get_slopes(cost_data)
-    upb = PSY.get_breakpoint_upperbounds(cost_data)
+    upb = PSY.get_x_lengths(cost_data)
     for i in 1:length(cost_data)
         JuMP.add_to_expression!(
             gen_cost,
@@ -614,9 +614,9 @@ end
 
 function update_variable_cost!(
     container::OptimizationContainer,
-    parameter_array::DenseAxisArray{Vector{NTuple{2, Float64}}},
+    parameter_array::DenseAxisArray{PSY.PiecewiseLinearPointData},
     parameter_multiplier::JuMPFloatArray,
-    ::CostFunctionAttributes{Vector{NTuple{2, Float64}}},
+    ::CostFunctionAttributes{PSY.PiecewiseLinearPointData},
     component::T,
     time_period::Int,
 ) where {T <: PSY.Component}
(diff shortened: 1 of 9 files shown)


GabrielKS:
The group of four PRs this is a part of:

1. https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/332
2. https://github.com/NREL-Sienna/PowerSystems.jl/pull/1059
3. https://github.com/NREL-Sienna/PowerSystemCaseBuilder.jl/pull/73
4. https://github.com/NREL-Sienna/PowerSimulations.jl/pull/1061

src/parameters/update_parameters.jl
```diff
@@ -543,10 +543,8 @@ function _update_parameter_values!(
             )
             variable_cost_forecast_values = TimeSeries.values(ts_vector)
             for (t, value) in enumerate(variable_cost_forecast_values)
-                if attributes.uses_compact_power
```
jd-lara, line 546: This over writes the value variable in cases where compact_power is used. It is not properly tested but do not remove it

src/parameters/update_parameters.jl
```diff
@@ -571,14 +569,14 @@ function _update_pwl_cost_expression(
...
-    cost_data::Vector{NTuple{2, Float64}},
+    cost_data::PSY.PiecewiseLinearPointData,
 ) where {T <: PSY.Component}
     pwl_var_container = get_variable(container, PieceWiseLinearCostVariable(), T)
     resolution = get_resolution(container)
     dt = Dates.value(Dates.Second(resolution)) / SECONDS_IN_HOUR
     gen_cost = JuMP.AffExpr(0.0)
     slopes = PSY.get_slopes(cost_data)
-    upb = PSY.get_breakpoint_upperbounds(cost_data)
+    upb = PSY.get_x_lengths(cost_data)
```
jd-lara, line 579: Not sure about the name for this function. @sourabhdalvi what do you think?

jd-lara (changes requested):
Overall looks ok, please restore the piece of code that was eliminated.

src/parameters/update_parameters.jl
GabrielKS: That helper function `_convert_variable_cost` does not seem to be defined anywhere, so this code would just error if it ever came up. Is there a definition somewhere I should reinstate?

src/parameters/update_parameters.jl
GabrielKS: My issue with just saying breakpoints is that it's specifically the difference between `x[n]` and `x[n-1]`, whereas from my understanding "breakpoints" can sometimes refer to just the x values themselves. Is that correct?

src/parameters/update_parameters.jl
```diff
@@ -600,7 +598,8 @@ function update_variable_cost!(
     dt = Dates.value(Dates.Second(resolution)) / SECONDS_IN_HOUR
     base_power = get_base_power(container)
     component_name = PSY.get_name(component)
-    cost_data = parameter_array[component_name, time_period]
+    cost_data = parameter_array[component_name, time_period]  # TODO is this a new-style cost?
+    println(typeof(cost_data))  # TODO REMOVE
```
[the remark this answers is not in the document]
GabrielKS: Done

src/parameters/update_parameters.jl
jd-lara: Looks like this code hasn't been updated in 2 years. Julia will throw a not implemented method anyway. Leave the code and we will implement the method when we make the tests and they inevitable will fail. For now, we can put a TODO

src/parameters/update_parameters.jl
jd-lara: I agree with you from the data point of view but the variables for PWL representation are called breakpoints and these will return the upper bounds for the break point variables in the optimization problem

src/parameters/update_parameters.jl
GabrielKS: Done

src/parameters/update_parameters.jl
GabrielKS: Done with a `get_breakpoint_upperbounds` alias in `powersystem_utils.jl`

jd-lara (approved):

PR #1066, opened 2024-02-29
Jd/psi decompositionv3

jd-lara:


src/utils/powersystems_utils.jl
```diff
@@ -1,28 +1,76 @@
...
-    return PSY.get_components(PSY.get_available, T, sys)
+    subsystem = get_subsystem(model)
+    return PSY.get_components(PSY.get_available, T, sys; subsystem_name = subsystem)
 end
 
 function get_available_components(
-    ::Type{T},
+    model::DeviceModel{T, <:AbstractDeviceFormulation},
+    sys::PSY.System,
+    f::Function,
```
daniel-thom, line 13: Does these functions need to reorder the parameters so that filter_function is first?

jd-lara: We can incorporate that change for sure. It is a legacy from old code.

daniel-thom (approved):
LGTM

PR #1068, opened 2024-02-29
Jd/move structs to is

jd-lara:
Companion PR to https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/335


src/simulation/simulation_results_export.jl
```diff
@@ -161,5 +161,5 @@ function _should_export(exports::SimulationResultsExport, tstamp, model, field_n
     end
 
     problem_exports = get_problem_exports(exports, model)
-    return _should_export(problem_exports, field_name, name)
+    return IS.Optimization._should_export(problem_exports, field_name, name)
```
jd-lara, line 164: Note for later: 

We should define `_should_export` in each package not in IS.

GabrielKS:
All tests are now passing here (except for `Test AGC`, which is broken in `psy4` for unrelated reasons)

rodrigomha:
Here is the test environment on which tests are passing locally:
image

PR #1069, opened 2024-03-02
Initial Conditions Reconciliation #1: Detect Mismatches

GabrielKS:
Here we implement the logic to automatically detect mismatches between initial conditions that `requires_reconciliation`. Some notes about the implementation:

- How `requires_reconciliation` figures in: first we make the set of all `ICKey`s to check, then as we are looping over them to check them we skip those whose `T::InitialConditionType` has `requires_reconciliation(T) == false`, so keys that don't require reconciliation never get compared across models even if other keys do require reconciliation. At one point we discussed comparing all the keys if at least one `requires_reconciliation`; this is not that. This behavior (coupled with erring on the side of `requires_reconciliation(T) == true`) makes the most sense to me — if a key is at all relevant during this process we should just mark it as requiring reconciliation — but that should be reviewed.
- We still need to finalize our list of `requires_reconciliation` methods before this draft PR becomes ready for review, I'd love to get @lpstreitmatter and José David's input on that.
- Mismatches are reported as `@info` for each mismatch, then if there are nonzero mismatches we report a `@warn` noting that there is not yet logic to resolve the mismatch.


src/core/initial_conditions.jl
```diff
@@ -159,3 +159,10 @@ struct InitialTimeDurationOn <: InitialConditionType end
 struct InitialTimeDurationOff <: InitialConditionType end
 struct InitialEnergyLevel <: InitialConditionType end
 struct AreaControlError <: InitialConditionType end
+
+# Decide whether to run the initial conditions reconciliation algorithm based on the presence of any of these
+requires_reconciliation(::Type{<:InitialConditionType}) = false
+
+requires_reconciliation(::Type{<:InitialTimeDurationOn}) = true
+requires_reconciliation(::Type{<:InitialTimeDurationOff}) = true
+requires_reconciliation(::Type{<:DeviceStatus}) = true
```
jd-lara, line 168:
```suggestion
requires_reconciliation(::Type{InitialTimeDurationOn}) = true
requires_reconciliation(::Type{InitialTimeDurationOff}) = true
requires_reconciliation(::Type{DeviceStatus}) = true
```

jd-lara (changes requested):

jd-lara:
@GabrielKS I changed the base to go to PSY4

src/core/initial_conditions.jl
GabrielKS: Done in @lpstreitmatter's commit

GabrielKS:
@jd-lara marking this as ready for review now — how does the current set of `requires_reconciliation` traits look to you?

jd-lara:
@GabrielKS can you fix the conflicts please

src/core/initial_conditions.jl
```diff
@@ -159,3 +159,14 @@ struct InitialTimeDurationOn <: InitialConditionType end
...
+# Decide whether to run the initial conditions reconciliation algorithm based on the presence of any of these
+requires_reconciliation(::Type{<:InitialConditionType}) = false
+
+requires_reconciliation(::Type{InitialTimeDurationOn}) = true
+requires_reconciliation(::Type{InitialTimeDurationOff}) = true
+requires_reconciliation(::Type{DeviceStatus}) = true
+requires_reconciliation(::Type{DevicePower}) = true # to capture a case when device is off in HA but producing power in ED
+requires_reconciliation(::Type{DeviceAboveMinPower}) = true # ramping limits may make power differences in thermal compact devices between models infeasible 
+requires_reconciliation(::Type{InitialEnergyLevel}) = true # large differences in initial storage levels could lead to infeasibilities
+# Not requiring reconciliation for AreaControlError
```
jd-lara, line 172: Probably needs a formatter run

src/simulation/simulation.jl
```diff
@@ -266,6 +266,62 @@ function _check_folder(sim::Simulation)
...
 
+# Compare initial conditions for all `InitialConditionType`s with the
+# `requires_reconciliation` trait across `models`, log @info messages for mismatches
+function _initial_conditions_reconciliation!(
+    models::Vector{DecisionModel{GenericOpProblem}})
+    model_names = get_name.(models)
+    has_mismatches = false
+    @info "Reconciling initial conditions across models $(join(model_names, ", "))"
+    # all_ic_keys: all the `ICKey`s that appear in any of the models
+    # TODO: incorporate requires_reconciliation
```
jd-lara, line 277: Is this TODO still required?

src/simulation/simulation.jl
```diff
@@ -266,6 +266,62 @@ function _check_folder(sim::Simulation)
...
+            ic_vals_per_component =
+                Dict(get_name(get_component(ic)) => get_condition(ic) for ic in ics[ic_key])
+            ic_vals_per_model[i] = ic_vals_per_component
+        end
+
+        # Assert that all models have the same components for current ic_key
+        @assert allequal(Set.(keys.(values(ic_vals_per_model)))) "For IC key $ic_key, not all models have the same components"
+
+        # For each component in current ic_key, compare values across models
+        component_names = collect(keys(first(values(ic_vals_per_model))))
```
jd-lara, line 301: there is no need to call collect here to loop over the keys.

src/core/initial_conditions.jl
GabrielKS: Done

src/simulation/simulation.jl
GabrielKS: No, removed

src/simulation/simulation.jl
GabrielKS: Removed

GabrielKS:
Fixed the merge conflicts, made the tests pass in `psy4` (depends on https://github.com/NREL-Sienna/StorageSystemsSimulations.jl/pull/35)

jd-lara (approved):

Issue #1072, opened 2024-03-04
`load_results!` incorrectly "loads" data from the cache to populate the cache


GabrielKS:
Two related issues here:

1. The documentation for `load_results!` says "Running this function twice overwrites the previously loaded results" but that is not correct: running the function a second time with the same keys ends up fetching the keys from the cache and updating them with themselves.
2. In doing so, the check to make sure the cache has all the timestamps we want gets sidestepped, so the cache can end up in a state where it think it has more timestamps than it actually does:

```julia
julia> timestamps = PSI._process_timestamps(res, DateTime("2035-01-01"), 7);

julia> variable_tuples = [(ActivePowerVariable, ThermalStandard)];

julia> variable_keys = [PSI.VariableKey(ActivePowerVariable, ThermalStandard)];

julia> load_results!(res, 6; variables=variable_tuples)  # loads 6 windows

julia> load_results!(res, 7; variables=variable_tuples)  # should load 7 windows, in fact updates res.results_timestamps to have 7 timestamps but does not load any more data

julia> PSI._read_results(res, variable_keys, timestamps, nothing) |> values |> first |> x->x.data |> keys |> collect  # uses the cache because the cache falsely claims to have the requisite 7 windows
6-element Vector{DateTime}:
2035-01-01T00:00:00
2035-01-02T00:00:00
2035-01-03T00:00:00
2035-01-04T00:00:00
2035-01-05T00:00:00
2035-01-06T00:00:00

julia> # only 6 timestamps in the result even though we asked for 7
```

In consultation with @daniel-thom, I plan to resolve as follows:

1. Edit the documentation for `load_results!` to say that running the function a second time will add any new keys/time periods to the cache that aren't already there but will not replace existing data.
2. Fix the above bug so that increasing the number of time periods actually loads more data.

GabrielKS:
Fixed in `psy4`

Issue #1018, opened 2023-11-02
Implement consistent interface for stores 


jd-lara:
The InMemoryStore requires an interface with 

```
with open(stores, "w") do store
   foo(store) 
end 
```

to avoid passing nothing valued store into several functions.

PR #1073, opened 2024-03-05
Fix caching behaviors: #1070, #1072, #1074

### LINKED ISSUE #1070: `_read_results` for `Matrix{Float64}` and `DecisionModelSimulationResults` does not use the cache
The implementations of the two `_read_results` methods in [`decision_model_simulation_results.jl`](https://github.com/NREL-Sienna/PowerSimulations.jl/blob/v0.27.4/src/simulation/decision_model_simulation_results.jl) show that one of them checks the cache and the other doesn't. This causes a *vast* difference in the performance of the two methods:

```Julia
# where `res` is a `PowerSimulations.SimulationProblemResults{PowerSimulations.DecisionModelSimulationResults}` with weekly data and a lot of components
julia> timestamps = PSI._process_timestamps(res, DateTime("2035-01-01"), 7);
julia> variables = [PSI.VariableKey(ActivePowerVariable, ThermalStandard)];
julia> load_results!(res, 7; variables=[(ActivePowerVariable, ThermalStandard)]);
julia> @time results_mat = PSI._read_results(Matrix{Float64}, res, variables, timestamps, nothing)
  0.091891 seconds (191.44 k allocations: 15.228 MiB)
Dict{PowerSimulations.OptimizationContainerKey, PowerSimulations.ResultsByTime{Matrix{Float64}}} with 1 entry:
  VariableKey{ActivePowerVariable, ThermalStandard}("") => ResultsByTime{Matrix{Float64}, 1}(VariableKey{ActivePowe…
julia> @time results_jump = PSI._read_results(res, variables, timestamps, nothing)
  0.000150 seconds (133 allocations: 4.953 KiB)
Dict{PowerSimulations.VariableKey{ActivePowerVariable, ThermalStandard}, PowerSimulations.ResultsByTime{JuMP.Containers.DenseAxisArray{Float64, 2, Ax, L} where {Ax, L<:Tuple{JuMP.Containers._AxisLookup, JuMP.Containers._AxisLookup}}, 1}} with 1 entry:
  VariableKey{ActivePowerVariable, ThermalStandard}("") => ResultsByTime{DenseAxisArray{Float64, 2, Ax, L} where {A…
```

The second one becomes as slow as the first if the cache check is commented out. This matters because the first method is what `read_results_with_keys` uses, and therefore this behavior contributes to a lot of slowdown in my PowerAnalytics rewrite, where I call `read_results_with_keys` for each component and need a cache to make this tolerable on large systems.

GabrielKS:
This fixes #1070 and #1072.

## On #1070:
There is some inherent overhead to creating large arrays that cannot be avoided, but now that we use the cache the time for repeated calls to `_read_results(Matrix{Float64}...` on my big dataset has decreased 30x from 0.09s to 0.003s:
```julia
julia> @time results_mat = PSI._read_results(Matrix{Float64}, res, variables, timestamps, nothing)
  0.003551 seconds (340 allocations: 7.419 MiB)
```

This is much better and closes the issue but is still unnecessarily heavy-handed for my PowerAnalytics component-wise reads, so later I may seek to add an option to PowerSimulations to perform `read_results_with_keys` for only one component at a time.

## On #1072:
Ended up being more refactoring than initially planned, but the upshot is that for the `DecisionModelSimulationResults` method of `load_results!` the documentation now precisely describes what happens, what happens is good, the bug preventing it from loading more timestamps is gone, some code duplication has been removed, and there are lots more tests. I did not significantly edit the `EmulationModelSimulationResults` method of `load_results!`; it appears to have a significantly different interface that does not even take in time steps.

```julia
# after the setup detailed in bug #1072
julia> PSI._read_results(res, variable_keys, timestamps, nothing) |> values |> first |> x->x.data |> keys |> collect
7-element Vector{DateTime}:
 2035-01-01T00:00:00
 2035-01-02T00:00:00
 2035-01-03T00:00:00
 2035-01-04T00:00:00
 2035-01-05T00:00:00
 2035-01-06T00:00:00
 2035-01-07T00:00:00
```

### DIFF
--- a/src/simulation/decision_model_simulation_results.jl
+++ b/src/simulation/decision_model_simulation_results.jl
@@ -267,10 +267,10 @@ function _read_results(
             (date_key, Matrix{Float64}(permutedims(inner_data.data)))
             for (date_key, inner_data) in result_data.data)
         converted_vals[result_key] = ResultsByTime{Matrix{Float64}, 1}(
-                        result_data.key,
-                        inner_converted,
-                        result_data.resolution,
-                        result_data.column_names)
+            result_data.key,
+            inner_converted,
+            result_data.resolution,
+            result_data.column_names)
     end
     return converted_vals
 end
@@ -524,12 +524,12 @@ function load_results!(
 
     function merge_results(store)
         for (key_type, new_items) in [
-                (ConstraintKey, duals),
-                (ParameterKey, parameters),
-                (VariableKey, variables),
-                (AuxVarKey, aux_variables),
-                (ExpressionKey, expressions),
-            ]
+            (ConstraintKey, duals),
+            (ParameterKey, parameters),
+            (VariableKey, variables),
+            (AuxVarKey, aux_variables),
+            (ExpressionKey, expressions),
+        ]
             new_keys = key_type[_deserialize_key(key_type, res, x...) for x in new_items]
             existing_results = get_cached_results(res, key_type)
             total_keys = union(collect(keys(existing_results)), new_keys)
--- a/src/simulation/decision_model_simulation_results.jl
+++ b/src/simulation/decision_model_simulation_results.jl
@@ -273,6 +273,18 @@ function _read_results(
     if _are_results_cached(res, result_keys, timestamps, keys(cached_results))
         @debug "reading results from SimulationsResults cache"  # NOTE tests match on this
         vals = Dict(k => cached_results[k] for k in result_keys)
+        # Cached data may contain more timestamps than we need, remove these if so
+        (timestamps == get_results_timestamps(res)) && return vals
+        filtered_vals = Dict{keytype(vals), valtype(vals)}()
+        for (result_key, result_data) in vals
+            inner_converted = filter((((k, v),) -> k in timestamps), result_data.data)
+            filtered_vals[result_key] = ResultsByTime{valtype(inner_converted), 1}(
+                result_data.key,
+                inner_converted,
+                result_data.resolution,
+                result_data.column_names)
+        end
+        return filtered_vals
     else
         @debug "reading results from data store"  # NOTE tests match on this
         vals = _get_store_value(res, result_keys, timestamps, store)
--- a/src/simulation/decision_model_simulation_results.jl
+++ b/src/simulation/decision_model_simulation_results.jl
@@ -126,19 +126,6 @@ function _get_store_value(
 end
 
 function _get_store_value(
-    T::Type{Matrix{Float64}},
-    res::SimulationProblemResults{DecisionModelSimulationResults},
-    container_keys::Vector{<:OptimizationContainerKey},
-    timestamps::Vector{Dates.DateTime},
-    ::Nothing,
-)
-    simulation_store_path = joinpath(get_execution_path(res), "data_store")
-    return open_store(HdfSimulationStore, simulation_store_path, "r") do store
-        _get_store_value(T, res, container_keys, timestamps, store)
-    end
-end
-
-function _get_store_value(
     sim_results::SimulationProblemResults{DecisionModelSimulationResults},
     container_keys::Vector{<:OptimizationContainerKey},
     timestamps::Vector{Dates.DateTime},
@@ -239,44 +226,6 @@ function _get_store_value(
     return results_by_time
 end
 
-function _get_store_value(
-    ::Type{Matrix{Float64}},
-    sim_results::SimulationProblemResults{DecisionModelSimulationResults},
-    container_keys::Vector{<:OptimizationContainerKey},
-    timestamps::Vector{Dates.DateTime},
-    store::SimulationStore,
-)
-    base_power = get_model_base_power(sim_results)
-    results_by_key = Dict{OptimizationContainerKey, ResultsByTime{Matrix{Float64}}}()
-    model_name = Symbol(get_model_name(sim_results))
-    resolution = get_resolution(sim_results)
-
-    for key in container_keys
-        n_dims = get_number_of_dimensions(store, DecisionModelIndexType, model_name, key)
-        if n_dims != 1
-            error(
-                "The number of dimensions $(n_dims) is not supported for $(encode_key_as_string(key))",
-            )
-        end
-        results_by_time = ResultsByTime{Matrix{Float64}, 1}(
-            key,
-            SortedDict{Dates.DateTime, Matrix{Float64}}(),
-            resolution,
-            get_column_names(store, DecisionModelIndexType, model_name, key),
-        )
-        for ts in timestamps
-            array = read_result(Array, store, model_name, key, ts)
-            if convert_result_to_natural_units(key)
-                array .*= base_power
-            end
-            results_by_time[ts] = array
-        end
-        results_by_key[key] = results_by_time
-    end
-
-    return results_by_key
-end
-
 function _process_timestamps(
     res::SimulationProblemResults,
     initial_time::Union{Nothing, Dates.DateTime},
@@ -305,20 +254,25 @@ function _process_timestamps(
 end
 
 function _read_results(
-    T::Type{Matrix{Float64}},
+    ::Type{Matrix{Float64}},
     res::SimulationProblemResults{DecisionModelSimulationResults},
     result_keys,
     timestamps::Vector{Dates.DateTime},
-    store::Nothing,
+    store::Union{Nothing, <:SimulationStore},
 )
-    isempty(result_keys) &&
-        return Dict{OptimizationContainerKey, ResultsByTime{Matrix{Float64}}}()
-
-    if res.store !== nothing
-        # In this case we have an InMemorySimulationStore.
-        store = res.store
+    vals = _read_results(res, result_keys, timestamps, store)
+    converted_vals = Dict{OptimizationContainerKey, ResultsByTime{Matrix{Float64}}}()
+    for (result_key, result_data) in vals
+        inner_converted = SortedDict(
+            (date_key, Matrix{Float64}(permutedims(inner_data.data)))
+            for (date_key, inner_data) in result_data.data)
+        converted_vals[result_key] = ResultsByTime{Matrix{Float64}, 1}(
+                        result_data.key,
+                        inner_converted,
+                        result_data.resolution,
+                        result_data.column_names)
     end
-    return _get_store_value(T, res, result_keys, timestamps, store)
+    return converted_vals
 end
 
 function _read_results(
@@ -344,6 +298,7 @@ function _read_results(
         @debug "reading results from data store"
         vals = _get_store_value(res, result_keys, timestamps, store)
     end
+    @assert all(length.(values(vals)) .== length(timestamps))
     return vals
 end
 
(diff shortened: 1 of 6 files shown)


src/simulation/realized_meta.jl
```diff
@@ -69,6 +69,8 @@ function _make_dataframe(
         row_end = row_index + last_id - first_id
         matrix[row_index:row_end, :] = array[first_id:last_id, :]
         row_index += last_id - first_id + 1
+        # Gracefully handle being given more results_by_time than we need
```
daniel-thom, line 72: When does this happen and why isn’t it an error?

src/simulation/decision_model_simulation_results.jl
```diff
@@ -305,20 +254,25 @@ function _process_timestamps(
 end
 
 function _read_results(
-    T::Type{Matrix{Float64}},
+    ::Type{Matrix{Float64}},
     res::SimulationProblemResults{DecisionModelSimulationResults},
     result_keys,
     timestamps::Vector{Dates.DateTime},
-    store::Nothing,
+    store::Union{Nothing, <:SimulationStore},
```
jd-lara, line 261: This weird interface relates to #1018 and could be fixed in this PR. @daniel-thom we discussed not doing this anymore.

src/simulation/decision_model_simulation_results.jl
```diff
@@ -90,23 +90,23 @@ get_cached_variables(res::SimulationProblemResults{DecisionModelSimulationResult
 
 get_cached_results(
```
jd-lara, line 91: I don't understand why passing the type and not the instance matters here.

src/simulation/decision_model_simulation_results.jl
```diff
@@ -336,12 +290,12 @@ function _read_results(
     end
     existing_keys = list_result_keys(res, first(result_keys))
     _validate_keys(existing_keys, result_keys)
-    cached_results = get_cached_results(res, first(result_keys))
+    cached_results = get_cached_results(res, typeof(first(result_keys)))
```
jd-lara, line 293:
```suggestion
    cached_results = get_cached_results(res, eltype(result_keys))
```

jd-lara: This change can be used all over

src/simulation/decision_model_simulation_results.jl
```diff
@@ -622,6 +546,7 @@ function load_results!(
             merge_results(store)
         end
     end
+    res.results_timestamps = new_timestamps
```
jd-lara, line 549: make and use a setter for this

jd-lara:
we should put this in the PSY4 branch to avoid massive rebases later. This improvement can wait until the PSY 4 release.

src/simulation/decision_model_simulation_results.jl
GabrielKS: I'm not sure exactly what the proposal is here. For now I'm just mimicking the other `_read_results` method and passing the store straight through

src/simulation/realized_meta.jl
jd-lara: This is an interface issue @daniel-thom discussed where we pass a store to read from or nothing when nothing signifies that it is an in-memory store which makes the code confusing since it isn't obvious.

src/simulation/decision_model_simulation_results.jl
GabrielKS: Philosophically, it makes more sense to me because we really aren't using any properties of the instance rather than its type. Practically, sometimes we don't necessarily *have* an instance, like [here](https://github.com/NREL-Sienna/PowerSimulations.jl/pull/1073/files/fbfb1ca3d4f1200b0299de04f1a5ef1d49f41a8f#diff-805b5dec2828ffb16a0f459d5e6c31d3f5bf3691a6e1562db8ad6030e81b2ca0R534).

src/simulation/decision_model_simulation_results.jl
jd-lara: This is an interface issue @daniel-thom discussed where we pass a store to read from or nothing when nothing signifies that it is an in-memory store which makes the code confusing since it isn't obvious.

GabrielKS: So you'd like me to resolve the `store` to the in-memory store or not somewhere upstream so there's always a valid store getting passed around?

src/simulation/decision_model_simulation_results.jl
daniel-thom: We’ve had this debate endlessly and have tried to use instances when possible because that’s what the Julia manual recommends. I may not be completely following the code, but don’t you have an instance for all cases where the user passed variables/duals/etc to load_results?

src/simulation/decision_model_simulation_results.jl
daniel-thom: I’m not sure that I would change this, at least not here. `nothing` also means that the user didn’t pass a store.

src/simulation/decision_model_simulation_results.jl
GabrielKS: Yeah, but consider the case where a user has loaded 2 time windows of a dual in their first call to `load_results!` and wants to load 3 time windows of a variable in their second call. Because the "what time periods do we have cached" information (`res.results_timestamps`) is across the whole results instance rather than per-variable/dual/etc., we have to now upgrade to 3 periods of that old dual. This requires fetching what duals we already have cached even if the user doesn't pass any new duals.

An alternative resolution would be allowing different variables/etc. to be cached to different numbers of time windows, and I'd be in support of that, but I was trying to limit the scope of this PR.

GabrielKS: Could I get a link to that section of the Julia manual? If this is a convention, it seems like one that needs to quite often be broken to avoid awkwardness, like we do with `get_components`.

jd-lara: @daniel-thom I think that was a preference for @kdheepak back in the day and not a hard rule

src/simulation/decision_model_simulation_results.jl
jd-lara: what does not passing a store to get result mean? I don't think that is right

src/simulation/decision_model_simulation_results.jl
jd-lara: Current guidance is captured here https://docs.julialang.org/en/v1/manual/types/#man-singleton-types

daniel-thom: I can’t remember and can try to look it up. I’m not saying it’s wrong to use types; clearly, we have to do that in get_components. I’m just saying that it you have an instance or can easily construct one, Julia prefers that you use it.

src/simulation/decision_model_simulation_results.jl
daniel-thom: We could easily be talking around each other. There are many nuances here. The user can do either of these:
```
read_variable(results, “ActivePowerVariable__ThermalStandard”)
open_store(HdfSimulationStore, simulation_store_path, "r") do store
    read_variable(results, “ActivePowerVariable__ThermalStandard”; store=store)
end
```
In the first case we have to get the store, which could in memory or h5. If it’s h5, we have to open it. It’s in-memory, we can just use it. There is one slightly awkward case one line 288.

src/simulation/decision_model_simulation_results.jl
GabrielKS: Sadly it cannot so easily because lots of existing code passes in arrays of `OptimizationContainerKey` even though they contain only one concrete type, but I'll work on finding a more elegant solution

src/simulation/decision_model_simulation_results.jl
GabrielKS: Done

src/simulation/realized_meta.jl
daniel-thom: `_read_results` should only respond with the requested data. That’s a bug.

GabrielKS: Okay, will fix

GabrielKS: https://github.com/NREL-Sienna/PowerSimulations.jl/issues/1074

src/simulation/decision_model_simulation_results.jl
jd-lara: no one does this 

```
open_store(HdfSimulationStore, simulation_store_path, "r") do store
    read_variable(results, “ActivePowerVariable__ThermalStandard”; store=store)
end
```

Assuming this patter will happen make the code way less understandable for developers. 

The most common way to access results is calling  `read_variables` or `read_realized_variables`

src/simulation/decision_model_simulation_results.jl
GabrielKS: Done

GabrielKS:
Now fixes #1074

src/simulation/decision_model_simulation_results.jl
GabrielKS: In the absence of consensus, are we okay keeping `store::Union{Nothing, <:SimulationStore}` in this non-user-facing method for the moment?

daniel-thom: I’ll propose a patch in a separate PR.

src/simulation/decision_model_simulation_results.jl
```diff
@@ -336,12 +269,24 @@ function _read_results(
     end
     existing_keys = list_result_keys(res, first(result_keys))
     _validate_keys(existing_keys, result_keys)
-    cached_results = get_cached_results(res, first(result_keys))
+    cached_results = get_cached_results(res, eltype(result_keys))
     if _are_results_cached(res, result_keys, timestamps, keys(cached_results))
-        @debug "reading results from SimulationsResults cache"
+        @debug "reading results from SimulationsResults cache"  # NOTE tests match on this
```
jd-lara, line 274: remove `# Note`

src/simulation/decision_model_simulation_results.jl
```diff
@@ -336,12 +269,24 @@ function _read_results(
...
+            filtered_vals[result_key] = ResultsByTime{valtype(inner_converted), 1}(
+                result_data.key,
+                inner_converted,
+                result_data.resolution,
+                result_data.column_names)
+        end
+        return filtered_vals
     else
-        @debug "reading results from data store"
+        @debug "reading results from data store"  # NOTE tests match on this
```
jd-lara, line 289: remove  `# Note `

jd-lara (approved):
LGTM, good effort. Minor comments to remove unnecessary comments in the code base

src/simulation/decision_model_simulation_results.jl
daniel-thom: As a developer I find this useful. It calls my attention to it.

daniel-thom (approved):
LGTM

src/simulation/decision_model_simulation_results.jl
jd-lara: ok, lets leave it

PR #1075, opened 2024-03-06
Jd/fix testsv3

jd-lara:


jd-lara:
@daniel-thom the only test failing is the services test which is expected due to the changes in the template. This won't affect the patch for the store.

daniel-thom (approved):
LGTM

PR #1078, opened 2024-03-06
Refactor store selection process

daniel-thom:
This PR attempts to address the concerns in #1018. I don’t think that issue is valid. There is no way to have a consistent interface with the in-memory store and the HDF5 store. The HDF5 store must be opened with a file path:
```
open_store(HdfSimulationStore, simulation_store_path, "r") do store
    # do stuff
end
```
The in-memory store is already open. What would it look like?
```
open(InMemorySimulationStore, _, “r”, store) do store
    # do stuff
end
```
It’s a different interface no matter how you do it.

There is a concern underlying the issue above that we have a code block that makes a non-straightforward inference about the type of the store. This addresses that.

There was another concern at https://github.com/NREL-Sienna/PowerSimulations.jl/pull/1073/files#r1512052861 which says that we should not have functions that take `Union{Nothing, <:SimulationStore}`. I don’t think this is a problem. We have code paths where the store is not passed (defaults to nothing and then is opened) and others where the store is passed.

### DIFF
--- a/src/simulation/decision_model_simulation_results.jl
+++ b/src/simulation/decision_model_simulation_results.jl
@@ -263,10 +263,7 @@ function _read_results(
     isempty(result_keys) &&
         return Dict{OptimizationContainerKey, ResultsByTime{DenseAxisArray{Float64, 2}}}()
 
-    if store === nothing && res.store !== nothing
-        # In this case we have an InMemorySimulationStore.
-        store = res.store
-    end
+    _store = try_resolve_store(store, res.store)
     existing_keys = list_result_keys(res, first(result_keys))
     _validate_keys(existing_keys, result_keys)
     cached_results = get_cached_results(res, eltype(result_keys))
@@ -287,7 +284,7 @@ function _read_results(
         return filtered_vals
     else
         @debug "reading results from data store"  # NOTE tests match on this
-        vals = _get_store_value(res, result_keys, timestamps, store)
+        vals = _get_store_value(res, result_keys, timestamps, _store)
     end
     return vals
 end
--- a/src/simulation/emulation_model_simulation_results.jl
+++ b/src/simulation/emulation_model_simulation_results.jl
@@ -197,11 +197,7 @@ function _read_results(
     len = nothing,
 )
     isempty(result_keys) && return Dict{OptimizationContainerKey, DataFrames.DataFrame}()
-    if store === nothing && res.store !== nothing
-        # In this case we have an InMemorySimulationStore.
-        store = res.store
-    end
-
+    _store = try_resolve_store(store, res.store)
     existing_keys = list_result_keys(res, first(result_keys))
     _validate_keys(existing_keys, result_keys)
     cached_results = Dict(
@@ -217,7 +213,7 @@ function _read_results(
             _get_store_value(
                 res,
                 result_keys,
-                store;
+                _store;
                 start_time = start_time,
                 len = len,
             )
--- a/src/simulation/simulation_problem_results.jl
+++ b/src/simulation/simulation_problem_results.jl
@@ -745,3 +745,9 @@ function export_optimizer_stats(
         throw(error("writing optimizer stats only supports csv or json formats"))
     end
 end
+
+# Chooses the user-passed store or results store for reading values. Either could be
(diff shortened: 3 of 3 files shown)


src/simulation/decision_model_simulation_results.jl
```diff
@@ -263,10 +263,7 @@ function _read_results(
     isempty(result_keys) &&
         return Dict{OptimizationContainerKey, ResultsByTime{DenseAxisArray{Float64, 2}}}()
 
-    if store === nothing && res.store !== nothing
-        # In this case we have an InMemorySimulationStore.
-        store = res.store
-    end
+    _store = _try_resolve_store(store, res.store)
```
daniel-thom, line 266: There is another location in emulation_model_simulation_results.jl that needs to be changed.

daniel-thom:
These tests fail.  Can you confirm that this is expected?
```
  UC with MarketBid Cost in ThermalGenerators                                                                             |     2     5             7      2.1s
  Decision Model test for Initialization with RTS GMLC system, Case 2                                                     |           1      1      2      0.8s
  Decision Model initial_conditions test for Hydro                                                                        |     3     2      1      6     10.0s
  Test for single row result variables                                                                                    |           1      2      3      3.3s
  Emulation Model initial_conditions test for Hydro                                                                       |     3     2      1      6      8.9s

```

jd-lara (approved):

jd-lara:
> These tests fail. Can you confirm that this is expected?
> 
> ```
>   UC with MarketBid Cost in ThermalGenerators                                                                             |     2     5             7      2.1s
>   Decision Model test for Initialization with RTS GMLC system, Case 2                                                     |           1      1      2      0.8s
>   Decision Model initial_conditions test for Hydro                                                                        |     3     2      1      6     10.0s
>   Test for single row result variables                                                                                    |           1      2      3      3.3s
>   Emulation Model initial_conditions test for Hydro                                                                       |     3     2      1      6      8.9s
> ```

Those should pass with the correct branch in HydroPowerSimulations and StorageSystemsSimulations

jd-lara:
> This PR attempts to address the concerns in #1018. I don’t think that issue is valid. There is no way to have a consistent interface with the in-memory store and the HDF5 store. The HDF5 store must be opened with a file path:
> 
> ```
> open_store(HdfSimulationStore, simulation_store_path, "r") do store
>     # do stuff
> end
> ```
> 
> The in-memory store is already open. What would it look like?
> 
> ```
> open(InMemorySimulationStore, _, “r”, store) do store
>     # do stuff
> end
> ```
> 
> It’s a different interface no matter how you do it.
> 
> There is a concern underlying the issue above that we have a code block that makes a non-straightforward inference about the type of the store. This addresses that.
> 
> There was another concern at https://github.com/NREL-Sienna/PowerSimulations.jl/pull/1073/files#r1512052861 which says that we should not have functions that take `Union{Nothing, <:SimulationStore}`. I don’t think this is a problem. We have code paths where the store is not passed (defaults to nothing and then is opened) and others where the store is passed.

I think that even for heavy Sienna developers understanding the pass or nothing or the store is confusing. My concern is that every time I need to look into this code I have to remember this nuances instead of having a simple consistent interface.

daniel-thom:
> I think that even for heavy Sienna developers understanding the pass or nothing or the store is confusing. My concern is that every time I need to look into this code I have to remember this nuances instead of having a simple consistent interface.

Consider this code path:
[export_results by user](https://github.com/daniel-thom/PowerSimulations.jl/blob/issue-1018/src/simulation/simulation_results.jl#L264)
    [export_results with store](https://github.com/daniel-thom/PowerSimulations.jl/blob/issue-1018/src/simulation/simulation_results.jl#L288)
        read_variable with store
        [_read_results with store](https://github.com/daniel-thom/PowerSimulations.jl/blob/issue-1018/src/simulation/decision_model_simulation_results.jl#L319)

vs the user just calling read_variable where the store defaults to nothing. We have to support both code paths. The alternative is that export_results and load_results! have to load the store hundreds of times when once is sufficient (and there is some cost to loading the store).

jd-lara:
ok, I think this is a step in the right direction. I think that at least more comments on how the process works would be useful.

PR #1062, opened 2024-02-26
Serialize System components to HDF5

daniel-thom:
Objective: The simulations postprocessing code in PowerAnalytics.jl needs to have access to the systems used by the simulations. Those systems are not serialized if the user opts out of that feature. Ensure that these are always available.

This is the PowerSimulations portion of the solution. I added comparable branches in PowerSystems.jl and InfrastructureSystems.jl. Let me know if the interfaces here meet your needs and then I’ll open sibling PRs.

PR #1080, opened 2024-03-06
Fix merge errors with #1060 and #1062

daniel-thom:
The PRs #1060 and #1062 were logically incompatible but didn’t have any git merge conflicts. This resolves the errors.

### DIFF
--- a/test/test_simulation_results.jl
+++ b/test/test_simulation_results.jl
@@ -259,6 +259,13 @@ function test_simulation_results(
         )
         results = SimulationResults(sim)
         test_decision_problem_results(results, c_sys5_hy_ed, c_sys5_hy_uc, in_memory)
+        if !in_memory
+            test_decision_problem_results_kwargs_handling(
+                dirname(results.path),
+                c_sys5_hy_ed,
+                c_sys5_hy_uc,
+            )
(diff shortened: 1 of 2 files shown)


test/test_simulation_results.jl
```diff
@@ -864,21 +891,6 @@ function test_simulation_results_from_file(path::AbstractString, c_sys5_hy_ed, c
 
     @test_throws IS.InvalidValue set_system!(results_uc, c_sys5_hy_ed)
 
-    current_file = joinpath(
```
daniel-thom, line 867: The system will always be available in the simulation_store.h5 and so this condition should be impossible.

jd-lara (approved):

test/test_simulation_results.jl
GabrielKS: Yeah, makes sense — @tengis-nrl sorry for making you write this test

GabrielKS (approved):
Looks good to me!

test/test_simulation_results.jl
tengis-nrl: It did allow me to learn more about system files. No problem there.

PR #1082, opened 2024-03-09
Use nameof instead of custom function

daniel-thom:
PSI portion of https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/339


daniel-thom:
We don’t want these changes. `strip_module_name` preserves the parametric types; `nameof` does not.

PR #1085, opened 2024-03-14
branch flow slacks

jd-lara:
This PR adds the capability to have slacks on the line flows for ACBranches. It doesn't apply to HVDC yet.

### DIFF
--- a/src/devices_models/device_constructors/branch_constructor.jl
+++ b/src/devices_models/device_constructors/branch_constructor.jl
@@ -227,15 +227,15 @@ function construct_device!(
     sys::PSY.System,
     ::ModelConstructStage,
     device_model::DeviceModel{T, StaticBranch},
-    network_model::NetworkModel{<:PM.AbstractActivePowerModel},
-) where {T <: PSY.ACBranch}
+    network_model::NetworkModel{U},
+) where {T <: PSY.ACBranch, U <: PM.AbstractActivePowerModel}
     @debug "construct_device" _group = LOG_GROUP_BRANCH_CONSTRUCTIONS
 
     devices =
         get_available_components(T, sys, get_attribute(device_model, "filter_function"))
     add_constraints!(container, RateLimitConstraint, devices, device_model, network_model)
     add_feedforward_constraints!(container, device_model, devices)
-    objective_function!(container, devices, device_model)
+    objective_function!(container, devices, device_model, U)
     add_constraint_dual!(container, sys, device_model)
     return
 end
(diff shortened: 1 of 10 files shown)


src/devices_models/device_constructors/branch_constructor.jl
```diff
@@ -315,49 +373,114 @@ end
...
 function construct_device!(
     container::OptimizationContainer,
     sys::PSY.System,
     ::ArgumentConstructStage,
-    ::DeviceModel{T, StaticBranchBounds},
+    device_model::DeviceModel{T, StaticBranchBounds},
     ::NetworkModel{<:PM.AbstractPowerModel},
-) where {T <: PSY.ACBranch} end
+) where {T <: PSY.ACBranch}
+    if get_use_slacks(device_model)
```
daniel-thom, line 433: duplicate?

src/devices_models/devices/AC_branches.jl
```diff
@@ -289,15 +317,30 @@ function add_constraints!(
...
+        if branch_name ∈ radial_branches_names
             continue
         end
-        for t in time_steps
-            constraint[r[1], t] = JuMP.@constraint(
-                get_jump_model(container),
-                var1[r[1], t]^2 + var2[r[1], t]^2 <= r[2]^2
-            )
+        if use_slacks
+            for t in time_steps
```
daniel-thom, line 330: This feels like it should be two functions.

src/devices_models/devices/AC_branches.jl
```diff
@@ -543,6 +586,14 @@ function add_constraints!(
     U <: AbstractBranchFormulation,
     V <: PM.AbstractActivePowerModel,
 }
+    error("slacks here")
```
daniel-thom, line 589: debug?

daniel-thom (changes requested):

src/devices_models/devices/AC_branches.jl
jd-lara: This is a deeper bug, this code isn't being called.

jd-lara: after testing this is stale code the whole function

src/devices_models/devices/AC_branches.jl
jd-lara: I would have to duplicate a lot more code or write functions that accept a lot of arguments

daniel-thom (approved):

PR #1090, opened 2024-03-18
Fix #1064

### LINKED ISSUE #1064: Bug in Compact UC
The Compact UC formulation with Linear $/MWh variable cost has a error in calculating the cost at no-load. This exists in both the current main and the psy4 re-factor of cost functions

https://github.com/NREL-Sienna/PowerSimulations.jl/blob/1d8e8574e472a607d98b71baec45c1a4aae7044f/src/devices_models/devices/thermal_generation.jl#L119C61-L119C78 
on main
`no_load_cost(cost::PSY.VariableCost{Float64}, ::OnVariable, d::PSY.ThermalGen, ::AbstractThermalFormulation) = PSY.get_cost(cost) * PSY.get_active_power_limits(d).min * PSY.get_base_power(d)`
on psy4
`no_load_cost(cost::PSY.LinearFunctionData, ::OnVariable, d::PSY.ThermalGen, ::AbstractThermalFormulation) = PSY.get_proportional_term(cost) * PSY.get_active_power_limits(d).min * PSY.get_base_power(d)`

The cost is multiplied by device base rather than system base, I couldn't find a way to get system other than `PSY.get_internal(d).units_info.base_value` maybe there is another method to do this.

GabrielKS:
Fixes #1064. Relies on https://github.com/NREL-Sienna/PowerSystems.jl/pull/1078. I have done what was requested in #1064, i.e., replaced the device base power with the system base power; someone with broader expertise here should verify that that is the correct thing to do in all cases.


jd-lara (approved):

jd-lara:
@GabrielKS there is an undefined function failing in the tests

jd-lara:
@GabrielKS bump the min version of PSY to 3.3.0 in the Project.toml file please

PR #1101, opened 2024-04-30
Fix check for in-memory time series

daniel-thom:
Requires
- https://github.com/NREL-Sienna/PowerSystems.jl/pull/1095
- https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/357


jd-lara:
@daniel-thom I've implemented this change in the integration branch. We can close this.

PR #1106, opened 2024-05-09
Rh/add vector to methods

jd-lara:
This PR hopefully won't conflict with PSY 4


GabrielKS (approved):
Not familiar with the context here so I can't vouch for the correctness of the code, but I am fairly confident it does not conflict with PSY4.

jd-lara:
> Not familiar with the context here so I can't vouch for the correctness of the code, but I am fairly confident it does not conflict with PSY4.

This is a bunch of bug fixes to handle HybridSystemsSimulations.jl

PR #1109, opened 2024-05-20
add new compact pwl constraints

rodrigomha:


src/utils/powersystems_utils.jl
```diff
@@ -340,6 +340,9 @@ end
 ############### Auxiliary Methods ################
 ##################################################
 
+# Compact Data does not exists anymore
```
jd-lara, line 343: @rodrigomha we need to clean up this code and add some testing probably.

jd-lara (changes requested):

PR #1115, opened 2024-06-17
Update ORDC model

rodrigomha:


src/devices_models/devices/common/add_to_expression.jl
```diff
@@ -39,7 +39,7 @@ function add_expressions!(
 } where {D <: PSY.Component}
     time_steps = get_time_steps(container)
     names = [PSY.get_name(d) for d in devices]
```
jd-lara, line 41:
```suggestion
@assert length(devices) == 1
```

src/devices_models/devices/common/add_to_expression.jl
```diff
@@ -39,7 +39,7 @@ function add_expressions!(
...
-    add_expression_container!(container, T(), D, names, time_steps)
+    add_expression_container!(container, T(), D, names, time_steps; meta = names[1])
```
jd-lara, line 42:
```suggestion
    add_expression_container!(container, T(), D, names, time_steps; meta = get_name(first(devices))
```

jd-lara (changes requested):
minor tweak but otherwise looks good to me.

jd-lara (approved):

PR #1117, opened 2024-06-21
Ignore known-benign extraneous files in simulation results

GabrielKS:
I got tired of these warnings:
Screenshot 2024-06-21 at 4 09 25 PM
and we should aim to eliminate warnings that cry wolf. Here I add a list of regexes similar to a `.gitignore` that represent files we know it is safe to ignore.

---
Happy to cherry-pick and set the base branch to something else, this was the most standard-looking branch I could find that actually compiles with PSY4.

### DIFF
--- a/src/simulation/simulation_results.jl
+++ b/src/simulation/simulation_results.jl
@@ -1,6 +1,7 @@
 function check_folder_integrity(folder::String)
     folder_files = readdir(folder)
     alien_files = setdiff(folder_files, KNOWN_SIMULATION_PATHS)
+    alien_files = filter(x -> !any(occursin.(IGNORABLE_FILES, x)), alien_files)
     if isempty(alien_files)
         return true
     else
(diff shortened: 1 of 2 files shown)


jd-lara (approved):

src/simulation/simulation_results.jl
```diff
@@ -1,6 +1,7 @@
 function check_folder_integrity(folder::String)
     folder_files = readdir(folder)
     alien_files = setdiff(folder_files, KNOWN_SIMULATION_PATHS)
+    alien_files = filter(x -> !any(occursin.(IGNORABLE_FILES, x)), alien_files)
```
daniel-thom, line 4: This has caused complaints for years. Who cares if there are extra files in the directory? Can we remove this altogether?

GabrielKS: Looks like it got added [here](https://github.com/NREL-Sienna/PowerSimulations.jl/commit/58d6e94e227b02ed4540969b0c6e4d0f147757e4), can we remember the rationale? @jd-lara

PR #1118, opened 2024-06-21
Make minimal tweaks for backwards compatibility

GabrielKS:
We don't guarantee backwards compatibility, but when it's this easy it seems like we might as well. With just these changes I was able to load some old simulation results I wanted.

---
As in https://github.com/NREL-Sienna/PowerSimulations.jl/pull/1117, happy to cherry-pick and set the base branch to something else, this was the most standard-looking branch I could find that actually compiles with PSY4.

### DIFF
--- a/src/core/definitions.jl
+++ b/src/core/definitions.jl
@@ -113,6 +113,10 @@ for enum in ENUMS
     end
 end
 
+# Special cases for backwards compatibility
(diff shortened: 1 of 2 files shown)


daniel-thom (approved):
Can you add an issue to track removing this at some point? Like for PSI 2.0.

jd-lara (approved):

PR #1123, opened 2024-07-09
Jd/decomposition support

jd-lara:


rodrigomha (approved):
LGTM: Not sure what of this is about decomposition support though

Issue #1125, opened 2024-07-16
`get_resolution` not implemented for `OptimizationProblemResults` 


jd-lara:
Issue in line: 

https://github.com/NREL-Sienna/PowerSimulations.jl/blob/b405b107547f8e6fd6d329af4259e67bb6623d28/src/simulation/decision_model_simulation_results.jl#L448

GabrielKS:
Reported by @gmantegna

GabrielKS:
A quick repro using a PowerAnalytics.jl REPL:
```julia
] activate test  # must instantiate correctly, etc.
include("test/runtests.jl")  # running the PA tests gives us an `OptimizationProblemResults` to work with
PSI.get_realized_timestamps(problem_results)  # yields `MethodError: no method matching get_resolution(::OptimizationProblemResults)`
```

GabrielKS:
The root cause is that `get_resolution(res::OptimizationProblemResults)` got moved to `IS.Optimization` and isn't being imported into PSI. It is worth noting that we now have (at least) three different `get_resolution` functions in three different packages, suggesting that we should put some effort towards standardizing how things are imported from IS:
<img width="996" alt="Screenshot 2024-07-24 at 10 23 42 AM" src="https://github.com/user-attachments/assets/c0a557c0-b5f7-4317-9e3c-515335344da9">

GabrielKS:
`get_forecast_horizon` is *four* different functions:
<img width="1043" alt="Screenshot 2024-07-24 at 10 33 35 AM" src="https://github.com/user-attachments/assets/dd99ed62-025e-4b2b-9cc9-4b7ea153f0e5">

jd-lara:
Yeah, we need to look into the method clean up here for the same function.

GabrielKS:
For now, I will just add the necessary imports to PSI to resolve this, but let's at some point come up with a principled solution to the whole importing problem [here](https://github.com/NREL-Sienna/InfrastructureSystems.jl/issues/388).

PR #1127, opened 2024-07-24
Add imports to fix bug 1125, test that

GabrielKS:
Fixes https://github.com/NREL-Sienna/PowerSimulations.jl/issues/1125 by adding some imports. At some point we should do this more holistically; see https://github.com/NREL-Sienna/InfrastructureSystems.jl/issues/388.


jd-lara (approved):
Reviewing from my phone but overall looks good. As soon as the tests pass we can merge.

PR #1139, opened 2024-09-09
Jd/param update fixes

jd-lara:


rodrigomha:
It seems that EmulationModel should also be updated for update_parameters to use SimulationState:

https://github.com/NREL-Sienna/PowerSimulations.jl/blob/f7e83f498422b85b4c7d943d91777b50e0f4f3a1/src/operation/emulation_model.jl#L419

rodrigomha (approved):
Looks good, although I'm not sure of other impacts of this change.

PR #1141, opened 2024-09-11
Export opt model setting

jd-lara:


rodrigomha (commented):
This looks okay. 

Are we fine keeping the name of the functions to be `serialize_problem` rather than `export_problem`?

jd-lara:
> This looks okay.
> 
> Are we fine keeping the name of the functions to be `serialize_problem` rather than `export_problem`?

IMO is fine, serialize is not exported.

src/operation/operation_model_interface.jl
```diff
@@ -104,13 +104,24 @@ end
 
 function solve_impl!(model::OperationModel)
     container = get_optimization_container(model)
+    model_name = get_name(model)
+    ts = get_current_timestamp(model)
+    output_dir = get_output_dir(model)
+
+    if get_export_optimization_model(get_settings(model))
+        model_output_dir = joinpath(output_dir, "optimization_model_exports")
+        if !ispath(model_output_dir)
```
daniel-thom, line 113: In case you care, this is slightly more convenient.
```suggestion
        mkpath(model_output_dir)
```

rodrigomha (approved):

PR #1145, opened 2024-09-21
Jd/multiple fixes

### LINKED ISSUE #1116: If ThermalUnit is set to must run don't create the UC constraints.
Must run don't create all the variables and constraints.

### LINKED ISSUE #1129: Must run constraints lot when calculating duals

### LINKED ISSUE #1133: Use only JSON3.jl and not JSON.jl
Minor maintenance request: currently, PowerSimulations uses both the JSON3.jl and JSON.jl JSON libraries. It seems to be the [only mainstream Sienna package](https://github.com/search?q=org%3ANREL-Sienna+%22682c06a0-de6a-54ab-a142-c8b1cf79cde6%22+project.toml&type=code) to use JSON.jl, whereas many use JSON3.jl. If there's no reason we need both, let's switch to just JSON3.jl.

### LINKED ISSUE #1140: Fix Compact Check for Market Bid Cost

### LINKED ISSUE #1148: Check monotonically increasing property quadratic
In some cases the cost functions given in the data are not monotonically increasing in the operating range of the unit resulting in strange results. E.g. cost function with negative a multipliers.

### LINKED ISSUE #1151: Bad fixed cost term
Quadratic fuel costs incorrectly uses 0.0 as a default for the `OnVariable` multiplier which prevents it from being added to the cost function

jd-lara:
fixes #1133
fixes #1140 
fixes #1129 
fixes #1116 
fixes #1148 
fixes #1151


src/devices_models/devices/common/add_to_expression.jl
rodrigomha, line 1: I'm concerned about this add to expression with must run are working properly without testing. Did we QA/QC those ones?

rodrigomha (approved):
This looks great. I have one major concern regarding the testing of the add to expression when must run is on, and I'm not exactly sure where the fix of the On variable happens now.

jd-lara:
@rodrigomha I have to add a test for that case to and fix StorageSystemsSimulations.jl

rodrigomha (approved):
This looks good. Tests are passing locally, but some weird export is happening in the Github actions

PR #1159, opened 2024-10-10
Fix Unbounded HVDC model for Copperplate

### LINKED ISSUE #1156: HVDCTwoTerminalUnbounded does not add variables for Copperplate with multiple subnetworks

rodrigomha:
Closes #1156


jd-lara (commented):
we need to add a test for this to avoid from it happening in the future.

PR #1160, opened 2024-10-11
change implementation of read dataframe

jd-lara:


src/utils/file_utils.jl
```diff
@@ -11,9 +11,7 @@ end
 Return a DataFrame from a CSV file.
 """
 function read_dataframe(filename::AbstractString)
-    open(filename, "r") do io
-        DataFrames.DataFrame(CSV.File(io))
-    end
+    return CSV.read(filename, DataFrames.DataFrame)
```
daniel-thom, line 14: Why does this fix the issue?

daniel-thom (approved):

PR #1162, opened 2024-10-13
Jd/add thermal time series

### LINKED ISSUE #1153: Add `max_active_power` time series to thermal standard
create the capability to pass a time series to Thermal formulations to model de-rates.

### LINKED ISSUE #1174: `UndefVarError` in `_add_vom_cost_to_objective!`
On the main branch (not yet released), this line:
https://github.com/NREL-Sienna/PowerSimulations.jl/blob/7fba6897b6671e8c3aa71f2ee1a66fc3e9783c22/src/devices_models/devices/common/objective_function/common.jl#L91

causes ```UndefVarError: `d` not defined```. Maybe a copypaste error suggesting a refactor to remove code duplication would be beneficial?

jd-lara:
closes #1153 partially. Still missing situations when this is mixed with dispatch models. 

If the approach in this PR is ok, I will implement for dispatch models and two stage simulations. 

closes #1174 also by necessity


jd-lara:
@rodrigomha and @sourabhdalvi take a look at the branch. I think this can benefit from more testing but this approach should suffice.

daniel-thom:
> @rodrigomha and @sourabhdalvi take a look at the branch. I think this can benefit from more testing but this approach should suffice.
> 
> @daniel-thom Can you confirm the use of `has_time_series` is adequate in this context? If this approach works we can implement for other situations where not all the devices might have time series and save the users the need to define vectors of 1.0's

I expect that this is substantially slower than the PSY 3.0 version because each call to has_time_series corresponds to one SQL query. It's probably the difference between nanoseconds and microseconds. If that is concerning, we could make a new function to get components with time series that makes a single SQL query. It could work with either get_components or an existing iterable of components.

This depends on how many times the function will be called (how many components).
```
julia> length(get_components(Component, sys))
525

julia> @time has_time_series(g)
  0.002146 seconds (31 allocations: 1.922 KiB)
```

src/parameters/add_parameters.jl
```diff
@@ -197,6 +197,10 @@ function _add_time_series_parameters!(
     device_names = String[]
     initial_values = Dict{String, AbstractArray}()
     for device in devices
+        if !PSY.has_time_series(device, ts_type, ts_name)
```
daniel-thom, line 200: This is currently slower than you want because it involves a database search. Does this need to be made faster? It's only called once during build, right? If it should be cached, should that be here in PSI or in IS?

jd-lara: We also make this calls for some parameter updates. I think we should have a cache for this in IS.

rodrigomha (approved):

PR #1169, opened 2024-10-30
Use cached time series UUID

daniel-thom:
This code was consuming a lot of time to lookup time series UUIDs when those values are already cached.

It should address https://github.com/NREL-Sienna/InfrastructureSystems.jl/issues/409


jd-lara (approved):
@joaquimg this should partially solve some of the problems from the profiling

PR #1172, opened 2024-11-08
use service name in meta

jd-lara:
this is required to fix the HSS code


rodrigomha (approved):
Lets wait for the tests though

PR #1182, opened 2024-12-09
Implement decremental market bid cost with pwl cost curve

purboday:
1. In src/devices_models/devices/common/objective_function/market_bid.jl added function _add_vom_cost_to_objective!.
2. Updated "Test Thermal Generation MarketBidCost models" in  test/test_device_thermal_generation_constructors.jl to make fixed_market_bid_cost run and pass.
3. Other diffs caused by the formatter.


test/run_market_bid_cost.jl
jd-lara, line 1: This file is useful but it shouldn't be checked out into the repo. I'd like the original test to be fixed.

purboday: Deleted the file.

src/devices_models/devices/common/objective_function/market_bid.jl
```diff
@@ -179,6 +179,36 @@ function _get_pwl_cost_expression(
     )
 end
 
+function _get_pwl_cost_expression_decremental(container::OptimizationContainer,
```
GabrielKS, line 182: Can this be implemented in a way that does not duplicate as much code from the incremental version? Maybe refactor both functions to call a helper function with the common code?

src/devices_models/devices/common/objective_function/market_bid.jl
```diff
@@ -320,6 +364,43 @@ function _add_pwl_term!(
     return pwl_cost_expressions
 end
 
+function _add_pwl_term_decremental!(container::OptimizationContainer,
```
GabrielKS, line 367: Same note about code duplication here

src/devices_models/devices/common/objective_function/market_bid.jl
```diff
@@ -275,6 +305,20 @@ end
 ######## MarketBidCost: Fixed Curves ##########
 ###############################################
 
+"""
+Check if deceremental pwl offer curve is monotonically decreasing.
+"""
+function _is_convex_decremental(pwl:: PSY.PiecewiseStepData)
```
GabrielKS, line 311: Ideally this function would go in the same place as the existing `is_convex`, in IS/PSY. Maybe `is_concave`?

src/devices_models/devices/common/objective_function/market_bid.jl
```diff
@@ -583,3 +739,59 @@ function _add_service_bid_cost!(
...
+    for t in get_time_steps(container)
+        exp = _add_proportional_term!(container, T(), d, cost_term_normalized * multiplier,
+            t)
+        add_to_expression!(container, ProductionCostExpression, exp, d, t)
+    end
+    return
+end
+
+
+function _add_vom_cost_to_objective!(container::OptimizationContainer,
```
GabrielKS, line 772: What is the difference between these two `_add_vom_cost_to_objective!` methods?

src/utils/powersystems_utils.jl
```diff
@@ -360,3 +360,7 @@ function get_deterministic_time_series_type(sys::PSY.System)
         )
     end
 end
+
+
+"""Overload get_variable for MarketBidCost. Returns nothing"""
+PSY.get_variable(value::PSY.MarketBidCost) = nothing
```
GabrielKS, line 366: What is this for? If it is needed, it should be put in PowerSystems. See [here](https://docs.julialang.org/en/v1/manual/style-guide/#avoid-type-piracy) on "type piracy," better definition [here](https://discourse.julialang.org/t/how-bad-is-type-piracy-actually/37913/6).

GabrielKS (commented):
A few unsolicited comments from the software development point of view

jd-lara:
@GabrielKS and @purboday I changed the target of this PR to a new branch to fully develop and test this feature including the simulation capability

GabrielKS:
@purboday any update on this? Would love to work towards getting this merged to the intermediate `market_bid_cost` branch so I have something somewhat stable to work off of

src/devices_models/devices/common/objective_function/market_bid.jl
```diff
@@ -9,16 +9,17 @@ function _add_pwl_variables!(
...
+    var_container = lazy_container_addition!(container, U(), T)
     # length(PiecewiseStepData) gets number of segments, here we want number of points
     break_points = PSY.get_x_coords(cost_data)
     pwlvars = Array{JuMP.VariableRef}(undef, length(break_points))
     for i in 1:(length(break_points) - 1)
         pwlvars[i] =
             var_container[(component_name, i, time_period)] = JuMP.@variable(
                 get_jump_model(container),
-                base_name = "PieceWiseLinearBlockOffer_$(component_name)_{pwl_$(i), $time_period}",
+                base_name = "$(string(U))_$(component_name)_{pwl_$(i), $time_period}",
```
GabrielKS, line 22: nitpick: `nameof` might be more appropriate here:
```suggestion
                base_name = "$(nameof(U))_$(component_name)_{pwl_$(i), $time_period}",
```

GabrielKS:
Also, It looks like something nontrivial is failing in the test suite.

GabrielKS:
Given that @purboday is gone for a while and that this is enough to build on top of, I will merge this to `NREL-Sienna:market_bid_cost` and create an issue to address the remaining issues.

GabrielKS:
Follow-up issue: https://github.com/NREL-Sienna/PowerSimulations.jl/issues/1260

src/utils/powersystems_utils.jl
jd-lara: This shouldn't be used at all. The implementation of MarketBidCost needs to be unique and avoid this call. Moreover, this is a legacy call to older implementations

PR #1196, opened 2024-12-12
Add `ComponentSelector` feature: PSI portion

GabrielKS:
Companion to and dependent on

- https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/342
- https://github.com/NREL-Sienna/PowerSystems.jl/pull/1197

. The main thing here is we want `get_components` to work on `IS.Results` such that `get_components` on a `Results` is like `get_available_components` on its attached `System` (or attached `IS.SystemData`, etc.). The implementation of this behavior is in the IS PR, but following the design set out [here](https://github.com/NREL-Sienna/PowerSystems.jl/blob/037435f4d7c0198c52e0fa29af0d05f7ee364bc7/src/get_components_interface.jl#L1), we need to define `PSY.get_components` aliases that redirect to the appropriate `IS.get_components` methods to keep the documentation user-friendly. That cannot be done in IS and (correct me if I'm wrong) it doesn't really make sense to do in PSY, so it is done here.

### DIFF
--- a/src/PowerSimulations.jl
+++ b/src/PowerSimulations.jl
@@ -398,7 +398,8 @@ import InfrastructureSystems.Optimization: read_results_with_keys, deserialize_k
 
 # PowerSystems imports
 import PowerSystems:
-    get_components, get_available_components, get_component
+    get_components, get_available_components, get_component,
+    get_groups, get_available_groups
 
 export get_name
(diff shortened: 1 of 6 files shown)


daniel-thom (approved):
LGTM

PR #1211, opened 2025-01-06
Jd/change axes checks

jd-lara:


src/simulation/simulation_state.jl
```diff
@@ -158,7 +158,12 @@ function _initialize_system_states!(
         dm_cols = get_column_names(key, get_dataset(decision_states, key))
         if has_dataset(emulator_states, key)
             em_cols = get_column_names(key, get_dataset(emulator_states, key))
-            @assert_op dm_cols == em_cols
+            @assert_op length(dm_cols) == length(em_cols)
```
daniel-thom, line 161: Quick check: asserts are for guarding against programmer errors and can be optimized out in production. Errors are for all bad inputs and are always enabled. Should these two checks be of the same type or are they correct as they are?

jd-lara: It was originally an assertion but in recent simulation cases there are situations where the data coming might* not have the same devices so I switched to an error. Before I never accounted for the possibility for this to be the result of input errors.

daniel-thom: So then should line 161 be a call to `error()`?

jd-lara: I should remove that, I put it while I was debugging.

daniel-thom (approved):

Issue #1000, opened 2023-08-28
Build times in services too large


jd-lara:
There is something that can be optimized in the construction of services. I have been observing large build times in services w.r.t to other devices 

<img width="635" alt="image" src="https://github.com/NREL-Sienna/PowerSimulations.jl/assets/16385323/3f5b3721-7e33-4e09-a0c7-f31a0f2239d0">

PR #1218, opened 2025-01-17
[DO NOT MERGE] Notes os JuMP issues

joaquimg:
main take aways:

1 - use `add_to_expression!` instead of `+=` (saves allocations of extra expressions)
2 - use `add_to_expression!(exp, number, var)` instead of `add_to_expression!(exp, number * var)` because `number * var` will allocate.
3 - prefer `(number * number) * var` or `var * (number * number)` to force computing product of floats first (very very fast) than `number * number * var` or `var * number * number` which will depend of the JuMP order of operations (one will be fast, the other slow). use style to keep all coefs in the left and variables in the right. Alsoo use parenthesis when there are multiple coefs.
4 - avoid operations (+, -, /, *, sum) between variables (or varaibles and constants) outside macros, because macros will try to allocate less.
5 - array[:, i] allocates new vectors, use `@view` to avoid this
6 - attention with type stability when: a) assining values in if-else blocks, b) re-using names inside a function
7 - attention with loop order: julia is column major so inner loops should change the leftmost index.

### DIFF
--- a/src/core/optimization_container.jl
+++ b/src/core/optimization_container.jl
@@ -1627,7 +1627,7 @@ function add_to_objective_invariant_expression!(
 ) where {T <: JuMP.AbstractJuMPScalar}
     T_cf = typeof(container.objective_function.invariant_terms)
     if T_cf <: JuMP.GenericAffExpr && T <: JuMP.GenericQuadExpr
-        # why not add_to...???
+        # TODO: why not add_to...???
         container.objective_function.invariant_terms += cost_expr
     else
         JuMP.add_to_expression!(container.objective_function.invariant_terms, cost_expr)
--- a/src/core/optimization_container.jl
+++ b/src/core/optimization_container.jl
@@ -1627,6 +1627,7 @@ function add_to_objective_invariant_expression!(
 ) where {T <: JuMP.AbstractJuMPScalar}
     T_cf = typeof(container.objective_function.invariant_terms)
     if T_cf <: JuMP.GenericAffExpr && T <: JuMP.GenericQuadExpr
+        # why not add_to...???
         container.objective_function.invariant_terms += cost_expr
     else
         JuMP.add_to_expression!(container.objective_function.invariant_terms, cost_expr)
--- a/src/devices_models/devices/AC_branches.jl
+++ b/src/devices_models/devices/AC_branches.jl
@@ -387,6 +387,7 @@ function add_constraints!(
         slack_lb = get_variable(container, FlowActivePowerSlackLowerBound(), T)
     end
 
+    # TODO: loop order
     for device in devices
         ci_name = PSY.get_name(device)
         if ci_name ∈ PNM.get_radial_branches(radial_network_reduction)
@@ -394,6 +395,7 @@ function add_constraints!(
         end
         limits = get_min_max_limits(device, RateLimitConstraint, U) # depends on constraint type and formulation type
         for t in time_steps
+            # TODO: ternary
             con_ub[ci_name, t] =
                 JuMP.@constraint(get_jump_model(container),
                     array[ci_name, t] - (use_slacks ? slack_ub[ci_name, t] : 0.0) <=
@@ -649,7 +651,7 @@ function _make_flow_expressions!(
 
     #= Leaving serial code commented out for debugging purposes in the future
     for name in branches
-        ptdf_col = ptdf[name, :]
+        ptdf_col = @view ptdf[name, :]
         branch_flow_expr[name, :] .= _make_flow_expressions!(
             jump_model,
             name,
@@ -734,7 +736,7 @@ function add_constraints!(
     jump_model = get_jump_model(container)
     for br in devices
         name = PSY.get_name(br)
-        ptdf_col = ptdf[name, :]
+        ptdf_col = @view ptdf[name, :]
         inv_x = 1 / PSY.get_x(br)
         for t in time_steps
             branch_flow[name, t] = JuMP.@constraint(
@@ -915,12 +917,12 @@ function add_constraints!(
     for br in devices
         name = PSY.get_name(br)
         inv_x = 1.0 / PSY.get_x(br)
-        flow_variables_ = flow_variables[name, :]
+        flow_variables_ = @view flow_variables[name, :]
         from_bus = PSY.get_name(PSY.get_from(PSY.get_arc(br)))
         to_bus = PSY.get_name(PSY.get_to(PSY.get_arc(br)))
-        angle_variables_ = ps_angle_variables[name, :]
-        bus_angle_from = bus_angle_variables[from_bus, :]
-        bus_angle_to = bus_angle_variables[to_bus, :]
+        angle_variables_ = @view ps_angle_variables[name, :]
+        bus_angle_from = @view bus_angle_variables[from_bus, :]
+        bus_angle_to = @view bus_angle_variables[to_bus, :]
         @assert inv_x > 0.0
         for t in time_steps
             branch_flow[name, t] = JuMP.@constraint(
--- a/src/devices_models/devices/AC_branches.jl
+++ b/src/devices_models/devices/AC_branches.jl
@@ -606,6 +606,7 @@ function _make_flow_expressions!(
             sum(
                 ptdf_col[i] * nodal_balance_expressions[i, t] for
                 i in 1:length(ptdf_col)
+                # maybe only sum if nonempy?
             )
         )
     end
--- a/src/devices_models/devices/common/add_to_expression.jl
+++ b/src/devices_models/devices/common/add_to_expression.jl
@@ -1633,10 +1633,10 @@ function add_to_expression!(
                 device_base_power,
             )
             for t in time_steps
-                fuel_expr = variable[name, t] * prop_term_per_unit * dt
                 JuMP.add_to_expression!(
                     expression[name, t],
-                    fuel_expr,
+                    prop_term_per_unit * dt,
+                    variable[name, t],
                 )
             end
         elseif value_curve isa PSY.QuadraticCurve
@@ -1709,22 +1709,33 @@ function add_to_expression!(
             for t in time_steps
                 sos_status = _get_sos_value(container, W, d)
                 if sos_status == SOSStatusVariable.NO_VARIABLE
-                    bin = 1.0
+                    JuMP.add_to_expression!(
+                        expression[name, t],
+                        P_min * prop_term_per_unit * dt,
+                    )
                 elseif sos_status == SOSStatusVariable.PARAMETER
                     param = get_default_on_parameter(d)
                     bin = get_parameter(container, param, V).parameter_array[name, t]
+                    JuMP.add_to_expression!(
+                        expression[name, t],
+                        P_min * prop_term_per_unit * dt,
+                        bin
+                    )
                 elseif sos_status == SOSStatusVariable.VARIABLE
                     var = get_default_on_variable(d)
                     bin = get_variable(container, var, V)[name, t]
+                    JuMP.add_to_expression!(
+                        expression[name, t],
+                        P_min * prop_term_per_unit * dt,
+                        bin
+                    )
                 else
                     @assert false
                 end
-                fuel_expr =
-                    variable[name, t] * prop_term_per_unit * dt +
-                    P_min * bin * prop_term_per_unit * dt
                 JuMP.add_to_expression!(
                     expression[name, t],
-                    fuel_expr,
+                    prop_term_per_unit * dt,
+                    variable[name, t],
                 )
             end
         elseif value_curve isa PSY.QuadraticCurve
--- a/src/devices_models/devices/common/duration_constraints.jl
+++ b/src/devices_models/devices/common/duration_constraints.jl
@@ -196,7 +196,8 @@ function device_duration_look_ahead!(
             lhs_off = JuMP.GenericAffExpr{Float64, JuMP.VariableRef}(0)
             for i in UnitRange{Int}(Int(t - duration_data[ix].down + 1), t)
                 if i in time_steps
-                    JuMP.add_to_expression!(lhs_off, (1 - varon[name, i]))
+                    JuMP.add_to_expression!(lhs_off, 1)
+                    JuMP.add_to_expression!(lhs_off, -1, varon[name, i])
                 end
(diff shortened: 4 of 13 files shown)


src/core/optimization_container.jl
```diff
@@ -1627,6 +1627,7 @@ function add_to_objective_invariant_expression!(
 ) where {T <: JuMP.AbstractJuMPScalar}
     T_cf = typeof(container.objective_function.invariant_terms)
     if T_cf <: JuMP.GenericAffExpr && T <: JuMP.GenericQuadExpr
+        # why not add_to...???
```
jd-lara, line 1630: before add didn't work for mixed ExpressionTypes. We can re-test

src/services_models/reserve_group.jl
```diff
@@ -60,10 +60,12 @@ function add_constraints!(
     for t in time_steps
         resource_expression = JuMP.GenericAffExpr{Float64, JuMP.VariableRef}()
         for reserve_variable in reserve_variables
-            JuMP.add_to_expression!(resource_expression, sum(reserve_variable[:, t]))
+            JuMP.add_to_expression!(resource_expression,
```
jd-lara, line 63: Is this more effective than pre-allocating the expression in 61?

src/services_models/reserve_group.jl
```diff
@@ -60,10 +60,12 @@ function add_constraints!(
...
+                JuMP.@expression(container.JuMPmodel, sum(reserve_variable[:, t])))
+            # consider a for loop to add the reserve variables
         end
         if use_slacks
-            resource_expression += slack_vars[t]
+            JuMP.add_to_expression!(resource_expression, slack_vars[t])
```
jd-lara, line 68: good catch

src/services_models/reserves.jl
```diff
@@ -161,19 +161,23 @@ function add_constraints!(
         param = get_parameter_column_refs(param_container, service_name)
         for t in time_steps
             if use_slacks
-                resource_expression = sum(reserve_variable[:, t]) + slack_vars[t]
+                resource_expression = JuMP.@expression(
```
jd-lara, line 164: this could be related to #1000

src/services_models/reserve_group.jl
jd-lara, line 1: These improvements should make it to main

src/services_models/reserves.jl
jd-lara, line 1: These ones are quite important too.

jd-lara (commented):
These are great catches @joaquimg I would like to put some of these into main sooner

src/core/optimization_container.jl
```diff
@@ -1627,7 +1627,7 @@ function add_to_objective_invariant_expression!(
 ) where {T <: JuMP.AbstractJuMPScalar}
     T_cf = typeof(container.objective_function.invariant_terms)
     if T_cf <: JuMP.GenericAffExpr && T <: JuMP.GenericQuadExpr
-        # why not add_to...???
+        # TODO: why not add_to...???
```
jd-lara, line 1630: The `add_to` here usually fails is not allowed by JuMP

src/parameters/update_cost_parameters.jl
```diff
@@ -181,8 +182,9 @@ function update_variable_cost!(
     end
     mult_ = parameter_multiplier[component_name, time_period]
     expression = get_expression(container, FuelConsumptionExpression(), T)
-    cost_expr = expression[component_name, time_period] * fuel_cost * mult_
+    cost_expr = expression[component_name, time_period] * (fuel_cost * mult_)
     add_to_objective_variant_expression!(container, cost_expr)
+    # TODO: missing _mult below?
```
jd-lara, line 187: no, one expression tracks the cost and the other the fuel consumption

src/services_models/reserve_group.jl
```diff
@@ -61,7 +61,7 @@ function add_constraints!(
         resource_expression = JuMP.GenericAffExpr{Float64, JuMP.VariableRef}()
         for reserve_variable in reserve_variables
             JuMP.add_to_expression!(resource_expression,
-                JuMP.@expression(container.JuMPmodel, sum(reserve_variable[:, t])))
+                JuMP.@expression(container.JuMPmodel, sum(@view reserve_variable[:, t])))
```
jd-lara, line 64: nice trick. @GabrielKS take note

joaquimg:
Feel free to add them in main before I do.
I was trying to catch as many possible issues as I could.

Many of these are general suggestions and should also make it to the JuMP docs.

The gain is not big in the RTS case, about 5% less allocations, but these should make a difference in larger cases with more details.

src/devices_models/devices/AC_branches.jl
```diff
@@ -387,13 +387,15 @@ function add_constraints!(
...
 
+    # TODO: loop order
     for device in devices
         ci_name = PSY.get_name(device)
         if ci_name ∈ PNM.get_radial_branches(radial_network_reduction)
             continue
         end
         limits = get_min_max_limits(device, RateLimitConstraint, U) # depends on constraint type and formulation type
         for t in time_steps
+            # TODO: ternary
```
joaquimg, line 398: ignore this: issue solved recently in MutableArithmetics

src/devices_models/devices/AC_branches.jl
```diff
@@ -606,6 +608,7 @@ function _make_flow_expressions!(
             sum(
                 ptdf_col[i] * nodal_balance_expressions[i, t] for
                 i in 1:length(ptdf_col)
+                # maybe only sum if nonempy?
```
joaquimg, line 611: ignore

src/devices_models/devices/common/add_to_expression.jl
```diff
@@ -1633,10 +1633,10 @@ function add_to_expression!(
                 device_base_power,
             )
             for t in time_steps
-                fuel_expr = variable[name, t] * prop_term_per_unit * dt
                 JuMP.add_to_expression!(
                     expression[name, t],
```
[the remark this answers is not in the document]
joaquimg: for more info:
`JuMP.add_to_expression!(exp1, float_coef * jump_var)` is worse because it creates an extra temporary expression `float_coef * jump_var`

src/devices_models/devices/common/duration_constraints.jl
```diff
@@ -196,11 +196,12 @@ function device_duration_look_ahead!(
             lhs_off = JuMP.GenericAffExpr{Float64, JuMP.VariableRef}(0)
             for i in UnitRange{Int}(Int(t - duration_data[ix].down + 1), t)
                 if i in time_steps
-                    JuMP.add_to_expression!(lhs_off, (1 - varon[name, i]))
+                    JuMP.add_to_expression!(lhs_off, 1)
+                    JuMP.add_to_expression!(lhs_off, -1, varon[name, i])
```
joaquimg, line 200: the previous allocates more because `(1 - varon[name, i])` becomes a new expression

src/devices_models/devices/electric_loads.jl
```diff
@@ -87,8 +87,8 @@ function add_constraints!(
         name = PSY.get_name(d)
         pf = sin(atan((PSY.get_max_reactive_power(d) / PSY.get_max_active_power(d))))
         reactive = get_variable(container, U(), V)[name, t]
-        real = get_variable(container, ActivePowerVariable(), V)[name, t] * pf
-        constraint[name, t] = JuMP.@constraint(jump_model, reactive == real)
+        real = get_variable(container, ActivePowerVariable(), V)[name, t]
+        constraint[name, t] = JuMP.@constraint(jump_model, reactive == real * pf)
```
joaquimg, line 91: the product outside the macro lead to additional allocations

src/parameters/update_cost_parameters.jl
```diff
@@ -131,7 +132,7 @@ function update_variable_cost!(
     end
     mult_ = parameter_multiplier[component_name, time_period]
     variable = get_variable(container, get_variable_type(attributes)(), T)
-    gen_cost = variable[component_name, time_period] * mult_ * cost_data * base_power * dt
+    gen_cost = variable[component_name, time_period] * (mult_ * cost_data * base_power * dt)
```
joaquimg, line 135: this is interesting:
the parenthesis forces the floats the be multiplied before and simplified into a single float. the previous one requires a bunch of operations in an expression. not an allocation issue, but slower than multiplying floats.

src/parameters/update_cost_parameters.jl
joaquimg: yup not much to do here, as the two need to be kept different

src/services_models/reserve_group.jl
joaquimg: not sure I got what you mean

src/services_models/reserves.jl
```diff
@@ -224,12 +228,12 @@ function add_constraints!(
             cons[name, t] =
                 JuMP.@constraint(
                     jump_model,
-                    var_r[name, t] <= param[t] * requirement * max_participation_factor
+                    var_r[name, t] <= (requirement * max_participation_factor) * param[t]
```
joaquimg, line 231: rule of thumb: first multiply floats and multiply them with jump variables

src/services_models/reserves.jl
```diff
@@ -473,7 +479,7 @@ function add_constraints!(
                 PSY.get_active_power_limits(d).min +
                 (reserve_response_time - startup_time) * minutes_per_period * ramp_limits.up
         else
-            reserve_limit = 0
+            reserve_limit = 0.0
```
joaquimg, line 482: type stability as the if block would return a float and `0` is ana Int

src/core/optimization_container.jl
joaquimg: This is weird, do you have an example? It does not happen in RTS. Maybe it was an old JuMP bug?

src/devices_models/devices/AC_branches.jl
```diff
@@ -387,13 +387,15 @@ function add_constraints!(
         slack_lb = get_variable(container, FlowActivePowerSlackLowerBound(), T)
     end
 
+    # TODO: loop order
```
joaquimg, line 390: this is interesting, and PSI uses mixed patterns.

Julia is column based, so if you vector is `array[name, t]`
you should iterate as:
```julia
for t in T
    for name in names
        # do something with array[name, t]
    end
end
```
which is equivalent to:
```julia
for t in T, name in names
    # do something with array[name, t]
end
```
In the case of sums:
```julia
sum(array[name, t] for t in T, name in names)
```
equivalent to:
```julia
sum(array[name, t] for t in T for name in names)
```
almost equivalent to:
```julia
sum(sum(array[name, t] for name in names) for t in T )
```

joaquimg: so names should typically be the inner loop and t the outer loop.
but if the operation inside the loop is complex both are almost the same.
I would keep the pattern so style would lead you to the correct choices in thigh loops.

src/devices_models/devices/AC_branches.jl
```diff
@@ -914,12 +917,12 @@ function add_constraints!(
...
-        flow_variables_ = flow_variables[name, :]
+        flow_variables_ = @view flow_variables[name, :]
         from_bus = PSY.get_name(PSY.get_from(PSY.get_arc(br)))
         to_bus = PSY.get_name(PSY.get_to(PSY.get_arc(br)))
-        angle_variables_ = ps_angle_variables[name, :]
-        bus_angle_from = bus_angle_variables[from_bus, :]
-        bus_angle_to = bus_angle_variables[to_bus, :]
+        angle_variables_ = @view ps_angle_variables[name, :]
+        bus_angle_from = @view bus_angle_variables[from_bus, :]
+        bus_angle_to = @view bus_angle_variables[to_bus, :]
```
joaquimg, line 925: not sure about these. the `@view` saves many allocations but the the iteration is worst in the loop below as there are many cache misses.

src/core/optimization_container.jl
jd-lara: This only happens when we have data that mixes PWL and Quadratic cost functions.

jd-lara:
> Feel free to add them in main before I do. I was trying to catch as many possible issues as I could.
> 
> Many of these are general suggestions and should also make it to the JuMP docs.
> 
> The gain is not big in the RTS case, about 5% less allocations, but these should make a difference in larger cases with more details.
> 
> My plan was to make cleaner PR's after this pass.

I can test this branch on the super large systems and report back w.r.t. to main.

src/devices_models/devices/AC_branches.jl
jd-lara: Yeah, I originally wrote all the loops in the right order and then had to modify some depending on the context.

src/services_models/agc.jl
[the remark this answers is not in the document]
joaquimg: POI is much better nowadays.
Moreover, there is room for improvement, so let me know whenever you want to test stuff and bring it back

src/core/optimization_container.jl
joaquimg: If there is any case I can test with those features I'd be happy to.

joaquimg:
That would be nice, just build time benchmarking should be enough. no need to run solvers.

src/devices_models/devices/common/objective_function/market_bid.jl
```diff
@@ -137,11 +137,12 @@ function _get_pwl_cost_expression(
     name = PSY.get_name(component)
     pwl_var_container = get_variable(container, PieceWiseLinearBlockOffer(), T)
     gen_cost = JuMP.AffExpr(0.0)
-    cost_data = PSY.get_y_coords(cost_data)
-    for (i, cost) in enumerate(cost_data)
+    _cost_data = PSY.get_y_coords(cost_data)
+    for (i, cost) in enumerate(_cost_data)
```
joaquimg, line 141: type instability

jd-lara: This is fixed in main now

jd-lara:
@joaquimg feel free to close this one and audit the current main. I added a lot of the comments already, some uses of the @view I left out.

jd-lara:
@joaquimg thanks for the notes, all of these minus the POI changes are now in main or in open PR's. I will close this PR now.

PR #1223, opened 2025-01-23
update cost functions for thermal dispatch no min

### LINKED ISSUE #1221: Fix PWL cost production when using ThermalDispatchNoMin
**Describe the bug**
PWL cost expressions are not working correctly with `ThermalDispatchNoMin` formulation.

**To Reproduce**
```
using PowerSystems
using PowerSimulations
using HiGHS
using PowerSystemCaseBuilder
using Dates

sys = build_system(PSISystems, "modified_RTS_GMLC_DA_sys")

# Update all thermal components in the system
# update_all_components!(sys)

template = ProblemTemplate(NetworkModel(PTDFPowerModel))
set_device_model!(template, ThermalStandard, ThermalDispatchNoMin)
set_device_model!(template, Line, StaticBranchBounds)
set_device_model!(template, TapTransformer, StaticBranchBounds)
set_device_model!(template, Transformer2W, StaticBranchBounds)
set_device_model!(template, PowerLoad, StaticPowerLoad)

solver = optimizer_with_attributes(HiGHS.Optimizer, "mip_rel_gap" => 0.01)
problem = DecisionModel(template, sys;  
                        optimizer = solver, 
                        horizon = Hour(1), 
                        optimizer_solve_log_print=true,
                        calculate_conflict=true,
                        store_variable_names=true,
                        detailed_optimizer_stats=false,
                        );

build!(problem; output_dir = mktempdir())

solve!(problem)
```

We observe that when units are being dispatch at maximum, the ProductionCost is zero. This is likely happening because in `ThermalDispatchNoMin` we extend the data to include the zero, but we are not updating properly the expressions with the new point. This is also happening in the objective function.

**Expected behavior**
Should create the proper expression to properly capture the cost.

**Screenshots**
Image

rodrigomha:
Closes #1221

### DIFF
--- a/src/utils/powersystems_utils.jl
+++ b/src/utils/powersystems_utils.jl
@@ -342,6 +342,29 @@ function is_time_variant(cost_function::PSY.FuelCurve{PSY.PiecewisePointCurve})
     return isa(PSY.get_fuel_cost(cost_function), IS.TimeSeriesKey)
 end
 
+function create_temporary_cost_function_in_system_per_unit(
+    original_cost_function::PSY.CostCurve,
+    new_data::PSY.PiecewiseLinearData,
+)
+    return PSY.CostCurve(
+        PSY.PiecewisePointCurve(new_data),
+        PSY.UnitSystem.SYSTEM_BASE,
+        PSY.get_vom_cost(original_cost_function),
+    )
+end
+
+function create_temporary_cost_function_in_system_per_unit(
+    original_cost_function::PSY.FuelCurve,
+    new_data::PSY.PiecewiseLinearData,
+)
+    return PSY.FuelCurve(
+        PSY.PiecewisePointCurve(new_data),
+        PSY.UnitSystem.SYSTEM_BASE,
+        PSY.get_fuel_cost(original_cost_function),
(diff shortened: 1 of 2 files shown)


src/utils/powersystems_utils.jl
```diff
@@ -342,6 +342,29 @@ function is_time_variant(cost_function::PSY.FuelCurve{PSY.PiecewisePointCurve})
     return isa(PSY.get_fuel_cost(cost_function), IS.TimeSeriesKey)
 end
 
+function create_temporary_cost_function_in_system_per_unit(
+    original_cost_function::PSY.CostCurve,
```
jd-lara, line 346: @rodrigomha do we need to make this a more specific type?

rodrigomha: It depends. The function just replace the data in the value curve with a new value curve, always assuming that the new point data comes in per-unit, so it does not really matter whatever the original data was from (Incremental or Point or Average).

Alternatively, using something like `PSY.CostCurve{PSY.PiecewisePointCurve}` and `PSY.FuelCurve{PSY.PiecewisePointCurve}` would work for the purposes of this PR, makes to only work if the data comes from a PointCurve.

PR #1232, opened 2025-01-31
update decision state for OnVariable and ActivePowerVariable

m-bossart:


m-bossart:
This change seems to give the desired behavior of the generator outage and then being brought back online: 
image

m-bossart:
When applying the event, if the status is zero moving forward from a status change, then the OnVariable and ActivePower should be forced to zero in the decision model as well?

jd-lara (approved):
This PR was tested synchronously and it fixed the observed problem

PR #1234, opened 2025-02-05
Jd/jump perf

jd-lara:


jd-lara:
Tests fail due to dependencies but this is my local tests 

image

daniel-thom (approved):
LGTM

PR #1238, opened 2025-02-07
AC power flow loss factor auxiliary variable

rbolgaryn:
Here we add the reading of the loss factors after a successful AC power flow calculation in `PowerFlows` as a new auxiliary variable `PowerFlowLossFactors`.

* Create a new auxiliary variable `PowerFlowLossFactors`
* Read the value from the results in `PowerFlowData` (only if `calc_loss_factors` was set to true in ACPowerFlow)
* Test that the variable is present when `calc_loss_factors` is `true` and that it is not present when it is `false`. The values of loss factors are tested only in `PowerFlows`

This PR depends on https://github.com/NREL-Sienna/PowerFlows.jl/pull/98


test/test_simulation_results.jl
```diff
@@ -349,6 +352,10 @@ function test_decision_problem_results_values(
         @test size(v) == (12, 6)
     end
 
+    loss_factors = read_variable(results_ed, PowerFlowLossFactors)
```
GabrielKS, line 355:
```suggestion
    loss_factors = read_aux_variable(results_ed, PowerFlowLossFactors)
```

src/network_models/power_flow_evaluation.jl
```diff
@@ -385,6 +385,8 @@ _get_pf_result(::Type{PowerFlowLineActivePowerFromTo}, pf_data::PFS.PowerFlowDat
     PFS.get_branch_activepower_flow_from_to(pf_data)
 _get_pf_result(::Type{PowerFlowLineActivePowerToFrom}, pf_data::PFS.PowerFlowData) =
     PFS.get_branch_activepower_flow_to_from(pf_data)
+_get_pf_result(::Type{PowerFlowLossFactors}, pf_data::PFS.PowerFlowData) =
+    [PFS.penalty_factors(aux_vars) for aux_vars in pf_data.aux_variables]
```
GabrielKS, line 389: First, a bit of syntactic sugar:
```suggestion
    PFS.penalty_factors.(pf_data.aux_variables)
```

GabrielKS: Second, maybe the issue you're having here is that this is a vector of vectors rather than a matrix? Might need to do some reshaping (see `hcat`, `vcat`, etc.).

GabrielKS: Third, it looks like we now have two different types of things called aux variables. That's going to get confusing.

src/PowerSimulations.jl
```diff
@@ -244,6 +244,7 @@ export PowerFlowVoltageAngle
 export PowerFlowVoltageMagnitude
 export PowerFlowLineReactivePowerFromTo, PowerFlowLineReactivePowerToFrom
 export PowerFlowLineActivePowerFromTo, PowerFlowLineActivePowerToFrom
+export PowerFlowLossFactors
```
GabrielKS, line 247: I've recently become aware that when we add new auxiliary variables we also have to add them to a documentation page. See https://github.com/NREL-Sienna/PowerSimulations.jl/pull/1248

test/test_simulation_results.jl
```diff
@@ -1030,6 +1030,7 @@ end
             power_flow_evaluation =
             ACPowerFlow(;
                 exporter = PSSEExportPowerFlow(:v33, pf_path; write_comments = true),
+                calc_loss_factors = true,
```
GabrielKS, line 1033: Also test with `false` — what is the desired behavior? That it doesn't create the `PowerFlowLossFactors` aux vars at all or that it fills them with `nothing`? I think probably the first one

test/test_simulation_results.jl
```diff
@@ -1044,6 +1045,24 @@ end
     first_result = first(thermal_results)
     last_result = last(thermal_results)
 
+    println(list_aux_variable_keys(results_ed))
```
GabrielKS, line 1048: remove `println` before merging

test/test_simulation_results.jl
```diff
@@ -1044,6 +1045,24 @@ end
...
+
+    loss_factors = read_aux_variable(results_ed, PowerFlowLossFactors, ACBus)
+
+    # here we check if the loss factors are stored in the results, the values are tested in PowerFlows.jl
+    @test loss_factors !== nothing
+
+    loss_factors = first(
```
GabrielKS, line 1055: why are you getting the `loss_factors` two different ways?

test/test_simulation_results.jl
```diff
@@ -1044,6 +1045,24 @@ end
...
+        values(
+            PSI.read_results_with_keys(results_ed,
+                [PSI.AuxVarKey(PowerFlowLossFactors, ACBus)]),
+        ),
+    )
+
+    # here we check if the loss factors are stored in the results, the values are tested in PowerFlows.jl
+    @test loss_factors !== nothing
+    @test nrow(loss_factors) == 576
```
GabrielKS, line 1064: Where does `576` come from? Can we do `... == length(get_components(...` instead to avoid hard-coding this?

src/network_models/power_flow_evaluation.jl
```diff
@@ -385,6 +386,8 @@ _get_pf_result(::Type{PowerFlowLineActivePowerFromTo}, pf_data::PFS.PowerFlowDat
     PFS.get_branch_activepower_flow_from_to(pf_data)
 _get_pf_result(::Type{PowerFlowLineActivePowerToFrom}, pf_data::PFS.PowerFlowData) =
     PFS.get_branch_activepower_flow_to_from(pf_data)
+_get_pf_result(::Type{PowerFlowLossFactors}, pf_data::PFS.PowerFlowData) =
+    pf_data.loss_factors === nothing ? nothing : pf_data.loss_factors
```
GabrielKS, line 390: nitpick: use `isnothing`

test/test_simulation_results.jl
rbolgaryn: It is 48 * 12 (see also line 1067). Do you have an idea where it is best to read it from?

test/test_simulation_results.jl
rbolgaryn: Just to make sure it works. Should I remove one of them?

test/test_simulation_results.jl
rbolgaryn: Added to a different test where we do not expect it to be there

rbolgaryn: The aux variable is not added then at all

src/network_models/power_flow_evaluation.jl
```diff
@@ -385,6 +390,8 @@ _get_pf_result(::Type{PowerFlowLineActivePowerFromTo}, pf_data::PFS.PowerFlowDat
     PFS.get_branch_activepower_flow_from_to(pf_data)
 _get_pf_result(::Type{PowerFlowLineActivePowerToFrom}, pf_data::PFS.PowerFlowData) =
     PFS.get_branch_activepower_flow_to_from(pf_data)
+_get_pf_result(::Type{PowerFlowLossFactors}, pf_data::PFS.PowerFlowData) =
+    isnothing(pf_data.loss_factors) ? nothing : pf_data.loss_factors
```
GabrielKS, line 394: wait doesn't this just simplify to
```suggestion
    pf_data.loss_factors
```
?

test/test_simulation_results.jl
GabrielKS: Yeah, I would say that the correspondence between `read_aux_variable` and `read_results_with_keys` is/ought to be tested elsewhere, so you should only need one of them

test/test_simulation_results.jl
GabrielKS: `48 * 12` works for me

GabrielKS (commented):
Looks good aside from this one comment and the redundant test

src/network_models/power_flow_evaluation.jl
rbolgaryn: this should be fine because of the change to only include the PowerFlowLossFactors in the list if it needs to be there and will not be nothing

rbolgaryn:
The PR should now be consistent with the PR https://github.com/NREL-Sienna/PowerFlows.jl/pull/97

jd-lara:
we can merge once we release PowerFlows.jl with the changes needed for the PowerFlow calculation

jd-lara:
@rbolgaryn I made a release https://github.com/JuliaRegistries/General/pull/127716 of PowerFlows.jl you can update this PR to include a minimum version of PowerFlows.jl compatible with this.

PR #1241, opened 2025-02-07
Fix HVDC Lossless model bug in Copperplate network

### LINKED ISSUE #1227: PSI Simulation Build fails: `NetworkModel` definition as `CopperPlatePowerModel` with `TwoTerminalHVDCLine` in the network
**Describe the bug**
Build fails when formulating the `NetworkModel` as `CopperPlatePowerModel` in a System that contains B2B HVDC links modelled as `TwoTerminalHVDCLine` and formulated as `HVDCTwoTerminalLossless`.

Suspect this is due to the bus/busses sitting behind the B2B `TwoTerminalHVDCLine` being "isolated" busses.
 
Stacktrace screenshot:

Image

**To Reproduce**
PSY `4.4.1` and PSI `0.29.2`

`template = PSI.ProblemTemplate(
                        PSI.NetworkModel(
                        PSI.CopperPlatePowerModel,
                        use_slacks = true, 
                                    )
                                )`

**Expected behavior**
Successful build considering this is the simplest `NetworkModel`.

rodrigomha:
Closes #1227

### DIFF
--- a/src/devices_models/devices/TwoTerminalDC_branches.jl
+++ b/src/devices_models/devices/TwoTerminalDC_branches.jl
@@ -418,6 +418,47 @@ function add_constraints!(
     return
 end
 
+function add_constraints!(
+    container::OptimizationContainer,
+    ::Type{T},
+    devices::Union{Vector{U}, IS.FlattenIteratorWrapper{U}},
+    ::DeviceModel{U, HVDCTwoTerminalLossless},
+    network_model::NetworkModel{CopperPlatePowerModel},
+) where {T <: FlowRateConstraint, U <: PSY.TwoTerminalHVDCLine}
+    time_steps = get_time_steps(container)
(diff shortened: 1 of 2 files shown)


src/devices_models/devices/TwoTerminalDC_branches.jl
```diff
@@ -418,6 +418,47 @@ function add_constraints!(
     return
 end
 
+function add_constraints!(
```
jd-lara, line 421: Why are we doing this? If its copper plate it should be no-opt

rodrigomha: But we model HVDC lines in different subnetworks right? Copperplate balance for each subnetwork using the ref buses, so we want to model HVDC lines between different subnetworks.

PR #1243, opened 2025-02-10
Mb/cross horizon outages

m-bossart:
This PR allows for outages that are not cleared within the span of the state horizon. This should work for both `GeometricDistributionForcedOutage` and `TimeSeriesForcedOutage`

@jd-lara In this implementation I preprocess the outage data so a one indicates a change in the status (see below). Where should this preprocessing occur in `PowerSimulations`? or can the API require data that represents changes and not the status? 

@llavin13 if you want to test this now with your system, use the function below to preprocess the data. This will likely change in the final implementation. 

```
function preprocess_outage_data(outage_data)
    status_change_data = zeros(length(outage_data))
    for ix in 1:length(outage_data)
        if ix > 1
            if outage_data[ix] != outage_data[ix - 1]
                status_change_data[ix - 1] = 1.0
            end
        end
    end
    return status_change_data
end
# This is ignored for now in this simulation setup
o_data = fill!(Vector{Float64}(undef, 8784), 0.0)
o_data[20:51] .= 1.0
outage_time_series1 = TimeArray(dates_ts, preprocess_outage_data(o_data))
```

### DIFF
--- a/src/simulation/simulation_state.jl
+++ b/src/simulation/simulation_state.jl
@@ -297,9 +297,9 @@ function update_decision_state!(
     simulation_time::Dates.DateTime,
     ::ModelStoreParams,
 ) where {T <: PSY.Component}
-    event_ocurrence_data =
+    event_occurrence_data =
         get_system_state_data(state, AvailableStatusChangeCountdownParameter(), T)
-    event_ocurrence_values = get_last_recorded_value(event_ocurrence_data)
+    event_occurrence_values = get_last_recorded_value(event_occurrence_data)
     # This is required since the data for outages (mttr and λ) is always assumed to be on hourly resolution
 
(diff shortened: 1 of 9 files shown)


src/simulation/simulation_state.jl
```diff
@@ -250,80 +250,85 @@ end
 
 function _get_time_to_recover(
     event::PSY.GeometricDistributionForcedOutage,
+    event_model::EventModel,
     simulation_time,
-    length,
 )
-    return PSY.get_mean_time_to_recovery(event)
+    timeseries_mapping = event_model.timeseries_mapping
```
jd-lara, line 256: this is the functionality of `has_time_series`

jd-lara (approved):

Issue #1168, opened 2024-10-30
Slackdown is not added to bus Q balance


yonghongchen8:
**If this is a question, something isn't working or an idea please start a Q&A discussion in the [Discussion tab](https://github.com/NREL-Sienna/PowerSimulations.jl/discussions)**
Open a bug report only if you can provide the details below

**Describe the bug**
In set_network_model!(template_ed, NetworkModel( ACPPowerModel,use_slacks = true ),)
The slackdown variable is not added to bus Q balance constraint.

**To Reproduce**
Paste the code we can run to reproduce the error you are seeing

using Pkg
Pkg.activate(".")
using PowerSystemCaseBuilder
using PowerSystems
using PowerSimulations
using Ipopt
using PowerFlows
using TimeSeries
using HiGHS
using Logging
using Revise
using Xpress
#in pkg dev PowerSimulations

const PSY = PowerSystems
const PSB = PowerSystemCaseBuilder
const PSI = PowerSimulations

#sys = build_system(PSISystems, "c_sys5_pjm")
PSB.clear_all_serialized_systems() 

c_sys5_pjm_da = PSB.build_system(PSISystems, "modified_RTS_GMLC_DA_sys"; skip_serialization = true)
#PSY.transform_single_time_series!(c_sys5_pjm_da, Hour(24), Hour(24))
#c_sys5_pjm_rt = PSB.build_system(PSISystems, "modified_RTS_GMLC_RT_sys"; skip_serialization = true)
c_sys5_pjm_rt = PSB.build_system(PSISystems, "modified_RTS_GMLC_DA_sys"; skip_serialization = true)
PSY.transform_single_time_series!(c_sys5_pjm_rt, Hour(2), Hour(1))

#sys_DA = build_system(PSISystems, "modified_RTS_GMLC_DA_sys"; skip_serialization = true)
#sys_RT = build_system(PSISystems, "modified_RTS_GMLC_RT_sys"; skip_serialization = true)

template_uc = template_unit_commitment()
pop!(template_uc.branches, :TwoTerminalHVDCLine)

set_network_model!(template_uc,    NetworkModel(PTDFPowerModel;),)
#set_network_model!(template_uc,    NetworkModel(DCPLLPowerModel;),)
set_device_model!(template_uc, ThermalStandard, ThermalStandardUnitCommitment)

template_ed = deepcopy(template_uc)
set_device_model!(template_ed, ThermalStandard, ThermalBasicDispatch)
[...]

PR #1167, opened 2024-10-29
Add VSC Two Terminal Loss Model

rodrigomha:
This PR will close: NREL-Sienna/PowerOperationsModels.jl#47 NREL-Sienna/PowerOperationsModels.jl#48 NREL-Sienna/PowerOperationsModels.jl#50 NREL-Sienna/PowerOperationsModels.jl#49

PR #1263, opened 2025-03-04
Jd/continuous uc model

jd-lara:
@rodrigomha I added code here to kinda make general implementations to approximate any function f with PWL. This could be useful in #1167 and #1168. 

Still needs some additional code. Opening as draft


jd-lara:
@abdelrahman-ayad I am going to close this. You pushed your results into the branch... please don't that moving forward.

PR #1264, opened 2025-03-04
Sm/dev scuc

SebastianManriqueM:
These are the modifications done so far to implement **Security Constrained PTDF** formulation. Currently it builds and run the simulation. Still there is an error during the simulation execution: "Could not write the result for PTDFOutagesBranchFLow" since the PTDFOutagesBranchFLow has 3 dimensions.


m-bossart:
Needs tests

src/core/network_model.jl
```diff
@@ -190,9 +191,72 @@ function instantiate_network_model(
...
+        throw(
+            IS.ConflictingInputsError(
+                "The provided PTDF Matrix has reduced radial branches and mismatches the network \\
+                model specification reduce_radial_branches = false. Set the keyword argument \\
+                reduce_radial_branches = true in your network model"),
+        )
+    end
+
+    if model.reduce_radial_branches &&
+       model.PTDF_matrix.network_reduction.reduction_type == PNM.NetworkReductionTypes.WARD       #FIX THIS LINE (SATURDAY)
```
m-bossart, line 217: Leftover comment? Is this done?

src/core/network_model.jl
```diff
@@ -99,6 +99,7 @@ function add_dual!(model::NetworkModel, dual)
     return
 end
 
+#TODO Check this with new ward reduction implementation
```
m-bossart, line 102: Extra comment

src/core/network_model.jl
```diff
@@ -190,9 +191,72 @@ function instantiate_network_model(
...
 
-    # Add new checkd for the LODF
+    get_PTDF_matrix(model).subnetworks
+    model.subnetworks = deepcopy(get_PTDF_matrix(model).subnetworks)
+    if length(model.subnetworks) > 1
+        @debug "System Contains Multiple Subnetworks. Assigning buses to subnetworks."
+        _assign_subnetworks_to_buses(model, sys)
+    end
+
+    #LODF
```
m-bossart, line 238: Unnecessary comment

m-bossart:
@SebastianManriqueM  I think in general when there are open questions about a PR, it is preferable to ask those questions in the pull request (you can add comments referencing specific lines to inform the review or leave more general comments) rather than include a bunch of comments in the code that will need to be removed before the PR is merged.

src/core/network_model.jl
SebastianManriqueM: Done!

src/core/network_model.jl
SebastianManriqueM: Done!

src/core/network_model.jl
SebastianManriqueM: I want to check this with you. For me the logic here looks good, but I left this comment since it is still in line 166 of Jose's Branch (jd/transmissions_SCUC). I'm not sure if you detected some issue here that needs to be addresed.

Issue #1260, opened 2025-03-01
Clean up market bid decremental code


GabrielKS:
Follow-up issue to address the comments in https://github.com/NREL-Sienna/PowerSimulations.jl/pull/1182 (@purboday). In particular:

- [ ] Remove code duplication from `_get_pwl_cost_expression_decremental`
- [ ] Remove code duplication from `_add_pwl_term_decremental!`
- [ ] `PSY._is_convex_decremental` asymmetry with `IS.is_convex`
- [ ] Remove code duplication from `_add_vom_cost_to_objective!`
- [ ] Address type piracy in `PSY.get_variable(value::PSY.MarketBidCost) = nothing`
- [ ] Use `nameof` instead of `string` for working with type names
- [x] Make tests pass

purboday:
Thanks @GabrielKS for the feedback and apologies for not getting to these earlier. I will work on these and create a new PR.

PR #1265, opened 2025-03-05
Fix issues in https://github.com/NREL-Sienna/PowerSimulations.jl/issues/1260

purboday:
- Removed code duplication from market_bid.jl.
- Removed `get_variable(cost::MarketBidCost) = nothing`
- Changed `string()` to `nameof()`
- Removed `is_convex_decremental` and used newly created `is_concave` from IS. Depends on related PRs in [IS](https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/445) and [PowerSystems](https://github.com/NREL-Sienna/PowerSystems.jl/pull/1295)

### DIFF
--- a/src/devices_models/devices/common/add_to_expression.jl
+++ b/src/devices_models/devices/common/add_to_expression.jl
@@ -40,7 +40,11 @@ function add_expressions!(
     names = String[]
     found_quad_fuel_functions = false
     for d in devices
-        fuel_curve = PSY.get_variable(PSY.get_operation_cost(d))
+        op_cost = PSY.get_operation_cost(d)
+        if op_cost isa PSY.MarketBidCost
+            continue
+        end
+        fuel_curve = PSY.get_variable(op_cost)
         if fuel_curve isa PSY.FuelCurve
             push!(names, PSY.get_name(d))
             if !found_quad_fuel_functions
@@ -1615,7 +1619,11 @@ function add_to_expression!(
     resolution = get_resolution(container)
     dt = Dates.value(resolution) / MILLISECONDS_IN_HOUR
     for d in devices
-        var_cost = PSY.get_variable(PSY.get_operation_cost(d))
+        op_cost = PSY.get_operation_cost(d)
+        if op_cost isa PSY.MarketBidCost
+            continue
+        end
+        var_cost = PSY.get_variable(op_cost)
         if !(var_cost isa PSY.FuelCurve)
             continue
         end
@@ -1688,7 +1696,11 @@ function add_to_expression!(
     resolution = get_resolution(container)
     dt = Dates.value(resolution) / MILLISECONDS_IN_HOUR
     for d in devices
-        var_cost = PSY.get_variable(PSY.get_operation_cost(d))
+        op_cost = PSY.get_operation_cost(d)
+        if op_cost isa PSY.MarketBidCost
+            continue
+        end
+        var_cost = PSY.get_variable(op_cost)
         if !(var_cost isa PSY.FuelCurve)
             continue
         end
--- a/src/devices_models/devices/common/add_to_expression.jl
+++ b/src/devices_models/devices/common/add_to_expression.jl
@@ -10,6 +10,9 @@ function _ref_index(::NetworkModel{AreaPTDFPowerModel}, device_bus::PSY.ACBus)
     return PSY.get_name(PSY.get_area(device_bus))
 end
 
+_get_variable_if_exists(::PSY.MarketBidCost) = nothing
+_get_variable_if_exists(cost::PSY.OperationalCost) = PSY.get_variable(cost)
+
 function add_expressions!(
     container::OptimizationContainer,
     ::Type{T},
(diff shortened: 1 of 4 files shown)


purboday:
@GabrielKS @jd-lara @rodrigomha since I do not have access to add reviewers, explicitly tagging to keep everyone updated.

src/devices_models/devices/common/add_to_expression.jl
```diff
@@ -40,7 +40,11 @@ function add_expressions!(
     names = String[]
     found_quad_fuel_functions = false
     for d in devices
-        fuel_curve = PSY.get_variable(PSY.get_operation_cost(d))
+        op_cost = PSY.get_operation_cost(d)
+        if op_cost isa PSY.MarketBidCost
```
GabrielKS, line 44: Instead of this, why not a helper function that does the multiple dispatch:
```julia
_get_variable_if_exists(::PSY.MarketBidCost) = nothing
_get_variable_if_exists(cost::PSY.OperationalCost) = PSY.get_variable(cost)

fuel_curve = _get_variable_skip_market_bid(op_cost)
```

src/devices_models/devices/common/objective_function/market_bid.jl
```diff
@@ -2,6 +2,12 @@
 ################# PWL Variables ##################
 ##################################################
 
+"""
+Enumeration representing the different types of offer curves for a MarketBidCost. 
+Used as an argument for calling the helper functions.
+"""
+@enum MarketBidOfferType Incremental Decremental
```
GabrielKS, line 9: Convention is to use `IS.@scoped_enum`

src/devices_models/devices/common/objective_function/market_bid.jl
```diff
@@ -642,12 +619,41 @@ function _add_variable_cost_to_objective!(container::OptimizationContainer,
...
+    offer_type::MarketBidOfferType,
+    ::U) where {T <: VariableType,
+    U <: AbstractDeviceFormulation}
     time_steps = get_time_steps(container)
     initial_time = get_initial_time(container)
     incremental_cost_curves = PSY.get_incremental_offer_curves(cost_function)
     decremental_cost_curves = PSY.get_decremental_offer_curves(cost_function)
-    if !(isnothing(incremental_cost_curves))
-        error("Component $(component_name) is not allowed to participate as a supply.")
+    if offer_type == Incremental
```
GabrielKS, line 649: Could do this using multiple dispatch with a helper function

src/devices_models/devices/common/objective_function/market_bid.jl
```diff
@@ -642,12 +619,41 @@ function _add_variable_cost_to_objective!(container::OptimizationContainer,
...
+    ::T,
+    component::PSY.Component,
+    cost_function::PSY.MarketBidCost,
+    offer_type::MarketBidOfferType,
+    ::U) where {T <: VariableType,
+    U <: AbstractDeviceFormulation}
     time_steps = get_time_steps(container)
     initial_time = get_initial_time(container)
     incremental_cost_curves = PSY.get_incremental_offer_curves(cost_function)
     decremental_cost_curves = PSY.get_decremental_offer_curves(cost_function)
```
GabrielKS, line 648: No need to get both if you're only going to use one

src/devices_models/devices/common/objective_function/market_bid.jl
```diff
@@ -340,22 +341,61 @@ function _add_pwl_term!(
...
+    ::V) where {T <: PSY.Component, U <: VariableType,
+    V <: AbstractDeviceFormulation}
+    if offer_type == Incremental
+        cost_data = PSY.get_incremental_offer_curves(cost_function)
+        data = _get_pwl_data(container, component, cost_data)
+        cost_is_convex = PSY.is_convex(data)
+        offer_variable = PieceWiseLinearBlockOffer
+        constraint_variable = PieceWiseLinearBlockOfferConstraint
+        _get_pwl_cost_expression_function = _get_pwl_cost_expression
+    elseif offer_type == Decremental
```
GabrielKS, line 388: Maybe do the `Incremental`/`Decremental`-specific stuff in the incremental/decremental-specific functions that call this helper function and pass the data into here as arguments?

src/devices_models/devices/common/objective_function/market_bid.jl
GabrielKS: although I think you might be able to eliminate this if you use my other suggestions

GabrielKS (changes requested):

src/devices_models/devices/common/add_to_expression.jl
purboday: Updated this portion to use multiple dispatch.

src/devices_models/devices/common/objective_function/market_bid.jl
purboday: Removed this.

src/devices_models/devices/common/objective_function/market_bid.jl
purboday: Passed them as arguments.

GabrielKS (approved):
LGTM assuming tests pass locally with PSY dev'd. Thanks!

PR #1271, opened 2025-03-11
Add LCC physical model

mcllerena:
TODO:

- [x] Docs formulation

### DIFF
--- a/src/core/constraints.jl
+++ b/src/core/constraints.jl
@@ -335,6 +335,113 @@ struct PhaseAngleControlLimit <: ConstraintType end
 struct InterfaceFlowLimit <: ConstraintType end
 struct HVDCFlowCalculationConstraint <: ConstraintType end
 
+
+"""
+Struct to create the constraint that calculates the Rectifier DC line voltage.
+
+```math
+v_d^r = \\frac{3}{\\pi}N^r \\left( \\sqrt{2}\frac{a^r v_\\text{ac}^r}{t^r}\\cos{\\alpha^r}-X^r I_d \\right)
+```
+"""
+struct HVDCRectifierDCLineVoltageConstraint <: ConstraintType end
+
+"""
+Struct to create the constraint that calculates the Inverter DC line voltage.
+
+```math
+v_d^i = \\frac{3}{\\pi}N^i \\left( \\sqrt{2}\frac{a^i v_\\text{ac}^i}{t^i}\\cos{\\gamma^i}-X^i I_d \\right)
+```
+"""
+struct HVDCInverterDCLineVoltageConstraint <: ConstraintType end
+
+"""
+Struct to create the constraint that calculates the Rectifier Overlap Angle.
+
+```math
+\\mu^r = \\arccos \\left( \\alpha^r - \\frac{\\sqrt{2} I_d X^r t^r}{a^r v_\\text{ac}^r} \\right) - \\alpha^r
+```
+"""
+struct HVDCRectifierOverlapAngleConstraint <: ConstraintType end
+
+"""
+Struct to create the constraint that calculates the Inverter Overlap Angle.
+
+```math
+\\mu^i = \\arccos \\left( \\gamma^i - \\frac{\\sqrt{2} I_d X^i t^r}{a^i v_\\text{ac}^i} \\right) - \\gamma^i
+```
+"""
+struct HVDCInverterOverlapAngleConstraint <: ConstraintType end
+
+"""
+Struct to create the constraint that calculates the Rectifier Power Factor Angle.
+
+```math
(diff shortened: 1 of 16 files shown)


rodrigomha:
Tests are passing locally with psy5 branch for PowerSystems and PowerNetworkMatrices

src/devices_models/devices/common/add_to_expression.jl
```diff
@@ -456,6 +456,130 @@ function add_to_expression!(
...
+}
+    var = get_variable(container, U(), V)
+    nodal_expr = get_expression(container, T(), PSY.ACBus)
+    radial_network_reduction = get_radial_network_reduction(network_model)
+    for d in devices
+        bus_no_from =
+            PNM.get_mapped_bus_number(radial_network_reduction, PSY.get_arc(d).from)
+        for t in get_time_steps(container)
+            flow_variable = var[PSY.get_name(d), t]
+            _add_to_jump_expression!(nodal_expr[bus_no_from, t], flow_variable, 1.0)
```
rodrigomha, line 546: @mcllerena are we fine with ReactivePower being added positive in both ends? Can you check in the test if the numbers for the reactive power are added correctly. It should consume reactive power in both ends.

mcllerena: Yes, if we look at the converters from the AC side, they drawn reactive power so it should be -1.0. 
After testing it I get better voltage profiles. I'll fix this.

test/test_device_lcc.jl
```diff
@@ -0,0 +1,66 @@
+@testset "LCC HVDC System Tests" begin
+    sys5 = build_system(PSISystems, "2Area 5 Bus System")
```
jd-lara, line 2: we should add this to PSB

src/core/constraints.jl
```diff
@@ -335,6 +335,111 @@ struct PhaseAngleControlLimit <: ConstraintType end
...
 struct HVDCFlowCalculationConstraint <: ConstraintType end
 
+"""
+Struct to create the constraint that calculates the Rectifier DC line voltage.
+
+```math
+v_d^r = \\frac{3}{\\pi}N^r \\left( \\sqrt{2}\frac{a^r v_\\text{ac}^r}{t^r}\\cos{\\alpha^r}-X^r I_d \\right)
+```
+"""
+struct HVDCRectifierDCLineVoltageConstraint <: ConstraintType end
```
jd-lara, line 345: @rodrigomha how does this definitions conflict with the VSC model?

src/core/definitions.jl
```diff
@@ -37,7 +37,8 @@ const JuMPFloatMatrix = DenseAxisArray{Float64, 2}
 const JuMPFloatArray = DenseAxisArray{Float64}
 const JuMPVariableArray = DenseAxisArray{JuMP.VariableRef}
 
-const TwoTerminalHVDCTypes = Union{PSY.TwoTerminalHVDCLine, PSY.TwoTerminalVSCDCLine}
+const TwoTerminalHVDCTypes =
```
jd-lara, line 40: we can eliminate this and now use an abstract type in PSY 5

src/core/variables.jl
```diff
@@ -242,6 +242,111 @@ Docs abbreviation: ``y``
 """
 struct HVDCActivePowerReceivedToVariable <: VariableType end
 
+"""
+Struct to dispatch the creation of HVDC Received Reactive Flow From Bus Variables
+
+Docs abbreviation: ``x^r``
+"""
+struct HVDCReactivePowerReceivedFromVariable <: VariableType end
```
jd-lara, line 250: I don't think we need HVDC specific variables for these reactive power variables on the AC side. We can consolidate with the AC power ones.

src/core/constraints.jl
rodrigomha: The model is completely different for the VSC here

src/core/variables.jl
rodrigomha: Agree here, but do we use a meta for ReactivePowerVariable, `from` & `to`?

src/core/constraints.jl
jd-lara: Yes, I understand that in some cases the model is different but we should still look into consolidating the variables as much as possible

rodrigomha: I can take a look if we need to have a constraint here in the VSC, I would suggest replacing this for something like`HVDCFromVoltageConstraint`, that could probably be re-used. I think the approach for now is to use the variables we have and once I start the VSC rewrite, I will look at all the variables and constraints to reuse, and modify accordingly at that point.

src/core/variables.jl
jd-lara: we can use the same model as AC Reactive Power and the key will have the DeviceType

src/core/definitions.jl
jd-lara: I just merged these definitions for PSY5. I am ok merging this and making the changes in a follow up PR

PR #1284, opened 2025-03-25
Implement time variable startup and shutdown cost functionality

GabrielKS:
Here I:

- Add parameters to represent `MarketBidCost`'s (possibly multi-start) startup and shutdown cost time series
- Add these time-varying costs to the optimization problem where relevant
- Write shutdown cost and fuel cost parameters to the results.

This PR does not:

- Write startup costs to the results — it will be a larger effort to support writing 3D parameters (time, component, hot vs. warm vs. cold) (@daniel-thom?)
- Handle other time-varying aspects of `MarketBidCost`.


src/core/optimization_container.jl
```diff
@@ -665,6 +665,12 @@ function build_impl!(
                     device_model,
                     transmission_model,
                 )
+                construct_market_bid!(
```
GabrielKS, line 668: Note from @jd-lara: should be doing this in the formulations even though it's lots of duplication

src/devices_models/devices/common/objective_function/market_bid.jl
```diff
@@ -1,3 +1,65 @@
...
+    if _consider_shutdown_time_series(container, model)
+        shutdown_devices =
+            filter(x -> _has_market_bid_cost(x) && _has_shutdown_time_series(x), devices)
+        if length(shutdown_devices) > 0
+            add_parameters!(container, ShutdownCostParameter, shutdown_devices, model)
+        end
+    end
+end
+
+# TODO stub for later implementation
```
GabrielKS, line 55: Not required for this PR

src/devices_models/devices/thermal_generation.jl
```diff
@@ -104,20 +104,35 @@ has_multistart_variables(::PSY.ThermalMultiStart, ::ThermalMultiStartUnitCommitm
...
+# Validators: check that the types match (formulation is optional) and redirect to the simpler methods
+start_up_cost(cost, ::PSY.ThermalGen, ::T, ::Union{AbstractThermalFormulation, Nothing} = nothing) where {T <: StartVariable} =
+    start_up_cost(cost, T())
+start_up_cost(cost, ::PSY.ThermalMultiStart, ::T, ::ThermalMultiStartUnitCommitment = ThermalMultiStartUnitCommitment()) where {T <: MultiStartVariable} =
+    start_up_cost(cost, T())
+
+# Implementations: given a single number, tuple, or StartUpStages and a variable, do the right thing
+# Single number to anything
+start_up_cost(cost::Float64, ::StartVariable) = cost
+# TODO in the case where we have a single number startup cost and we're modeling a multi-start, do we set all the values to that number?
```
GabrielKS, line 122: looking for input on this

src/devices_models/devices/thermal_generation.jl
```diff
@@ -104,20 +104,35 @@ has_multistart_variables(::PSY.ThermalMultiStart, ::ThermalMultiStartUnitCommitm
...
+
+# 3-tuple to anything
+start_up_cost(cost::NTuple{3, Float64}, ::T) where {T <: VariableType} =
+    start_up_cost(StartUpStages(cost), T())
+
+# `StartUpStages` to anything
+start_up_cost(cost::StartUpStages, ::ColdStartVariable) = cost.cold
+start_up_cost(cost::StartUpStages, ::WarmStartVariable) = cost.warm
+start_up_cost(cost::StartUpStages, ::HotStartVariable) = cost.hot
+# TODO in the opposite case, do we want to get the maximum or the hot?
```
GabrielKS, line 134: looking for input on this — how to extract a single startup cost from a multistart?

src/parameters/update_cost_parameters.jl
```diff
@@ -15,80 +16,168 @@ function _update_parameter_values!(
...
+            attributes,
+            container,
+            initial_forecast_time,
+            horizon,
+        )
+    end
+    return
+end
+
+# TODO storing this old code here for now, it is currently unreachable
```
GabrielKS, line 39: not needed for this PR

src/parameters/update_cost_parameters.jl
```diff
@@ -15,80 +16,168 @@ function _update_parameter_values!(
...
+# This is implicated in the MarketBidCost time varying incremental/decremental implementation, which will be addressed later
```
GabrielKS, line 40: not needed for this PR

GabrielKS:
I am now calling the startup and shutdown cost parameter creation from `construct_device!` rather than at the level above that. Currently have added it only for thermals, if we like this approach I can add it for the rest.

src/core/optimization_container.jl
GabrielKS: Done — I am now calling the startup and shutdown cost parameter creation from `construct_device!` rather than at the level above that. Currently have added it only for thermals, if we like this approach I can add it for the rest (in this PR).

src/core/parameters.jl
```diff
@@ -177,22 +177,13 @@ function get_column_names(key::ParameterKey, c::ParameterContainer)
...
-    array::AbstractArray{Float64},
-    ::JuMP.Model,
-    value::Float64,
-    ixs::Tuple,
-)
-    array[ixs...] = value
-    return
-end
-
+const _ValidDataParamEltypes = Union{Float64, IS.FunctionData, Tuple{Vararg{Float64}}}
```
rodrigomha, line 180: Any reason for this to be here and with an `_`?

src/devices_models/devices/common/objective_function/common.jl
```diff
@@ -29,11 +79,18 @@ function add_shut_down_cost!(
...
+
         for t in get_time_steps(container)
-            _add_proportional_term!(container, U(), d, cost_term * multiplier, t)
+            my_cost_term = get_shutdown_cost_value(
+                container,
+                d,
+                t,
+                is_time_variant(PSY.get_shut_down(PSY.get_operation_cost(d))),
+            )
+            # iszero(my_cost_term) && continue  # TODO do we want this?
```
rodrigomha, line 90: Probably? If the cost is zero, why bother adding it?

src/devices_models/devices/thermal_generation.jl
rodrigomha: I think this is okay. The MultiStart modeling is always expecting the three numbers right?

src/devices_models/devices/thermal_generation.jl
rodrigomha: I think that was the approach before. Using the maximum was our intended version

src/parameters/update_cost_parameters.jl
```diff
@@ -114,30 +203,45 @@ function _update_pwl_cost_expression(
...
+# General case
+# (TODO this seemed rather decrepit before I got here and made big changes, make sure I didn't break anything)
 function update_variable_cost!(
     container::OptimizationContainer,
-    parameter_array::JuMPFloatArray,
-    parameter_multiplier::JuMPFloatArray,
-    attributes::CostFunctionAttributes{Float64},
-    component::T,
+    parameter_array::DenseAxisArray{T},
+    parameter_multiplier::JuMPFloatArray,  # TODO is the multiplier always `JuMPFloatArray`?
```
rodrigomha, line 216: I think so, the multiplier should be an array of floats, at least that is our intended purpose

src/parameters/update_cost_parameters.jl
```diff
@@ -114,30 +203,45 @@ function _update_pwl_cost_expression(
...
-    dt = Dates.value(resolution) / MILLISECONDS_IN_HOUR
-    base_power = get_base_power(container)
+) where {T, U <: PSY.Component}
     component_name = PSY.get_name(component)
-    cost_data = parameter_array[component_name, time_period]  # TODO is this a new-style cost?
-    if iszero(cost_data)
-        return
-    end
+    cost_data = parameter_array[component_name, time_period]
+    # TODO is this really correct? If it's zero now we can just leave it, it won't keep the value from last round?
```
rodrigomha, line 223: This function is getting called for each time period, if the cost term is zero, then why bother adding it to the variant terms or the cost expression

src/core/parameters.jl
GabrielKS: Not really, was just trying to communicate the private-ness of it. But I'll remove the underscore.

src/parameters/add_parameters.jl
```diff
@@ -264,6 +264,28 @@ function _add_time_series_parameters!(
...
+_get_time_series_name(::ShutdownCostParameter, device::PSY.Component, ::DeviceModel) =
+    get_name(PSY.get_shut_down(PSY.get_operation_cost(device)))
+
+# Layer of indirection to figure out what eltype we expect to find in various time series
+# (we could just read the time series and figure it out dynamically if this becomes too brittle)
+_get_expected_time_series_eltype(::T) where {T <: ParameterType} = Float64
+_get_expected_time_series_eltype(::StartupCostParameter) = NTuple{3, Float64}
+
+# Lookup that defines which variables the ObjectiveFunctionParameter corresponds to
+_param_to_vars(::FuelCostParameter, ::AbstractDeviceFormulation) = (ActivePowerVariable,)
```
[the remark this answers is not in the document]
GabrielKS: Not sure either, but before I wrote this it was just hardcoded to use `ActivePowerVariable` in all cases

src/devices_models/device_constructors/thermalgeneration_constructor.jl
```diff
@@ -1,3 +1,16 @@
+# TODO call this for non-thermal generators too?
```
jd-lara, line 1: we can but that is probably a later thing to do. Other devices might have other implementations of what attributed to look for.

src/parameters/update_cost_parameters.jl
jd-lara: yes, multipliers are always floats

jd-lara (approved):
The PR looks pretty good to me. We can merge as is and continue the other details in subsequent PRs

Issue #1269, opened 2025-03-07
Ability to run AreaBalancePowerModel with multiple asynch areas (e.g., connected by HVDC)


llavin13:
Both Jarrad and I seem to hit this issue running `AreaBalancePowerModel` on TI/ERCOT (https://github.nrel.gov/SIIP/TexasInterconnection.jl/ if have access internally). 

```
Error: Simulation not built InfrastructureSystems.ConflictingInputsError("AreaBalancePowerModel doesn't support systems with multiple asynchronous areas")
```

The proximate cause appears to be 3 nodes that are seen as their own subnetworks at the end of `TwoTerminalHVDC` lines.

Just want to make sure this behavior is as intended; and, if not, what the fix is!

jarry7:
Adding a screenshot of the topology for further information.  

<img width="247" alt="Image" src="https://github.com/user-attachments/assets/2ac439bc-2c2b-4937-9812-8c5c4a508be3" />
<img width="247" alt="Image" src="https://github.com/user-attachments/assets/88802964-a6fe-4bad-9985-dbadeeb0c494" />

PR #1285, opened 2025-03-28
hvdc area fix

jd-lara:
Still needs testing but this should address the issue with the area/synch regions mismatch @llavin13 reported in #1269. 

One situation to note here is that we need to throw meaningful errors when users add TransmissionInterfaces to AreaBalance models because it doesn't make sense since there aren't any line flows calculated.


rodrigomha:
Lets check https://github.com/NREL-Sienna/PowerSimulations.jl/issues/1288 in this PR

PR #1293, opened 2025-04-03
fix: Adding implementation for `set_device_model` for `HydroReservoir`

pesap:
…s service.


pesap:
@abdelrahman-ayad , we can close this PR since we changed the formulation to Device. It should be able to go directly there.

PR #1300, opened 2025-04-18
Fix #1298, fix #1297 for thermals, test that

### LINKED ISSUE #1297: `MarketBidCost` formulation incorrectly refers to no load cost rather than min gen cost
When creating the proportional cost for the `OnVariable` given a `MarketBidCost`, we read the `no_load_cost`:
https://github.com/NREL-Sienna/PowerSimulations.jl/blob/ba846178b5c4dc377f9bbc655bc725ee30f06db3/src/devices_models/devices/thermal_generation.jl#L93. This is incorrect: we should instead be reading the min gen cost, labeled as `initial_input`.

For non-time series data, the `initial_input` is embedded within the variable cost curves:
```julia
get_initial_input(get_incremental_offer_curves(market_bid_op_cost))
get_initial_input(get_decremental_offer_curves(market_bid_op_cost))
```
For time series data, we cannot currently create time series of `CostCurve`, so we have to represent the `initial_input` separately:
```julia
get_incremental_initial_input(market_bid_op_cost)
get_decremental_initial_input(market_bid_op_cost)
```
. `PSY.get_variable_cost(device, cost; start_time, len)` is meant to provide a layer of abstraction over this inconsistency, though it might not be working properly right now.

### LINKED ISSUE #1298: `AbstractThermalUnitCommitment` formulations add spurious constant to cost expression
In this method:
https://github.com/NREL-Sienna/PowerSimulations.jl/blob/eddad7a76cfd3b5e95d7d19df59be22abcd8255c/src/devices_models/devices/common/objective_function/common.jl#L107-L133
we get the `OnVariable`'s proportional cost `C`, add the term `C*OnVariable` to the `ProductionCostExpression` if the unit is not `must_run`, then add a constant term `C` to the `ProductionCostExpression` no matter what. This leads to incorrect `ProductionCostExpression`s with that proportional cost represented twice:

Image

I believe lines 123-129 here should be in an `else` block.

### LINKED ISSUE #1299: `MarketBidCost` loads do not account for initial input on `PowerLoadDispatch` formulation
When a load (e.g., `InterruptiblePowerLoad`) is given a `MarketBidCost` and simulated under the `PowerLoadDispatch` formulation, no `OnVariable` proportional cost is created to model the `initial_input`.

GabrielKS:
Fixes #1298
Fixes #1297

Does not handle the time varying case or non-thermal devices (e.g., does not fix #1299). Adds tests sufficient to fail in the absence of these fixes, but I still think we could use more rigorous testing of this kind of thing in general.

### DIFF
--- a/src/devices_models/devices/common/objective_function/common.jl
+++ b/src/devices_models/devices/common/objective_function/common.jl
@@ -116,17 +116,12 @@ function add_proportional_cost!(
         for t in get_time_steps(container)
             cost_term = proportional_cost(container, op_cost_data, U(), d, V(), t)
             iszero(cost_term) && continue
-            if !PSY.get_must_run(d)
-                exp = _add_proportional_term!(container, U(), d, cost_term * multiplier, t)
-                add_to_expression!(container, ProductionCostExpression, exp, d, t)
+            exp = if PSY.get_must_run(d)
+                cost_term * multiplier  # note we do not add this to the objective function
+            else
+                _add_proportional_term!(container, U(), d, cost_term * multiplier, t)
             end
-            add_to_expression!(
-                container,
-                ProductionCostExpression,
-                cost_term * multiplier,
-                d,
-                t,
-            )
+            add_to_expression!(container, ProductionCostExpression, exp, d, t)
         end
     end
     return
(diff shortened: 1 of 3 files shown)


GabrielKS:
I can change the target to `main` if desired

jd-lara (approved):

PR #1301, opened 2025-04-20
Add Security Constrained G-1 Transmission Models

purboday:
1. Created new thermal model, variable and constraints for the generator post-contingency power balance with reserves.
2. Added new service constructor to add the post contingency flow expressions and constraints.
3. Use either the max active power or the dispatched power in constraint formulation through an attribute in the service model.
4. Use attribute to determine the reserve that need to be deployed for a generator contingency.
5. Added test.


src/core/formulations.jl
```diff
@@ -171,6 +171,11 @@ struct PTDFPowerModel <: AbstractPTDFModel end
 Linear active power approximation using the power transfer distribution factor [PTDF](https://nrel-sienna.github.io/PowerNetworkMatrices.jl/stable/tutorials/tutorial_PTDF_matrix/) matrix and line outage distribution factors [LODF](https://nrel-sienna.github.io/PowerNetworkMatrices.jl/stable/tutorials/tutorial_LODF_matrix/) for Line outages. If exists, the rating b is considered as the branch power limit for post-contingency flows, otherwise the standard reting is considered.
 """
 struct SecurityConstrainedPTDFPowerModel <: AbstractSecurityConstrainedPTDFModel end
+
+"""
+Linear active power approximation using the power transfer distribution factor [PTDF](https://nrel-sienna.github.io/PowerNetworkMatrices.jl/stable/tutorials/tutorial_PTDF_matrix/) matrix for Generator outages. If exists, the rating b is considered as the branch power limit for post-contingency flows, otherwise the standard reting is considered.
+"""
+struct SecurityConstrainedG1PTDFPowerModel <: AbstractSecurityConstrainedPTDFModel end
```
jd-lara, line 178: This should be a ThermalModel not a Network one

purboday: Noted.

src/PowerSimulations.jl
```diff
@@ -338,6 +343,7 @@ export FuelConsumptionExpression
 export ActivePowerRangeExpressionLB
 export ActivePowerRangeExpressionUB
 export PTDFPostContingencyBranchFlow
+export PTDFPostContingencyBranchFlowWithReserves
```
jd-lara, line 346: Why is a different variable needed for the reserve case than the others?

jd-lara (commented):
@purboday you don't need to create repeated variables for the reserves vs no-reserves cases

src/PowerSimulations.jl
purboday: Since the flow variable was keyed by the branch, if both N-1 and G-1 were used simultaneously it would create a conflict.

PR #1303, opened 2025-04-23
Implement time variable `MarketBidCost` for thermals (only incremental)

GabrielKS:
In this sequel to https://github.com/NREL-Sienna/PowerSimulations.jl/pull/1284, I add the remaining `MarketBidCost` time variability for thermal generators: the "min gen cost" `OnVariable` coefficient and piecewise incremental variable costs. Extension to non-thermal generators, and implementation of time variable _decremental_ costs for loads, are left to a future PR, as is the writing of greater-than-two-dimensional parameters to the results.


GabrielKS:
Progress update: I've successfully validated time-varying min gen cost, breakpoints, and slopes for incremental thermal `MarketBidCost`!

Remaining TODOs in this PR:

- [ ] Solve bug #1318 for real. There is a stopgap in there right now, but it's not great
- [ ] Verify the rest of the formulation, e.g., should the generator's active power limits play a role at all?
- [ ] Rebase (or more realistically at this point, merge) to `psy5`
- [ ] Some code cleanup in the implementation. I don't think we need that bespoke subtype of `FunctionData` after all, there's some incremental vs. decremental stuff that could be done more cleanly, some lingering TODOs, etc.
- [ ] A few more tests that I doubt will catch anything huge
- [ ] Some test refactoring and cleanup

Things I propose to do in subsequent PR(s):

- Handle the non-thermal case
- Handle the decremental case
- Write 3D parameters to results (currently blocked by https://github.com/NREL-Sienna/PowerSimulations.jl/issues/1304 @daniel-thom)

Miscellaneous notes:

- It seems like sometimes, rather than reporting an infeasibility, HiGHS will solve a relaxation of the problem and silently give you that. For instance, I was getting fractional `OnVariable` values. Might be worth digging into and reproducing more minimally

src/core/definitions.jl
```diff
@@ -10,6 +10,7 @@ const RUN_OPERATION_MODEL_TIMER = TimerOutputs.TimerOutput()
 const RUN_SIMULATION_TIMER = TimerOutputs.TimerOutput()
 
 # Type Alias for JuMP containers
+const JuMPOrFloat = Union{JuMP.AbstractJuMPScalar, Float64}
```
jd-lara, line 13: We should open an issue to handle this better in the future. These unions of abstract + Float64 are weird. I wonder what the compiler does.

src/core/parameters.jl
```diff
@@ -159,7 +159,7 @@ function get_parameter_values(
     param_array::DenseAxisArray,
     multiplier_array::DenseAxisArray,
 )
-    return jump_value.(param_array) .* multiplier_array
+    return (.*).(jump_value.(param_array), multiplier_array)
```
jd-lara, line 166: Why this syntax?

src/devices_models/devices/common/objective_function/common.jl
```diff
@@ -85,16 +85,18 @@ function add_shut_down_cost!(
...
-                is_time_variant(PSY.get_shut_down(PSY.get_operation_cost(d))),
+                add_as_time_variant,
             )
             iszero(my_cost_term) && continue
-            exp = _add_proportional_term_variant!(
-                container, U(), d, my_cost_term * multiplier, t)
+            exp = _add_proportional_term_maybe_variant!(
+                Val(add_as_time_variant), container, U(), d, my_cost_term * multiplier,
+                t)
             # add_to_expression!(container, ProductionCostExpression, exp, component, t)  # TODO do we want this?
```
jd-lara, line 100: Do we want to fix this too?

src/devices_models/devices/common/objective_function/market_bid.jl
```diff
@@ -2,10 +2,37 @@
...
+get_initial_input_maybe_decremental(::Val{false}, device::PSY.StaticInjection) =
+    PSY.get_incremental_initial_input(PSY.get_operation_cost(device))
+
+get_offer_curves_maybe_decremental(::Val{true}, device::PSY.StaticInjection) =
+    get_input_offer_curves(PSY.get_operation_cost(device))
+get_offer_curves_maybe_decremental(::Val{false}, device::PSY.StaticInjection) =
+    get_output_offer_curves(PSY.get_operation_cost(device))
+
+# Dictionaries to handle more incremental (false) vs. decremental (true) cases
+const SLOPE_PARAMS = Dict(
```
jd-lara: Define types

jd-lara: Also if this const is only used here do `_NAME_OF_CONST`. We should include this in the style guide

src/devices_models/devices/common/objective_function/market_bid.jl
```diff
@@ -2,10 +2,37 @@
...
+    false => IncrementalPiecewiseLinearSlopeParameter,
+    true => DecrementalPiecewiseLinearSlopeParameter)
+const BREAKPOINT_PARAMS = Dict(
```
jd-lara: ditto from above

src/devices_models/devices/common/objective_function/market_bid.jl
```diff
@@ -14,40 +41,241 @@ _consider_startup_time_series(
...
     ::DeviceModel{T, D},
 ) where {T, D} =
     haskey(get_variables(container), VariableKey(StopVariable, T))
 
+_consider_parameter(
+    ::IncrementalCostAtMinParameter,
+    container::OptimizationContainer,
+    ::DeviceModel{T, D},
+) where {T, D} =
+    haskey(get_variables(container), VariableKey(OnVariable, T))
```
jd-lara: we should have a method to do this without creating `VariableKey` objects. @m-bossart I think was looking at a similar application.

src/devices_models/devices/common/objective_function/market_bid.jl
```diff
@@ -14,40 +41,241 @@ _consider_startup_time_series(
...
+            Float64, device, initial_input, " for initial_input",
+        )
+    end
+end
+
+function validate_mbc_breakpoints_slopes(device::PSY.StaticInjection, decremental::Bool)
+    offer_curves = get_offer_curves_maybe_decremental(Val(decremental), device)
+    device_name = get_name(device)
+    is_ts = is_time_variant(offer_curves)
+    location = is_ts ? " in time series $(get_name(offer_curves))" : " in offer curves"
```
jd-lara: Can we do this with a type instead of a string and dispatch on a type in the following function. It seems we are abusing a bit the use of Val here.

GabrielKS: This isn't Val-based dispatch but it does create the string whether it's needed or not — TODO refactor

src/core/definitions.jl
GabrielKS: -> https://github.com/NREL-Sienna/PowerSimulations.jl/issues/1344

src/core/parameters.jl
GabrielKS: This is to support the case where the `param_array` is 3D and the `multiplier_array` is 2D so we have to broadcast. ~If we go back to the design where multiplier arrays are always the same size as parameter arrays, this shouldn't be necessary.~ We have gone back to the 3D multiplier design but this is still necessary for multistart, where the `param_array` is technically 2D but its eltype is tuple while the `multiplier_array`'s eltype is scalar — unless we want tuple multipliers in this case?

src/devices_models/devices/common/objective_function/common.jl
GabrielKS: My understanding is that on this first pass we aren't too concerned about `ProductionCostExpression`?

src/devices_models/devices/common/objective_function/market_bid.jl
GabrielKS: Done

src/devices_models/devices/common/objective_function/market_bid.jl
GabrielKS: Done

src/devices_models/devices/common/objective_function/market_bid.jl
GabrielKS: Done

src/devices_models/devices/common/objective_function/market_bid.jl
GabrielKS: Done using `has_container_key`

src/parameters/add_parameters.jl
```diff
@@ -225,27 +213,52 @@ function _add_time_series_parameters!(
         error("add_parameters! for TimeSeriesParameter is not compatible with $ts_type")
     end
     time_steps = get_time_steps(container)
-    ts_name = get_time_series_names(model)[T]
+
+    ts_names = String[]
+    device_names = String[]
+    active_devices = D[]
+    for device::D in devices
```
jd-lara, line 207: I have never seen this syntax 

```julia
    for device::D in devices
```

jd-lara:
working with @daniel-thom on this. We think there is an issue with the axis definitions in these functions for `IncrementalPiecewiseLinearBreakpointParameter`. This function 

```Julia
function _additional_axes(
    ::OptimizationContainer,
    ::IncrementalPiecewiseLinearBreakpointParameter,
    devices::U,
    ::DeviceModel{D, W},
) where {
    U <: Union{Vector{D}, IS.FlattenIteratorWrapper{D}},
} where {D <: PSY.Component}
    curves = PSY.get_incremental_offer_curves.(PSY.get_operation_cost.(devices))
    max_tranches = maximum(get_max_tranches.(devices, curves))
    return (1:(max_tranches + 1),)  # one more breakpoint than tranches
end
```

seems to generate one additional breakpoint index entry than the code that initializes the state parameter sizes for the simulation in `simulation_state.jl:95`.

src/parameters/add_parameters.jl
GabrielKS: It's essentially just asserting that we expect each device to be a `D` and will throw an error if that's not the case

test/test_market_bid_cost.jl
```diff
@@ -824,7 +824,7 @@ end
         )
         @test all(isapprox.(decisions1_2, decisions2_2))
         # Make sure our tests included all types of startups and shutdowns
-        @assert all(approx_geq_1.(decisions1 .+ decisions1_2))
+        @test all(approx_geq_1.(decisions1 .+ decisions1_2))
```
GabrielKS, line 827: This was intentionally an `@assert` not an `@test` — if this fails, it doesn't mean there's a bug in the implementation, it means the test code itself is bad

jd-lara:
@GabrielKS the code here is causing some dimensionality mismatch. I am not sure how to change the code to avoid this problem with the `expand_ixs` code 

```Julia 
function expand_ixs(ixs::Tuple, dest::AbstractArray)
    if length(ixs) <= ndims(dest)
        return (ixs[1:end-1]..., fill(:, ndims(dest) - length(ixs))..., ixs[end])
    else
        throw(ArgumentError("`ixs` must not index more dimensions than `dest` has"))
    end
end

function assign_expand(dest::AbstractArray, src, ixs::Tuple)
    dest[expand_ixs(ixs, dest)...] .= src
    return
end
```

GabrielKS:
> @GabrielKS the code here is causing some dimensionality mismatch.
Yes, this is one of the pieces that kind of assumes that you want to slice into the first few dimensions and then assign across all the rest — perfect for the old way of assigning all the tranches at once, but will need major revision if we're putting the extra dimension in the middle

PR #1329, opened 2025-07-01
Mb/nongen events

jd-lara:
@m-bossart I am going to open this PR to review it because without this the simulation loops are broken.

### DIFF
--- a/src/contingency_model/contingency_arguments.jl
+++ b/src/contingency_model/contingency_arguments.jl
@@ -11,7 +11,8 @@ function add_event_arguments!(
         event_type = get_entry_type(key)
         devices_with_attributes =
             [d for d in devices if PSY.has_supplemental_attributes(d, event_type)]
-        @assert !isempty(devices_with_attributes)
+        isempty(devices_with_attributes) &&
+            error("no devices found with a supplemental attribute for event $event_type")
         for p_type in [AvailableStatusChangeCountdownParameter, AvailableStatusParameter]
             add_parameters!(
                 container,
@@ -39,7 +40,8 @@ function add_event_arguments!(
         event_type = get_entry_type(key)
         devices_with_attributes =
             [d for d in devices if PSY.has_supplemental_attributes(d, event_type)]
-        @assert !isempty(devices_with_attributes)
+        isempty(devices_with_attributes) &&
+            error("no devices found with a supplemental attribute for event $event_type")
         for p_type in [AvailableStatusChangeCountdownParameter, AvailableStatusParameter]
             add_parameters!(
                 container,
@@ -82,7 +84,8 @@ function add_event_arguments!(
         event_type = get_entry_type(key)
         devices_with_attributes =
             [d for d in devices if PSY.has_supplemental_attributes(d, event_type)]
-        @assert !isempty(devices_with_attributes)
+        isempty(devices_with_attributes) &&
+            error("no devices found with a supplemental attribute for event $event_type")
         for p_type in [AvailableStatusChangeCountdownParameter, AvailableStatusParameter]
             add_parameters!(
                 container,
@@ -139,7 +142,8 @@ function add_event_arguments!(
         event_type = get_entry_type(key)
         devices_with_attributes =
             [d for d in devices if PSY.has_supplemental_attributes(d, event_type)]
-        @assert !isempty(devices_with_attributes)
+        isempty(devices_with_attributes) &&
+            error("no devices found with a supplemental attribute for event $event_type")
         for p_type in [AvailableStatusChangeCountdownParameter, AvailableStatusParameter]
             add_parameters!(
                 container,
@@ -180,7 +184,8 @@ function add_event_arguments!(
         event_type = get_entry_type(key)
         devices_with_attributes =
             [d for d in devices if PSY.has_supplemental_attributes(d, event_type)]
-        @assert !isempty(devices_with_attributes)
+        isempty(devices_with_attributes) &&
+            error("no devices found with a supplemental attribute for event $event_type")
         for p_type in [AvailableStatusChangeCountdownParameter, AvailableStatusParameter]
             add_parameters!(
                 container,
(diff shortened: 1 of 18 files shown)


m-bossart:
I've added outstanding tasks as separate issues so this can be merged after it is reviewed

src/contingency_model/contingency_arguments.jl
```diff
@@ -28,18 +28,204 @@ function add_event_arguments!(
...
+    devices::T,
+    device_model::DeviceModel{U, V},
+    network_model::NetworkModel{<:PM.AbstractActivePowerModel},
+) where {
+    T <: Union{Vector{U}, IS.FlattenIteratorWrapper{U}},
+    V <: Union{StaticPowerLoad, PowerLoadDispatch, PowerLoadInterruption},
+} where {U <: PSY.PowerLoad}
+    for (key, event_model) in get_events(device_model)
+        event_type = get_entry_type(key)
+        devices_with_attrbts =
```
jd-lara: @daniel-thom do we have a faster way to do this mapping search? If not, no problem just to confirm

src/contingency_model/contingency_arguments.jl
```diff
@@ -28,18 +28,204 @@ function add_event_arguments!(
...
+            [d for d in devices if PSY.has_supplemental_attributes(d, event_type)]
+        @assert !isempty(devices_with_attrbts)
+        parameter_type = get_parameter_type(event_type, event_model, U)
+        for p_type in [AvailableStatusChangeCountdownParameter, parameter_type]
```
jd-lara: Open issue to clean up this loop

m-bossart: I removed `get_parameter_type` because it is unnecessary (always returns `AvailablaStatusParameter`) which cleans up this loop

src/contingency_model/contingency_arguments.jl
daniel-thom: We do something similar to this when checking for time series, so it's likely not an issue.

Each call to `has_supplemental_attributes` results in one SQLite call (~4 us). This [PR](https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/471) adds the method `get_component_supplemental_attribute_pairs` to get everything in one SQL query. It would return a vector of named tuples (component, attribute) after performing lots of dictionary lookups. You would then have to filter that vector to match the devices input to this function. I would want to test the performance.

If we wanted to optimize this code path we could make a new function that filters the input component iterable to those associated with the attribute type. It would not be difficult. We risk API overload...people can't find what they want because of too many options.

src/contingency_model/contingency_arguments.jl
```diff
@@ -28,18 +28,204 @@ function add_event_arguments!(
...
+    devices::T,
+    device_model::DeviceModel{U, V},
+    network_model::NetworkModel{<:PM.AbstractPowerModel},
+) where {
+    T <: Union{Vector{U}, IS.FlattenIteratorWrapper{U}},
+    V <: Union{StaticPowerLoad, PowerLoadDispatch, PowerLoadInterruption},
+} where {U <: PSY.PowerLoad}
+    for (key, event_model) in get_events(device_model)
+        event_type = get_entry_type(key)
+        devices_with_attrbts =
```
daniel-thom: Nit: avoid non-standard abbreviations that don't really buy you anything. `devices_with_attributes` is slightly longer but more readable.

src/contingency_model/contingency_arguments.jl
```diff
@@ -28,18 +28,204 @@ function add_event_arguments!(
...
+            [d for d in devices if PSY.has_supplemental_attributes(d, event_type)]
+        @assert !isempty(devices_with_attrbts)
```
daniel-thom: If this could happen because of bad input data, change this to an exception (because asserts could be disabled in the future). Disregard this comment if this is only guarding against programmer error.

src/contingency_model/contingency_arguments.jl
daniel-thom: Another idea: the new function I linked does not support a filter function. We could add that. In this example you could pass a lambda function that checks against a Set of component UUIDs.

src/simulation/simulation_state.jl
```diff
@@ -312,6 +312,44 @@ function update_decision_state!(
...
+    model_resolution = get_resolution(model_params)
+    state_resolution = get_data_resolution(state_data)
+    resolution_ratio = model_resolution ÷ state_resolution
+    @assert_op resolution_ratio >= 1
+    state_timestamps = state_data.timestamps
+    state_data_index = find_timestamp_index(state_timestamps, simulation_time)
+    for name in column_names
+        if event_occurrence_data.values[name, state_data_index] == 1.0
+            outage_index = state_data_index + 1     #outage occurs at the following timestep
+            while true
```
daniel-thom: Seems like you could make this a for loop. It would be more obvious that it is a bounded loop.

src/contingency_model/contingency_arguments.jl
daniel-thom: @jd-lara @m-bossart What would you like here?
1. Keep the [current interface](https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/471) as it is.
2. Extend the interface to support a filter function.
3. Extend the interface to support a Set of component UUIDs to be limited to.
4. Other?

src/contingency_model/contingency_arguments.jl
m-bossart: Data issues (i.e. an event with no devices) are caught with an exception when the events are attached to the simulation.

src/contingency_model/contingency_arguments.jl
m-bossart: I'll defer to Jose on this one as he is more familiar with the various use cases.

src/contingency_model/contingency_arguments.jl
daniel-thom: Ok, then I definitely recommend calling `error()` or some other exception. Here's the official statement:
```
Warning
An assert might be disabled at some optimization levels. Assert should therefore only be used as a debugging tool and not used for authentication verification (e.g., verifying passwords or checking array bounds). The code must not rely on the side effects of running cond for the correct behavior of a function.
```
https://docs.julialang.org/en/v1/base/base/#Base.@assert

src/contingency_model/contingency_arguments.jl
jd-lara: @m-bossart lets open an issue to look into the performance of this line of code later when we scale it to decide.

PR #1340, opened 2025-07-22
Jd/area balance fixes

jd-lara:
The tests need QA/QC for the new Interface for Interchanges. It requires https://github.com/NREL-Sienna/PowerSystems.jl/pull/1485

### DIFF
--- a/src/services_models/services_constructor.jl
+++ b/src/services_models/services_constructor.jl
@@ -563,7 +563,7 @@ function construct_service!(
     network_model::NetworkModel{AreaBalancePowerModel},
 )
     interfaces = get_available_components(model, sys)
-    interface = PSY.get_component(T, sys, get_service_name(model))
+    interface = PSY.get_component(PSY.TransmissionInterface, sys, get_service_name(model))
     if get_use_slacks(model)
         # Adding the slacks can be done in a cleaner fashion
         @assert PSY.get_available(interface)
@@ -573,7 +573,7 @@ function construct_service!(
     lazy_container_addition!(
         container,
         InterfaceTotalFlow(),
-        T,
+        PSY.TransmissionInterface,
         PSY.get_name.(interfaces),
         get_time_steps(container),
     )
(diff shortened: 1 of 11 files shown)


src/services_models/services_constructor.jl
```diff
@@ -553,52 +553,46 @@ function construct_service!(
     return
 end
```
m-bossart, line 554: Dispatch on `PSY.TransmissionInterface` and not `T` in line 531?

src/services_models/services_constructor.jl
```diff
@@ -676,7 +671,7 @@ function construct_service!(
             add_parameters!(container, MaxInterfaceFlowLimitParameter, device, model)
         end
     end
-    interface = PSY.get_component(T, sys, get_service_name(model))
+    interface = PSY.get_component(PSY.TransmissionInterface, sys, get_service_name(model))
     add_feedforward_arguments!(container, model, interface)
     return
 end
```
m-bossart, line 677: Dispatch on `PSY.TransmissionInterface` and not `T` in line 688?

test/test_services_constructor.jl
```diff
@@ -438,7 +438,7 @@ end
           PSI.ModelBuildStatus.BUILT
     moi_tests(model, 312, 0, 288, 288, 168, false)
 
-    #= TODO: Fix this test
+    #= TODO: Fix this tes6
```
m-bossart, line 441: test

test/test_services_constructor.jl
```diff
@@ -586,3 +586,164 @@ end
     )
     @test_throws ArgumentError build!(sim; console_level = Logging.AboveMaxLevel)
 end
+
+@testset "2 Areas AreaBalance With Transmission Interface" begin
```
m-bossart, line 590: This test name seems misleading, because it only models the interchanges not the interfaces?

test/test_services_constructor.jl
```diff
@@ -586,3 +586,164 @@ end
...
+    @test size(copper_plate_constraints) == (3, 24)
+
+    interchange_constraints_ub =
+        PSI.get_constraint(opt_container, InterfaceFlowLimit(), TransmissionInterface, "ub")
+    interchange_constraints_lb =
+        PSI.get_constraint(opt_container, InterfaceFlowLimit(), TransmissionInterface, "lb")
+    @test size(interchange_constraints_ub) == (1, 24)
+    @test size(interchange_constraints_lb) == (1, 24)
+
+    interchange_constraints_ub["interface1_2_3", 1]
```
m-bossart, line 740: This line is unused. Delete?

test/test_services_constructor.jl
```diff
@@ -586,3 +586,164 @@ end
...
+    # This interface is limiting all the flows into 1
+    interface = TransmissionInterface(;
+        name = "interface1_2_3",
+        available = true,
+        active_power_flow_limits = (min = 0.0, max = 1.0),
+        violation_penalty = 1000.0,
+        direction_mapping = Dict("interchange1_2" => 1,
+            "interchange1_3" => -1,
+        ),
+    )
```
m-bossart, line 695: interchange1_2 and interchange1_3 are both defined as the flows out of area 1, so wouldn't the direction be -1 for both of them when defining the interface if we want to limit all the flows into 1?

m-bossart (changes requested):

PR #1362, opened 2025-09-02
Support 3-dimensional results

daniel-thom:
@jd-lara @GabrielKS This PR enables progress of 3d results in PSI. It requires this [IS branch](https://github.com/NREL-Sienna/InfrastructureSystems.jl/tree/feat/3d-results).

Lots of tests are failing in the psy5 branch of PSI, so this probably still needs work. The tests in `test/test_market_bid_cost.jl` cover a lot, and they all pass with one exception: comparison of the objective_value fails when the simulation store is in-memory. It passes with the HDF5 store. That must be some bug in existing code. I bypassed that check for that case.

The changes in test_market_bid_cost.jl and test_simulation_results.jl illustrate the amount of changes we are imposing on users due to the change of wide table format to long format. There is a flag to toggle the behavior, but currently, it defaults to long.

I have noted several todos that need completion.


src/utils/jump_utils.jl
```diff
@@ -105,31 +211,228 @@ function encode_tuple_to_column(val::Tuple{String, Int})
...
+  The resulting DataFrame will have the column "DateTime" if timestamps is not nothing.
+  Otherwise, it will have the column "time_index", representing the index of the time
+  dimension.
+- `::Val{TableFormat}`: Format of the table to create.
+  If it is TableFormat.LONG, the DataFrame will have the column "component", and, if
+  the data has three dimensions, "component_x."
+  If it is TableFormat.WIDE, the DataFrame will have columns for each component. Wide
+  format does not support arrays with more than two dimensions.
+"""
+# TODO: Are "component" and "component_x" the right names?
```
daniel-thom: What should we use for this column?

test/test_model_decision.jl
```diff
@@ -91,600 +95,600 @@ end
...
-        c_sys5_bat;
-        optimizer = HiGHS_optimizer,
-    )
-    @test build!(model; output_dir = output_dir) == PSI.ModelBuildStatus.BUILT
-    @test solve!(model) == PSI.RunStatus.SUCCESSFULLY_FINALIZED
-    res = OptimizationProblemResults(model)
-    shortage = read_variable(res, "StorageEnergyShortageVariable__EnergyReservoirStorage")
-    @test nrow(shortage) == 1
-end
+# @testset "Test optimization debugging functions" begin
```
daniel-thom: I'm guessing that these are known failures.

test/test_simulation_results.jl
```diff
@@ -269,48 +269,48 @@ function test_simulation_results(
...
-                    ActivePowerVariable,
-                    ThermalStandard,
-                )],
-            )
-            @test !isempty(sim.internal.store.dm_data[:ED].optimizer_stats)
-            empty!(sim.internal.store)
-            @test isempty(sim.internal.store.dm_data[:ED].variables)
-            @test isempty(sim.internal.store.dm_data[:ED].optimizer_stats)
-        end
+        # if !in_memory
```
daniel-thom: Ditto

src/core/dataset.jl
```diff
@@ -223,7 +223,7 @@ end
 function HDF5Dataset{1}(
     values::HDF5.Dataset,
     column_dataset::HDF5.Dataset,
-    ::Tuple,
+    ::NTuple{1, Int},
```
daniel-thom, line 226: I refactored these constructors to make the behavior more clear. They were previously correct; I just had difficultly following the logic.

src/core/dataset.jl
```diff
@@ -242,43 +242,17 @@ end
...
-        1,
-        0,
-        resolution,
-        initial_time,
-        UNSET_INI_TIME,
-        (col1, col2),
-    )
-end
-
-function HDF5Dataset{2}(
```
daniel-thom, line 268: This version of the function should not be needed.

src/core/optimization_container.jl
```diff
@@ -2087,6 +2087,32 @@ function get_time_series_initial_values!(
...
+
+# Arguments
+- `container::OptimizationContainer`: The optimization container.
+- `field::Symbol`: The field for which to retrieve the column names.
+- `key::OptimizationContainerKey`: The key for which to retrieve the column names.
+
+# Returns
+- `Tuple`: Tuple of Vector{String}.
+"""
+function get_column_names(
```
daniel-thom, line 2100: I had difficultly following when `get_column_names` would run on one of our containers and when it would run on a JuMP container. I split the function names into two: this one and `get_column_names_from_axis_array`. They seem like different things to me, but let me know if you disagree.

src/core/parameters.jl
```diff
@@ -407,8 +407,8 @@ should_write_resulting_value(::Type{<:EventParameter}) = true
 should_write_resulting_value(::Type{<:FuelCostParameter}) = true
 should_write_resulting_value(::Type{<:ShutdownCostParameter}) = true
 should_write_resulting_value(::Type{<:AbstractCostAtMinParameter}) = true
-should_write_resulting_value(::Type{<:AbstractPiecewiseLinearBreakpointParameter}) = false  # because 3D is currently unsupported
-should_write_resulting_value(::Type{<:AbstractPiecewiseLinearSlopeParameter}) = false  # because 3D is currently unsupported
+should_write_resulting_value(::Type{<:AbstractPiecewiseLinearSlopeParameter}) = true
```
daniel-thom, line 410: @GabrielKS Is the TODO on line 407 complete?

src/operation/emulation_model_store.jl
```diff
@@ -97,10 +97,20 @@ function write_result!(
     key::OptimizationContainerKey,
     index::EmulationModelIndexType,
     update_timestamp::Dates.DateTime,
-    array::DenseAxisArray{<:Any, 2},
+    array::DenseAxisArray{Float64, 2},
 )
-    @assert_op size(array)[2] == 1
-    write_result!(store, name, key, index, update_timestamp, array[:, 1])
+    if size(array, 2) == 1
+        write_result!(store, name, key, index, update_timestamp, array[:, 1])
```
daniel-thom, line 103: @jd-lara This function call seems odd. I think the caller should do this slicing instead. Should we change it?

src/operation/operation_model_interface.jl
```diff
@@ -42,8 +42,12 @@ end
 
 get_problem_base_power(model::OperationModel) = PSY.get_base_power(model.sys)
 get_settings(model::OperationModel) = get_optimization_container(model).settings
+
 get_optimizer_stats(model::OperationModel) =
-    get_optimizer_stats(get_optimization_container(model))
+# This deepcopy is important because the optimization container is overwritten
```
daniel-thom, line 47: This caused the InMemorySimulationStore to report the same stats at every time step.

src/operation/operation_model_interface.jl
```diff
@@ -390,27 +393,9 @@ read_aux_variable(model::OperationModel, key::AuxVarKey) = _read_results(model,
 read_variable(model::OperationModel, key::VariableKey) = _read_results(model, key)
 read_expression(model::OperationModel, key::ExpressionKey) = _read_results(model, key)
 
-function _read_col_name(axes)
```
daniel-thom, line 393: Refert to `jump_utils.jl` for the new logic.

src/simulation/hdf_simulation_store.jl
```diff
@@ -1061,10 +1057,20 @@ end
...
     dataset::HDF5.Dataset,
-    array::Array{Float64, 4},
+    array::Matrix{Float64},
     index::EmulationModelIndexType,
 )
     dataset[index, :, :] = array
+    return
+end
+
+# TODO DT: this looked wrong. Was it tested?
```
daniel-thom, line 1086: Do we have a script to test this?

src/simulation/realized_meta.jl
```diff
@@ -52,6 +52,33 @@ function _make_dataframe(
     num_rows::Int,
     meta::RealizedMeta,
     key::OptimizationContainerKey,
+    ::Val{TableFormat.LONG},
+)
+    # TODO: Implement long format directly.
```
daniel-thom, line 57: I'm not planning on implementing this unless we can prove it's warranted for performance. Differences could be negligible.

src/simulation/simulation_state.jl
```diff
@@ -104,10 +104,11 @@ function _initialize_model_states!(
     for field in fieldnames(DatasetContainer)
         field_containers = getfield(container, field)
         field_states = getfield(states, field)
-        for (key, value) in field_containers
+        for (key, field_container) in field_containers
             !should_write_resulting_value(key) && continue
             value_counts = params[key].horizon ÷ params[key].resolution
-            column_names = get_column_names(key, value)
+            column_names = get_column_names(container, field, field_container, key)
+            # TODO DT: why would we overwrite a key? Is this a bug?
```
daniel-thom, line 111: Open question.

src/utils/jump_utils.jl
```diff
@@ -44,57 +45,162 @@ function fix_parameter_value(input::JuMP.VariableRef, value::Float64)
     return
 end
 
-function to_matrix(array::DenseAxisArray{T, 1, K}) where {T, K <: NTuple{1, Any}}
+"""
+Convert Vectors, DenseAxisArrays, and SparkAxisArrays to a matrix.
+
+- If the input is a 1d array or DenseAxisArray, the returned matrix will have
```
daniel-thom, line 52: I documented our current logic, but find it confusing. A 5-element vector could be reshaped to a 1x5 or 5x1 matrix. Perhaps we should remove this method and let the caller specify exactly what they want.

src/utils/jump_utils.jl
```diff
@@ -44,57 +45,162 @@ function fix_parameter_value(input::JuMP.VariableRef, value::Float64)
...
 end
 
-function get_column_names(
-    key::OptimizationContainerKey,
-    ::DenseAxisArray{T, 1, K},
-) where {T, K <: NTuple{1, Any}}
-    return get_column_names(key)
+function to_matrix(array::SparseAxisArray{T, N, K}) where {T, N, K <: NTuple{N, Any}}
+    # Don't use get_column_names_from_axis_array to avoid additional string conversion
+    # TODO DT: I don't understand why we have two mechanisms of creating columns.
```
daniel-thom, line 91: Should this be changed?

src/utils/jump_utils.jl
```diff
@@ -44,57 +45,162 @@ function fix_parameter_value(input::JuMP.VariableRef, value::Float64)
...
+There are two variants of this function:
+  - get_column_names_from_axis_array(array)
+  - get_column_names_from_axis_array(::OptimizationContainerKey, array)
+  
+When the variant with the key is called:
+  In cases where the array has one dimension, retrieve the column names from the key.
+  In cases where the array has two or more dimensions, retrieve the column names from the
+  axes.
+"""
+# TODO DT: the docstring describes what the code does.
```
daniel-thom, line 112: After going through all this, I think that we should change it so that we don't need the key. We can always store the column names in the axes. I tried to maintain current behavior in order not to break things.

src/utils/jump_utils.jl
```diff
@@ -44,57 +45,162 @@ function fix_parameter_value(input::JuMP.VariableRef, value::Float64)
...
 end
 
-function get_column_names(
+function get_column_names_from_axis_array(
+    array::DenseAxisArray{T, 2, <:Tuple{Vector{Int}, UnitRange{Int}}},
+) where {T}
+    return (string.(axes(array, 1)),)
+end
+
+# TODO DT: do we need this one?
```
daniel-thom: All these variants imply to me that we can simplify.

src/utils/jump_utils.jl
```diff
@@ -44,57 +45,162 @@ function fix_parameter_value(input::JuMP.VariableRef, value::Float64)
...
+) where {T}
+    return (string.(axes(array, 1)),)
+end
+
+function get_column_names_from_axis_array(
+    array::DenseAxisArray{T, 2, <:Tuple{Vector{String}, Vector{String}}},
+) where {T}
+    # Currently, variables that don't have timestamps have a dummy axes to keep
+    # two axes in the Store (HDF or Memory).
+    # TODO DT: confirm this with Jose.
```
daniel-thom: Open question.

src/utils/jump_utils.jl
GabrielKS: See discussion at https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/482#discussion_r2316281832

src/core/parameters.jl
GabrielKS: I think we can just have the general case
```julia
should_write_resulting_value(::Type{<:ObjectiveFunctionParameter}) = true
```
now. If not, the remaining special case to complete the TODO would be
```julia
should_write_resulting_value(::Type{<:StartupCostParameter}) = true
```

src/operation/emulation_model_store.jl
daniel-thom: Update: try to change it.

src/simulation/hdf_simulation_store.jl
daniel-thom: Will be tested by contingencies.

src/simulation/simulation_state.jl
daniel-thom: Existing code is correct

src/utils/jump_utils.jl
daniel-thom: Defer

src/utils/jump_utils.jl
daniel-thom: Defer. Talk to Gabriel because this method might not be needed any longer.

src/operation/decision_model_store.jl
```diff
@@ -122,9 +140,11 @@ function write_optimizer_stats!(
     stats::OptimizerStats,
     index::DecisionModelIndexType,
 )
-    if index in keys(store.optimizer_stats)
-        @warn "Overwriting optimizer stats"
-    end
+    # TODO: This check is incompatible with test calls to psi_checksolve_test
```
daniel-thom, line 136: It would be nice to keep this check. Would need to workaround the test behavior.

src/simulation/hdf_simulation_store.jl
```diff
@@ -1061,10 +1057,22 @@ end
 
 function _write_dataset!(
     dataset::HDF5.Dataset,
-    array::Array{Float64, 4},
+    array::Matrix{Float64},
     index::EmulationModelIndexType,
 )
+    # TODO: something is broken here. In at least one case, dataset is 2d, not 3d.
```
daniel-thom: I'm not yet sure what is wrong here.

src/utils/jump_utils.jl
```diff
@@ -105,31 +210,256 @@ function encode_tuple_to_column(val::Tuple{String, Int})
...
+    end
+
+    return DataFrame(
+        :time_index => time_col,
+        :name => name_col,
+        :value => reshape(permutedims(array.data), num_rows),
+    )
+end
+
+# TODO: do we really need this? Can we change the storage to not add the dummy ["1"] value?
```
daniel-thom: Open question

PR #1368, opened 2025-09-08
Various outages fixes 

m-bossart:
- Fixes the moi_tests for cases where `built_for_recurrent_solves = true`. 
When adding an event to a generator with `built_for_recurrent_solves = true`, the number of `JuMP.VariableRef`s in the problem increases by `2 * n_timesteps` due to the addition of the  `AvailableStatusParameter` and `AvailableStatusChangeCountdownParameter` parameters.
- Adds a quadratic outage constraint on reactive power. The quadratic constraint is necessary to ensure the reactive power goes to zero during the outage condition. 
- Adds a test case for reactive power + thermal outage.
- Fixes the moi tests for the case when adding event models.

### DIFF
--- a/test/test_device_thermal_generation_constructors.jl
+++ b/test/test_device_thermal_generation_constructors.jl
@@ -144,10 +144,7 @@ end
     psi_aux_variable_test(model, aux_variables_keys)
     model = DecisionModel(MockOperationProblem, ACPPowerModel, c_sys5_uc)
     mock_construct_device!(model, device_model; add_event_model = true)
-    moi_tests(model, 600, 0, 624, 240, 120, true)
-    # Outage constraint for reactive power is quadratic: 
-    @test JuMP.num_constraints(PSI.get_jump_model(model), GQEVF, MOI.LessThan{Float64}) ==
-          24
+    moi_tests(model, 600, 0, 624, 240, 120, true, 24)
 
     device_model = DeviceModel(ThermalStandard, ThermalStandardUnitCommitment)
 
@@ -159,10 +156,7 @@ end
     psi_checkobjfun_test(model, GQEVF)
     model = DecisionModel(MockOperationProblem, ACPPowerModel, c_sys14;)
     mock_construct_device!(model, device_model; add_event_model = true)
-    moi_tests(model, 600, 0, 384, 240, 120, true)
-    # Outage constraint for reactive power is quadratic: 
-    @test JuMP.num_constraints(PSI.get_jump_model(model), GQEVF, MOI.LessThan{Float64}) ==
-          24
+    moi_tests(model, 600, 0, 384, 240, 120, true, 24)
 end
 
 @testset "Thermal MultiStart UC With DC - PF" begin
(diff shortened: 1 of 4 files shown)


src/contingency_model/contingency_constraints.jl
```diff
@@ -208,3 +208,75 @@ function add_event_constraints!(
...
+    ::Type{T},
+    ::Type{R},
+    ::Type{P},
+    devices::Union{Vector{V}, IS.FlattenIteratorWrapper{V}},
+    model::DeviceModel{V, W},
+    ::Type{X},
+) where {
+    T <: ConstraintType,
+    R <: VariableType,
+    P <: ParameterType,
```
jd-lara: Can't we make this more concrete? We know exactly which parameters, variables and constraints they are.

src/contingency_model/contingency_constraints.jl
```diff
@@ -208,3 +208,75 @@ function add_event_constraints!(
...
+
+function _add_reactive_power_contingency_constraint_impl!(
+    container::OptimizationContainer,
+    ::Type{T},
+    array_reactive,
+    param::P,
+    devices::Union{Vector{V}, IS.FlattenIteratorWrapper{V}},
+    model::DeviceModel{V, W},
+) where {
+    T <: ConstraintType,
```
jd-lara: Same here

test/test_device_thermal_generation_constructors.jl
```diff
@@ -709,7 +758,10 @@ end
     psi_checkobjfun_test(model, GAEVF)
     model = DecisionModel(MockOperationProblem, ACPPowerModel, c_sys5)
     mock_construct_device!(model, device_model; add_event_model = true)
-    moi_tests(model, 600, 0, 408, 240, 120, true)
+    moi_tests(model, 600, 0, 384, 240, 120, true)
+    # Outage constraint for reactive power is quadratic: 
+    @test JuMP.num_constraints(PSI.get_jump_model(model), GQEVF, MOI.LessThan{Float64}) ==
```
jd-lara: can you put this in the MOI tests? we can me the quadratic count to be `Union{Int, nothing}` and default it to nothing so we don't need to redo all the tests.

test/test_device_thermal_generation_constructors.jl
```diff
@@ -721,7 +773,10 @@ end
     psi_checkobjfun_test(model, GAEVF)
     model = DecisionModel(MockOperationProblem, ACPPowerModel, c_sys5_pglib;)
     mock_construct_device!(model, device_model; add_event_model = true)
-    moi_tests(model, 432, 0, 192, 96, 144, true)
+    moi_tests(model, 432, 0, 168, 96, 144, true)
+    # Outage constraint for reactive power is quadratic: 
+    @test JuMP.num_constraints(PSI.get_jump_model(model), GQEVF, MOI.LessThan{Float64}) ==
```
jd-lara: ditto here

PR #1372, opened 2025-09-10
fix power flow in the loop

luke-kiernan:
Most of the tests in `test_power_flow_in_the_loop.jl` now succeed, with 2 exceptions:
- "HVDC with AC PF in the loop": the `RTS_GMLC_DA_sys` system doesn't build correctly for me.
- "Test DC power flow in the loop setup: RTS ED, PTDF, no export": `PowerFlows.jl` objects to the lack of sources at a PV bus, even though it's a DC power flow so non-slack bus types don't matter.

This relies on PR [#239](https://github.com/NREL-Sienna/PowerFlows.jl/pull/239) in `PowerFlows.jl` for one function, `get_network_reduction_data`.

### DIFF
--- a/src/network_models/power_flow_evaluation.jl
+++ b/src/network_models/power_flow_evaluation.jl
@@ -305,14 +305,9 @@ bus_aux_vars(::PFS.PTDFPowerFlowData) = DataType[]
 bus_aux_vars(::PFS.vPTDFPowerFlowData) = DataType[]
 bus_aux_vars(::PFS.PSSEExporter) = DataType[]
 
-function _get_branch_component_tuples(pfd::PFS.PowerFlowData)
-    branch_types = PFS.get_branch_type(pfd)
-    return [(branch_types[val], key) for (key, val) in pairs(PFS.get_branch_lookup(pfd))]
-end
-
-_get_branch_component_tuples(pfd::PFS.SystemPowerFlowContainer) = [
+_get_branch_component_tuples(sys::PSY.System) = [
     (typeof(c), get_name(c)) for
-    c in PSY.get_available_components(PSY.Branch, PFS.get_system(pfd))
+    c in PSY.get_available_components(PSY.Branch, sys)
 ]
 
 _get_bus_component_tuples(pfd::PFS.PowerFlowData) =
@@ -343,7 +338,7 @@ function add_power_flow_data!(
         my_branch_aux_vars = branch_aux_vars(pf_data)
         my_bus_aux_vars = bus_aux_vars(pf_data)
 
-        my_branch_components = _get_branch_component_tuples(pf_data)
+        my_branch_components = _get_branch_component_tuples(sys)
         for branch_aux_var in my_branch_aux_vars
             to_add_to = get!(
                 branch_aux_var_components,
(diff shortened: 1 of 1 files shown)


src/core/auxiliary_variables.jl
```diff
@@ -28,25 +28,30 @@ Auxiliary Variable for the bus voltage magnitued results from power flow evaluat
 """
 struct PowerFlowVoltageMagnitude <: PowerFlowAuxVariableType end
 
+"""
```
jd-lara, line 31: PowerFlowLineFlowType is kinda strange semantics....

src/network_models/power_flow_evaluation.jl
```diff
@@ -305,14 +305,9 @@ bus_aux_vars(::PFS.PTDFPowerFlowData) = DataType[]
...
-function _get_branch_component_tuples(pfd::PFS.PowerFlowData)
-    branch_types = PFS.get_branch_type(pfd)
-    return [(branch_types[val], key) for (key, val) in pairs(PFS.get_branch_lookup(pfd))]
-end
-
-_get_branch_component_tuples(pfd::PFS.SystemPowerFlowContainer) = [
+_get_branch_component_tuples(sys::PSY.System) = [
     (typeof(c), get_name(c)) for
-    c in PSY.get_available_components(PSY.Branch, PFS.get_system(pfd))
+    c in PSY.get_available_components(PSY.Branch, sys)
```
jd-lara, line 310: Are we sure Branch is the correct super type here?

src/network_models/power_flow_evaluation.jl
```diff
@@ -594,39 +589,75 @@ _get_pf_result(::Type{PowerFlowVoltageAngle}, pf_data::PFS.PowerFlowData) =
...
 _get_pf_result(::Type{PowerFlowLossFactors}, pf_data::PFS.PowerFlowData) =
     PFS.get_loss_factors(pf_data)
 _get_pf_result(::Type{PowerFlowVoltageStabilityFactors}, pf_data::PFS.PowerFlowData) =
     PFS.get_voltage_stability_factors(pf_data)
 
-_get_pf_lookup(::Type{<:PSY.Bus}, pf_data::PFS.PowerFlowData) = PFS.get_bus_lookup(pf_data)
-_get_pf_lookup(::Type{<:PSY.Branch}, pf_data::PFS.PowerFlowData) =
-    PFS.get_branch_lookup(pf_data)
+function calculate_aux_variable_value!(container::OptimizationContainer,
+    key::AuxVarKey{T, <:PSY.ACBus}, # TODO: does this work ok with DCBuses, too?
```
jd-lara, line 605: DC Buses don't apply here. We don't have a multi-terminal HVDC interface or methods for this yet

src/network_models/power_flow_evaluation.jl
```diff
@@ -594,39 +589,75 @@ _get_pf_result(::Type{PowerFlowVoltageAngle}, pf_data::PFS.PowerFlowData) =
...
     @debug "Updating $key from PowerFlowData"
     pf_data = get_power_flow_data(pf_e_data)
     src = _get_pf_result(T, pf_data)
-    lookup = _get_pf_lookup(U, pf_data)
     dest = get_aux_variable(container, key)
-    for component_id in axes(dest, 1)  # these are bus numbers or branch names
-        dest[component_id, :] = src[lookup[component_id], :]
+    nrd = PFS.get_network_reduction_data(pf_data)
+    arc_lookup = PFS.get_arc_lookup(pf_data)
+    # PERF: could pre-compute a Dict of branch type to arcs, then intersect the arcs
```
jd-lara, line 631: open an issue about this

src/network_models/power_flow_evaluation.jl
```diff
@@ -594,39 +589,75 @@ _get_pf_result(::Type{PowerFlowVoltageAngle}, pf_data::PFS.PowerFlowData) =
...
+                dest[name, :] = 1 / n_branches .* src[arc_ix, :]
+            end
+        end
     end
     return
 end
 
 function calculate_aux_variable_value!(container::OptimizationContainer,
-    key::AuxVarKey{T, <:Any} where {T <: PowerFlowAuxVariableType},
+    key::AuxVarKey{<:PowerFlowAuxVariableType, <:Any},
```
jd-lara, line 660: why Any? Should this be a ACTransmission type?

luke-kiernan: No: in this function, `key` could be a bus-associated variable, and have a type like `AuxVarKey{PowerFlowVoltageAngle, ACBus}`. The `calculate_aux_variable_value!` at the end of this function is then a multiple dispatch on bus vs branch.

src/network_models/power_flow_evaluation.jl
luke-kiernan: I want 2-terminal ACTransmission objects: looking at PSY, I believe that's branch.

src/core/power_flow_data_wrapper.jl
```diff
@@ -12,7 +12,22 @@ mutable struct PowerFlowEvaluationData{T <: PFS.PowerFlowContainer}
...
 end
 
+check_network_reduction(::PFS.SystemPowerFlowContainer) = nothing
+
+function check_network_reduction(pfd::PFS.PowerFlowData)
+    nrd = PFS.get_network_reduction_data(pfd)
+    if length(PNM.get_reverse_bus_search_map(nrd)) != 0
+        throw(
+            IS.NotImplementedError(
+                "Network reduction is not supported in PowerSimulations.jl.",
```
GabrielKS, line 22: nitpick: what about the network reduction isn't supported?

src/network_models/power_flow_evaluation.jl
```diff
@@ -594,39 +589,75 @@ _get_pf_result(::Type{PowerFlowVoltageAngle}, pf_data::PFS.PowerFlowData) =
...
-    lookup = _get_pf_lookup(U, pf_data)
     dest = get_aux_variable(container, key)
-    for component_id in axes(dest, 1)  # these are bus numbers or branch names
-        dest[component_id, :] = src[lookup[component_id], :]
+    nrd = PFS.get_network_reduction_data(pf_data)
+    arc_lookup = PFS.get_arc_lookup(pf_data)
+    # PERF: could pre-compute a Dict of branch type to arcs, then intersect the arcs
+    # for the type U with the keys of the branch maps.
+    for (arc, br) in nrd.direct_branch_map
+        if br isa U # always a concrete class, so same as: typeof(br) == U
```
GabrielKS, line 634: concrete *type

src/network_models/power_flow_evaluation.jl
```diff
@@ -594,39 +589,75 @@ _get_pf_result(::Type{PowerFlowVoltageAngle}, pf_data::PFS.PowerFlowData) =
...
     src = _get_pf_result(T, pf_data)
-    lookup = _get_pf_lookup(U, pf_data)
     dest = get_aux_variable(container, key)
-    for component_id in axes(dest, 1)  # these are bus numbers or branch names
-        dest[component_id, :] = src[lookup[component_id], :]
+    nrd = PFS.get_network_reduction_data(pf_data)
+    arc_lookup = PFS.get_arc_lookup(pf_data)
+    # PERF: could pre-compute a Dict of branch type to arcs, then intersect the arcs
+    # for the type U with the keys of the branch maps.
+    for (arc, br) in nrd.direct_branch_map
```
GabrielKS, line 633: nitpick, is that stable interface? same below

src/network_models/power_flow_evaluation.jl
```diff
@@ -594,39 +589,75 @@ _get_pf_result(::Type{PowerFlowVoltageAngle}, pf_data::PFS.PowerFlowData) =
...
+    end
+    for (arc, parallel_brs) in nrd.parallel_branch_map # parallel_brs is Set{ACTransmission}
+        sample_line = first(parallel_brs)
+        impedance = PSY.get_r(sample_line) + im * PSY.get_x(sample_line)
+        n_branches = length(parallel_brs)
+        for br in parallel_brs
+            if br isa U
+                @assert T <: PowerFlowLineFlowType "Only PowerFlowLineFlowType aux vars " *
+                                                   "can be used for parallel branches: got $T"
+                @assert PSY.get_r(br) + im * PSY.get_x(br) == impedance "All parallel " *
```
GabrielKS, line 648: is this properly an `@assert` or an error? I think it's only an `@assert` if we aim to make it impossible to end up in this situation using public interface

src/network_models/power_flow_evaluation.jl
GabrielKS: (this is an existing issue, but I could have done `<:PSY.Component`, right?)

test/test_power_flow_in_the_loop.jl
```diff
@@ -54,6 +54,108 @@
...
+                im * vd["PowerFlowLineReactivePowerFromTo__Line"][:, name][1]
+        else
+            name = get_name(line)
+            original_line_flow =
+                vd["PowerFlowLineActivePowerFromTo__Line"][:, name][1] +
+                im * vd["PowerFlowLineReactivePowerFromTo__Line"][:, name][1]
+        end
+    end
+
+    @test isapprox(
```
GabrielKS, line 116: nice I like it

test/test_power_flow_in_the_loop.jl
```diff
@@ -54,6 +54,108 @@
...
+        NetworkModel(
+            PTDFPowerModel;
+            PTDF_matrix = PTDF(system),
+            power_flow_evaluation = ACPowerFlow(),
+        ),
+    )
+    model_m = DecisionModel(template, system; optimizer = HiGHS_optimizer)
+    # ideally, would @test_logs for IS.NotImplementedError, but build! catches all errors
+    # and does `@error "DecisionModel Build Failed"`, attaching the original error as
+    # a field named "exception." This means @test_logs can't see the original error message.
```
GabrielKS, line 152: You can use `PSI.build_impl!` to achieve `@test_logs`, see https://github.com/NREL-Sienna/PowerSimulations.jl/blob/9ec20bc98bac0005637686d07217c869b2b2f902/test/test_market_bid_cost.jl#L934-L939

src/network_models/power_flow_evaluation.jl
luke-kiernan: The first one is definitely an `@assert`: it's a safeguard against unintended consequences of future codebase changes. Second one I'll change to an error.

PR #1373, opened 2025-09-12
Sm/fix_ambiguities

SebastianManriqueM:
This pull request fixes the ambiguities detected in the tests for HVDC and thermal models.

It was solved using a helper function.

All uncomented test passed locally.

### DIFF
--- a/src/devices_models/devices/HVDCsystems.jl
+++ b/src/devices_models/devices/HVDCsystems.jl
@@ -136,7 +136,7 @@ function add_to_expression!(
     ::Type{T},
     ::Type{U},
     devices::IS.FlattenIteratorWrapper{V},
-    ::DeviceModel{V, W},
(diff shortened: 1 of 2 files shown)


src/devices_models/devices/HVDCsystems.jl
```diff
@@ -130,6 +130,32 @@ function add_to_expression!(
...
+    V <: PSY.InterconnectingConverter,
+    W <: AbstractConverterFormulation,
+    X <: AreaPTDFPowerModel,
+}
+    _add_to_expression!(
+        container,
+        T,
+        U,
+        devices,
+        DeviceModel{V, W}(),
```
jd-lara: pass the model upstream, there is no need to make a new instance

SebastianManriqueM: Fixed. Thanks @jd-lara

PR #1378, opened 2025-09-15
address gabriel's comments

luke-kiernan:
Address Gabriel's comments on previous PR.

Also I split the "there's a network reduction" error into 2 cases: if there's just a breaker-switch, log an error but continue; if there's a full-fledged network reduction, throw a `NotImplemented` error as before.

### DIFF
--- a/src/core/power_flow_data_wrapper.jl
+++ b/src/core/power_flow_data_wrapper.jl
@@ -16,12 +16,15 @@ check_network_reduction(::PFS.SystemPowerFlowContainer) = nothing
 
 function check_network_reduction(pfd::PFS.PowerFlowData)
     nrd = PFS.get_network_reduction_data(pfd)
-    if length(PNM.get_reverse_bus_search_map(nrd)) != 0
+    if !isempty(PNM.get_reductions(nrd))
         throw(
             IS.NotImplementedError(
-                "Network reduction is not supported in PowerSimulations.jl.",
+                "Network reductions of types $(PNM.get_reductions(nrd)) are not supported " *
+                "in PowerSimulations.jl.",
             ),
         )
+    elseif length(PNM.get_reverse_bus_search_map(nrd)) != 0
+        @error("Buses are reduced, due to breaker-switches. This will likely cause errors.")
     end
     return
 end
--- a/src/network_models/power_flow_evaluation.jl
+++ b/src/network_models/power_flow_evaluation.jl
@@ -631,14 +631,14 @@ function calculate_aux_variable_value!(container::OptimizationContainer,
     arc_lookup = PFS.get_arc_lookup(pf_data)
     # PERF: could pre-compute a Dict of branch type to arcs, then intersect the arcs
     # for the type U with the keys of the branch maps.
-    for (arc, br) in nrd.direct_branch_map
-        if br isa U # always a concrete class, so same as: typeof(br) == U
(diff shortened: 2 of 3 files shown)


luke-kiernan:
I figured out how to allow for breaker-switches in power flow in-the-loop. Instead of opening a separate PR for that, I've just added it on to this one, as it's only about 50 lines across 3 files.

src/core/power_flow_data_wrapper.jl
```diff
@@ -16,10 +16,11 @@ check_network_reduction(::PFS.SystemPowerFlowContainer) = nothing
 
 function check_network_reduction(pfd::PFS.PowerFlowData)
     nrd = PFS.get_network_reduction_data(pfd)
-    if length(PNM.get_reverse_bus_search_map(nrd)) != 0
+    if !isempty(PNM.get_reductions(nrd))
         throw(
             IS.NotImplementedError(
-                "Network reduction is not supported in PowerSimulations.jl.",
+                "Network reductions of types $(PNM.get_reductions(nrd)) are not supported " *
+                "in PowerSimulations.jl.",
```
m-bossart: This comment is misleading: network reductions are supported in PowerSimulations.jl, just not yet for power flow in the loop.

test/test_power_flow_in_the_loop.jl
```diff
@@ -147,11 +149,12 @@ end
...
-    # ideally, would @test_logs for IS.NotImplementedError, but build! catches all errors
-    # and does `@error "DecisionModel Build Failed"`, attaching the original error as
-    # a field named "exception." This means @test_logs can't see the original error message.
+
     @test build!(model_m; output_dir = mktempdir(; cleanup = true)) ==
-          PSI.ModelBuildStatus.FAILED
+          PSI.ModelBuildStatus.BUILT
+    @test solve!(model_m) == PSI.RunStatus.SUCCESSFULLY_FINALIZED
+    # the interface currently doesn't allow for building a model with network
+    # reductions. we'd have to pass kwargs all the way down to add_power_flow_data!.
```
m-bossart: The network reductions for the simulation and the power flow should be handled independently right? Because we can have one without the other.

m-bossart (changes requested):

test/test_power_flow_in_the_loop.jl
luke-kiernan: What I meant here is that there is no interface that allows for running a power flow in the loop from PSI on a system with a network reduction (besides parallel branches and breaker-switches).

src/network_models/power_flow_evaluation.jl
```diff
@@ -631,24 +635,33 @@ function calculate_aux_variable_value!(container::OptimizationContainer,
...
         n_branches = length(parallel_brs)
         for br in parallel_brs
             if br isa U
                 @assert T <: LineFlowAuxVariableType "Only LineFlowAuxVariableType aux vars " *
                                                      "can be used for parallel branches: got $T"
-                @assert PSY.get_r(br) + im * PSY.get_x(br) == impedance "All parallel " *
-                                                                        "branches must have the same impedance"
+                if PSY.get_r(br) + im * PSY.get_x(br) != impedance
+                    throw(
+                        error(
```
jd-lara, line 655: we should test this error in the EI or WECC to confirm this doesn't happen

jd-lara (approved):

PR #1379, opened 2025-09-15
Fix Hydro and Source tests

rodrigomha:
Requires https://github.com/NREL-Sienna/PowerSystemCaseBuilder.jl/pull/160


src/devices_models/devices/common/objective_function/import_export.jl
jd-lara, line 1: @GabrielKS this is the only file you need to review

PR #1383, opened 2025-09-23
Sm/restore_scuc

SebastianManriqueM:
This PR address:
- Compatibility with network reductions of G-1 with reserves and G-1.
- Restoring all SCUC tests (N-1, G-1 with reserves)
- Restoring DLR tests
- Adding initial tests for G-1 formulation
- Address most of the comments made by Gabriel in the last review, especially the ones related to compare devices and not names.

PENDING
- Ensure compatibility with Network reductions for N-1 formulation (improve testing and cases)
- Refactoring N-1 to become a branch formulation
- Code cleaning, especially for functions and implementations related to network reduction compatibility. Some review and improvement on this before the N-1 refactoring might be helpful.

SUGGESTION: 
- One way to reuse more code between G-1 and G-1 with reserves could be refactor G-1 code to become also a reserves formulation. Something like “SecurityConstrainedImplicitReservesFormulation”. Besides that advantage, it will also prevent the user to use simultaneously G-1 and G-1 with reserves formulation.

NOTE
- Currently G-1 formulation is implemented for thermal generators in such a way that thermal generators respond to outages in all generator types. I think that if we refactor this code to become a reserves formulation the user might be able to select the types or even the units that might be part of the reserves, which I think is more interesting and very straightforward to implement.

NOTE ON THIS SPECIFIC PATTERN:
I used for g-1 this pattern (And probably N-1 refactoring will use):

```

generator_outages_pairs = PSY.get_component_supplemental_attribute_pairs(
        PSY.Generator,
        PSY.UnplannedOutage,
        sys,
    )

    associated_outages = unique([outage for (_, outage) in generator_outages_pairs]) 

```

The current design it is required that the user adds the outage supplemental attribute (SA) to the generator (One SA instance for each outage that will be considered). This means that for a G-2, G-3... cases I can have 2 generators associated with the same SA instance.
All tests passed locally.

So I know that pattern is not the best (uses "unique()") but I did that since I dont have available a function that allows me to get the SAs attached to an specific component type. SUgestions are welcome!. My sugestion would be implement that function in IS and PSY5.

PD @jd-lara please review the last commit, for some reason I had to push my project.toml. Hopefully this will not affect too much your merge.


m-bossart:
Issues to address before merge of this PR

- [ ] Improve test for DLRs 
- [ ] Fix getting the branch rating to get rating_b for reserve deliverability

src/core/network_reductions.jl
```diff
@@ -34,6 +43,68 @@ function _has_keys_nested(nested_dict::Dict, keys::Vector)
...
+    return (false, EMPTY_BRANCH_NAME_MATCH)
+end
+
+function _search_for_reduced_branch_expression(
+    reduction_tracker::BranchReductionOptimizationTracker,
+    series_chain::Vector{Any},
+    component_key_1::Type{U},
+    component_key_2::String,
+    expression_key_1::Type{T},
+    expression_key_2::Int,
```
m-bossart, line 81: Can we make these more descriptive names: `component_type`, `component_name`, `expression_type`, `expression_stage`

m-bossart: Should make this change throughout

src/core/network_reductions.jl
```diff
@@ -83,6 +154,102 @@ function _search_for_reduced_branch_variable(
...
 end
 
+function _add_expression_to_tracker!(
+    ::BranchReductionOptimizationTracker,
+    ::JuMP.AffExpr,
+    reduction_entry::PSY.ACTransmission,
+    component_key_1::Type{U},
+    component_key_2::String,
+    expression_key_1::Type{T},
+    expression_key_2::Int,
```
m-bossart, line 164: Can remove the names here because they are unused

src/core/network_reductions.jl
```diff
@@ -176,6 +367,17 @@ function get_branch_name_constraint_axis(
     return name_axis
 end
 
+function get_branch_name_variable_axis(nrd::PNM.NetworkReductionData)
+    ac_transmission_types = PNM.get_ac_transmission_types(nrd)
+    all_branch_maps_by_type = nrd.all_branch_maps_by_type
+    name_axis = Vector{String}()
+    for T in ac_transmission_types
+        names_by_type = get_branch_name_variable_axis(all_branch_maps_by_type, T)
+        name_axis = vcat(name_axis, names_by_type) # TODO change implementation to avoid repeated concatonating
```
m-bossart, line 376: Address TODO (I will do this)

src/devices_models/device_constructors/branch_constructor.jl
```diff
@@ -322,49 +322,45 @@ function construct_device!(
...
+        @info "No associated outage supplemental attributes found assicuiatted with devices: $V. Skipping contingency variable addition for that service."
+        return
     end
 
-    lodf = get_LODF_matrix(network_model)
-    removed_branches = PNM.get_removed_branches(lodf.network_reduction_data)
+    network_reduction = get_network_reduction(network_model)
+    branches_names = PNM.get_retained_branches_names(network_reduction)
+
     # TODO: Security constrained. This method might not be needed. Analyze why is here
```
m-bossart, line 338: Is this comment still relevant?

src/devices_models/devices/AC_branches.jl
```diff
@@ -497,17 +519,158 @@ function get_min_max_limits(
...
+end
+
+function _get_min_max_limits(
+    branch::PSY.ACBranch,
+    T::Type{<:PostContingencyEmergencyRateLimitConstraint},
+    ::Type{<:AbstractBranchFormulation},
+)
+    rating = get_rating(branch, T)
+    return (min = -1 * rating, max = rating)
+end
```
m-bossart, line 551: Do we need these two additional methods? I think we can just define the inner method in line 525 to get the rating and return `(min = -1 * rating, max = rating)`

src/devices_models/devices/static_injection_security_constrained_models.jl
```diff
@@ -316,40 +450,54 @@ function add_constraints!(
...
             limits = get_min_max_limits(
                 device,
                 ActivePowerVariableLimitsConstraint,
-                U,
+                W,
             )
 
             for t in time_steps
-                con_ub[gen_outage_name, device_name, t] =
+                #TODO HOW WE SHOULD HANDLE THE EXPRESSIONS AND CONSTRAINTS RELATED TO THE OUTAGE OF THE GENERATOR RESPECT TO ITSELF?
```
m-bossart, line 484: I think you addressed this issue below and the comment can be deleted?

src/devices_models/devices/static_injection_security_constrained_models.jl
```diff
@@ -358,349 +506,611 @@ function add_constraints!(
...
-                        expression[name, t],
-                        variable[name, t],
-                        -1.0,
-                    )
-                end
-                continue
+    for (device, outage) in generator_outages_pairs
+        # if !(outage in associated_outages)
+        #     continue
+        # end
```
m-bossart, line 556: Why is this commented? The expression is built based on unique outages, but the loop is for all of the pairs. It seems like we might have repeated expression in this case?

src/devices_models/devices/static_injection_security_constrained_models.jl
```diff
@@ -358,349 +506,611 @@ function add_constraints!(
...
-                    JuMP.@constraint(get_jump_model(container),
-                        expressions[outage_name, branch_name, t] >=
-                        limits.min)
+    expressions = get_expression(container, U(), R, "")
+    for ac_type in ac_transmission_types
+        !(ac_type in network_model.modeled_branch_types) && continue
+        for map in NETWORK_REDUCTION_MAPS
+            network_reduction_map = all_branch_maps_by_type[map]
+            !haskey(network_reduction_map, ac_type) && continue
+            #!haskey(network_model.modeled_branch_types, ac_type) && continue
```
m-bossart, line 1074: The check here replacing line 1073 and 1074 should be: `!(ac_type in modeled_branch_types) && continue`. modeled_branch_types is what is used when constructing the container so this will ensure we are matching the container.

src/devices_models/devices/static_injection_security_constrained_models.jl
```diff
@@ -712,15 +1122,16 @@ function add_constraints!(
...
 ) where {
     T <: PostContingencyRampConstraint,
     U <: AbstractContingencyVariableType,
     V <: PSY.Generator,
-    R <: Union{PSY.Reserve{PSY.ReserveDown}, PSY.Reserve{PSY.ReserveUp}, Nothing},
+    R <: Union{PSY.Reserve{PSY.ReserveDown}, PSY.Reserve{PSY.ReserveUp}, PSY.Generator},
     F <: Union{AbstractSecurityConstrainedUnitCommitment,
         AbstractSecurityConstrainedReservesFormulation},
     N <: AbstractPTDFModel,
+    S <: Union{PSY.Reserve{PSY.ReserveDown}, PSY.Reserve{PSY.ReserveUp}, String},
```
m-bossart, line 1134: We should reconsider the design which leads to a union of a reserve and a string.

src/devices_models/devices/static_injection_security_constrained_models.jl
```diff
@@ -1496,50 +1957,271 @@ function add_to_expression!(
...
+            T(),
+            R,
+            string.(IS.get_uuid.(associated_outages)),
+            device_names,
+            time_steps;
+            meta = "$service_name -ub",
+        )
+    expressions = get_expression(container, U(), R, service_name)
+    for ac_type in ac_transmission_types
+        !(ac_type in network_model.modeled_branch_types) && continue
```
m-bossart, line 2147: Change to `!(ac_type in modeled_branch_types)`

src/devices_models/devices/static_injection_security_constrained_models.jl
SebastianManriqueM: In this expression I'm adding the power of all generators that are with outage (Associated to the same outage instance). This is to allow G-1, G-2, G-3...etc. At the beginning we were considering only G-1 and I forgot to remove the comment. Thanks Matt!

src/devices_models/devices/static_injection_security_constrained_models.jl
SebastianManriqueM: That mess was caused because I had added ramp constraints on the post contingency variables and we have a service and thermal formulation. Now both will be reserve formulations and the ramps will be in the procurement

PR #1384, opened 2025-09-23
add market bid cost for renewable

luke-kiernan:
To run this, you'll need to revert the most recent commit to `InfrastructureSystems.jl`.

### DIFF
--- a/src/devices_models/devices/common/objective_function/market_bid.jl
+++ b/src/devices_models/devices/common/objective_function/market_bid.jl
@@ -210,6 +210,55 @@ function validate_mbc_component(::ShutdownCostParameter, device::PSY.StaticInjec
     _validate_eltype(Float64, device, shutdown, " for shutdown cost")
 end
 
+# Renewable-specific validations that warn when costs are nonzero.
+# There warnings are captured by the with_logger, though, so we don't actually see them.
+function validate_mbc_component(
+    ::StartupCostParameter,
+    device::PSY.RenewableDispatch,
+    model,
+)
+    startup = PSY.get_start_up(PSY.get_operation_cost(device))
+    apply_maybe_across_time_series(device, startup) do x
+        if x != PSY.single_start_up_to_stages(0.0)
+            # println("should see warning about startup")
+            @warn "Nonzero startup cost detected for renewable generation $(get_name(device))." maxlog =
+                1
+        end
+    end
+end
+
+function validate_mbc_component(
+    ::ShutdownCostParameter,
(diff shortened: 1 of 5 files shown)


test/test_market_bid_cost.jl
jd-lara, line 1: These tweaks and changes to the system should go on PSB so that we have good testing systems there. We would want to use some of those for tutorials and other testing

test/test_market_bid_cost.jl
```diff
@@ -47,6 +47,109 @@ const SEL_MULTISTART = make_selector(ThermalMultiStart, "115_STEAM_1")
...
+        val = getfield(mbc, field)
+        if val isa IS.TimeSeriesKey
+            ts = PSY.get_time_series(old_comp, val)
+            add_time_series!(sys, new_comp, ts)
+        end
+    end
+    return
+end
+
+function replace_with_renewable!(
```
jd-lara, line 67: update PSB case

src/devices_models/devices/renewable_generation.jl
```diff
@@ -46,6 +46,7 @@ function get_min_max_limits(
     return PSY.get_reactive_power_limits(device)
 end
 
+# aha. this is why it's always expecting 2 time series.
```
jd-lara, line 49: ?

src/devices_models/devices/common/objective_function/market_bid.jl
```diff
@@ -210,6 +210,55 @@ function validate_mbc_component(::ShutdownCostParameter, device::PSY.StaticInjec
...
+function validate_mbc_component(
+    ::IncrementalCostAtMinParameter,
+    device::PSY.RenewableDispatch,
+    model,
+)
+    no_load_cost = PSY.get_no_load_cost(PSY.get_operation_cost(device))
+    if !isnothing(no_load_cost)
+        apply_maybe_across_time_series(device, no_load_cost) do x
+            if x != 0.0
+                # println("should see warning about no-load")
```
jd-lara, line 254: remove comments print or use a logging Debug

luke-kiernan: The issue is that the `with_logger` context manager inside `build!` grabs all log messages, so neither `@warn` nor `@debug` messages show up in the output. Nevertheless, I'll remove them.

test/test_market_bid_cost.jl
luke-kiernan: Will get a solid test case in PSB once I have things working and code is solidified.

PR #1388, opened 2025-09-26
Mb/interface double circuit

m-bossart:
- Uses the network reduction maps when adding to interface expressions. 
- Adds tests for adding an interface to a double circuit and series reductions (including testing for bad user data input).
- Changes the rating of a double circuit to be the sum of the circuit ratings.
- Adds a test that branch bounds are set as expected for reductions (parallel and series chain)

### DIFF
--- a/src/devices_models/devices/AC_branches.jl
+++ b/src/devices_models/devices/AC_branches.jl
@@ -403,7 +403,7 @@ end
 ################################## Rate Limits constraint_infos ############################
 
 function get_rating(double_circuit::Set{PSY.ACTransmission})
-    return minimum([PSY.get_rating(circuit) for circuit in double_circuit])
+    return sum([PSY.get_rating(circuit) for circuit in double_circuit])
 end
 function get_rating(series_chain::Vector{Any})
     return minimum([get_rating(segment) for segment in series_chain])
--- a/src/devices_models/devices/common/add_to_expression.jl
+++ b/src/devices_models/devices/common/add_to_expression.jl
@@ -1629,33 +1629,166 @@ function add_to_expression!(
     ::Type{FlowActivePowerVariable},
     service::PSY.TransmissionInterface,
     model::ServiceModel{PSY.TransmissionInterface, V},
+    network_model::NetworkModel{<:PM.AbstractActivePowerModel},
 ) where {V <: Union{ConstantMaxInterfaceFlow, VariableMaxInterfaceFlow}}
+    network_reduction_data = get_network_reduction(network_model)
+    all_branch_maps_by_type = network_reduction_data.all_branch_maps_by_type
     expression = get_expression(container, InterfaceTotalFlow(), PSY.TransmissionInterface)
     service_name = get_service_name(model)
     direction_map = PSY.get_direction_mapping(service)
-    specified_direction_mapping = isempty(direction_map)
-    for (device_type, devices) in get_contributing_devices_map(model)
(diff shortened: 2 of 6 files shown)


src/devices_models/devices/AC_branches.jl
```diff
@@ -403,7 +403,7 @@ end
 ################################## Rate Limits constraint_infos ############################
 
 function get_rating(double_circuit::Set{PSY.ACTransmission})
-    return minimum([PSY.get_rating(circuit) for circuit in double_circuit])
+    return sum([PSY.get_rating(circuit) for circuit in double_circuit])
 end
 function get_rating(series_chain::Vector{Any})
```
jd-lara, line 408:
```suggestion
function get_rating(series_chain::Vector{<:ACTransmission})
```

src/devices_models/devices/common/add_to_expression.jl
```diff
@@ -1629,35 +1629,168 @@ function add_to_expression!(
...
+    end
+end
+
+function _get_direction(
+    arc_tuple::Tuple{Int, Int},
+    reduction_entry::Set{PSY.ACTransmission},
+    direction_map::Dict{String, Int},
+    network_reduction_data::PNM.NetworkReductionData,
+)
+    directions = [
```
jd-lara, line 1701: This code can be simplified into a single loop instead of two. If there are only a few items in each set it should be fine but leave a comment that this is fine because there aren't like 10_000 parallel branches.

src/devices_models/devices/common/add_to_expression.jl
```diff
@@ -1629,35 +1629,168 @@ function add_to_expression!(
...
+    ]
+    # direction of segments relative to the reduced degree two chain: 
+    _, segment_orientations =
+        PNM._get_chain_data(arc_tuple, reduction_entry, network_reduction_data)
+    segment_directions = [x == :FromTo ? 1.0 : -1.0 for x in segment_orientations]
+    net_directions = mapping_directions .* segment_directions
+    if allequal(net_directions)
+        return first(net_directions)
+    else
+        throw(
```
jd-lara, line 1735: Nice catch, I wouldn't be surprised if we hit this somewhere.

src/devices_models/devices/common/add_to_expression.jl
```diff
@@ -1629,35 +1629,168 @@ function add_to_expression!(
...
+            ArgumentError(
+                "The interface direction mapping for degree two chain with arc $(arc_tuple) is inconsistent. Check the mapping entries and the orientation of the segment arcs within the chain.",
+            ),
+        )
+    end
+end
+
+function _reduction_entry_in_interface(
```
jd-lara: nit: 
```suggestion
function _reduced_entry_in_interface(
```

src/devices_models/devices/AC_branches.jl
m-bossart: The series chain isn't `Vector{<:ACTransmission}` because it can have double circuits. In the reduction object it is `Vector{Any}`. We should have a bigger conversation about how to improve that design

jd-lara: Put this as a comment in the code please so we remember to come back to it

m-bossart: I added a general comment in the network reduction file because this happens all over in PSI and should be updated everywhere. I also created an issue in PNM.

jd-lara (approved):

PR #1389, opened 2025-09-29
Mb/fixes for decomposition

m-bossart:


src/core/network_reductions.jl
```diff
@@ -335,6 +336,7 @@ function get_branch_name_constraint_axis(
     ac_transmission_types = PNM.get_ac_transmission_types(nrd)
     name_axis = Vector{String}()
     for ac_type in ac_transmission_types
+        (ac_type ∉ modeled_branch_types) && continue
```
SebastianManriqueM, line 339: Hello Matt, I did not catch this. I think I lost my changes with the mess after the rebase.

I've reviewed and I think we can merge it!. Thanks

PR #1396, opened 2025-10-01
no min gen power constraint for hydro storage

luke-kiernan:
Resolve following error:
```
MethodError: no method matching proportional_cost(::PowerSimulations.OptimizationContainer, ::HydroGenerationCost, ::OnVariable, ::HydroDispatch, ::HydroCommitmentRunOfRiver, ::Int64)
```

### DIFF
--- a/src/devices_models/devices/common/objective_function/market_bid.jl
+++ b/src/devices_models/devices/common/objective_function/market_bid.jl
@@ -373,7 +373,14 @@ end
 ################# PWL Constraints ################
 ##################################################
 
-# without this, you get "variable OnVariable__RenewableDispatch is not stored"
+_include_min_gen_power_in_constraint(
+    ::PSY.EnergyReservoirStorage,
+    ::ActivePowerInVariable,
+) = false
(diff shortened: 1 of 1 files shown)


src/devices_models/devices/common/objective_function/market_bid.jl
```diff
@@ -373,7 +373,14 @@ end
 ################# PWL Constraints ################
 ##################################################
 
-# without this, you get "variable OnVariable__RenewableDispatch is not stored"
+_include_min_gen_power_in_constraint(
```
jd-lara, line 376: This should go in HPS

luke-kiernan: That'll require adding HydroPowerSimulations as a dependency of StorageSystemSimulations. Should it go in StorageSystemSimulations instead?
edit: or at least I suspect it will. let me check.
edit^2: no, seems like it's fine. It'll need to be a dependency of the tests, but that's already the case.

PR #1397, opened 2025-10-01
Support 3d realized results

daniel-thom:


src/core/results_by_time.jl
```diff
@@ -46,6 +46,19 @@ function _check_column_consistency(
     # TODO:
 end
 
+function _check_column_consistency(
+    data::SortedDict{Dates.DateTime, DataFrame},
+    cols::NTuple{N, Vector{String}},
+) where {N}
```
jd-lara, line 52: is there a reason to leave this N untyped?

jd-lara (approved):
LGTM

PR #1398, opened 2025-10-01
Support 3d initial conditions

daniel-thom:
I tested this change with HydroPowerSimulations branch feat/3d-resultsv2.

### DIFF
--- a/src/core/initial_conditions.jl
+++ b/src/core/initial_conditions.jl
@@ -60,18 +60,18 @@ get_ic_type(
 Stores data to populate initial conditions before the build call
 """
 mutable struct InitialConditionsData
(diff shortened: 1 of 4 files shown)


luke-kiernan (commented):
I still get a few errors along the following lines when I run the HPS tests, but I believe they're unrelated.
```
MethodError: no method matching get_column_names_from_axis_array(::JuMP.Containers.DenseAxisArray{Float64, 2, Tuple{Vector{String}, Vector{Int64}}, Tuple{JuMP.Containers._AxisLookup{Dict{String, Int64}}, JuMP.Containers._AxisLookup{Dict{Int64, Int64}}}})
```

PR #1400, opened 2025-10-07
Lk/storage mbc

luke-kiernan:
Decremental support, for storage market bid cost objects. Need to write a few more tests, but mostly there.

### DIFF
--- a/src/devices_models/devices/common/objective_function/market_bid.jl
+++ b/src/devices_models/devices/common/objective_function/market_bid.jl
@@ -49,7 +49,7 @@ _consider_parameter(
 ) where {T, D} = has_container_key(container, StopVariable, T)
 
 _consider_parameter(
-    ::IncrementalCostAtMinParameter,
+    ::AbstractCostAtMinParameter,
     container::OptimizationContainer,
     ::DeviceModel{T, D},
 ) where {T, D} = has_container_key(container, OnVariable, T)
@@ -57,13 +57,13 @@ _consider_parameter(
 # For slopes and breakpoints, the relevant variables won't have been created yet, so we'll
 # just check all components for the presence of the relevant time series
 _consider_parameter(
-    ::IncrementalPiecewiseLinearSlopeParameter,
+    ::AbstractPiecewiseLinearSlopeParameter,
     ::OptimizationContainer,
     ::DeviceModel{T, D},
 ) where {T, D} = true
 
 _consider_parameter(
-    ::IncrementalPiecewiseLinearBreakpointParameter,
+    ::AbstractPiecewiseLinearBreakpointParameter,
     ::OptimizationContainer,
     ::DeviceModel{T, D},
 ) where {T, D} = true
@@ -83,6 +83,10 @@ _has_parameter_time_series(::IncrementalCostAtMinParameter, device::PSY.StaticIn
     _has_market_bid_cost(device) &&
     is_time_variant(PSY.get_incremental_initial_input(PSY.get_operation_cost(device)))
 
+_has_parameter_time_series(::DecrementalCostAtMinParameter, device::PSY.StaticInjection) =
+    _has_market_bid_cost(device) &&
+    is_time_variant(PSY.get_decremental_initial_input(PSY.get_operation_cost(device)))
+
 _has_parameter_time_series(
     ::IncrementalPiecewiseLinearSlopeParameter,
     device::PSY.StaticInjection,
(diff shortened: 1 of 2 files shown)


src/parameters/update_cost_parameters.jl
```diff
@@ -114,52 +90,26 @@ function handle_variable_cost_parameter(
...
-            parameter_array,
-            parameter_multiplier,
-            attributes,
-            component,
-            t,
-        )
-    end
-    return
-end
+# Helper function to get the appropriate getter function
```
jd-lara: is this being compiled appropriately?

luke-kiernan: Looking into the output of `@code_warntype` on `handle_variable_cost_parameter`, here's what I notice:
1. Instances of type instability match the field types: for example, `Union{Nothing, CostCurve{PiecewiseIncrementalCurve}, TimeSeriesKey}` for the output of `_get_parameter_field` when the parameter is of type `IncrementalPiecewiseLinearSlopeParameter`.
2. `param` is marked as `param::Core.Const(PowerSimulations.FooParameter())`, which says that the `Union` type is being specialized.

That leads me to think it's fine. (I expected to see type instability from `_maybe_tuple`, but if parameter type is known at compile time, I guess it's a non-issue.)

I could separate it out, though. Let me know what you decide

src/devices_models/devices/common/objective_function/market_bid.jl
```diff
@@ -216,43 +220,63 @@ end
 # There warnings are captured by the with_logger, though, so we don't actually see them.
 function validate_mbc_component(
     ::StartupCostParameter,
-    device::PSY.RenewableDispatch,
+    device::Union{PSY.RenewableDispatch, PSY.Storage},
```
GabrielKS, line 227: maybe we should define a `const MBCDemandable` (but with a better name) somewhere

src/devices_models/devices/common/objective_function/market_bid.jl
```diff
@@ -216,43 +220,63 @@ end
...
     model,
 )
     startup = PSY.get_start_up(PSY.get_operation_cost(device))
     apply_maybe_across_time_series(device, startup) do x
         if x != PSY.single_start_up_to_stages(0.0)
-            @warn "Nonzero startup cost detected for renewable generation $(get_name(device))." maxlog =
-                1
+            println(
```
GabrielKS: shouldn't have printlns in merged code

src/devices_models/devices/common/objective_function/market_bid.jl
```diff
@@ -742,7 +766,7 @@ function _add_variable_cost_to_objective!(
         error("Component $(component_name) is not allowed to participate as a demand.")
     end
     add_pwl_term!(
-        false,
+        false, # I suspect this is a problem. could very well be decremental, storage
```
GabrielKS: ??

src/devices_models/devices/common/objective_function/market_bid.jl
luke-kiernan: I know...this is my workaround for being unable to see those `@warn` messages, while still using `build!` (not the `build_impl!`-and-capture-logs trick). I'll look into better solutions: maybe I grab the log file after-the-fact and print its contents to stdout.

src/devices_models/devices/common/objective_function/market_bid.jl
luke-kiernan: My guess: the decremental function has type parameter `AbstractControllablePowerLoadFormulation`, which means that we haven't addressed storage yet. It's a strange place for the comment and not terribly clear, though

luke-kiernan: Oh, there's a `PSI._add_variable_cost_to_objective!` is defined in SSS. So it isn't a problem.

It would be cleaner to have types (like your `const MBCDemandable`) for this sort of stuff, so we wouldn't need to define it separately in SSS...but the device formulations might mean we can't do that.

I'll remove the comment.

src/devices_models/devices/common/objective_function/market_bid.jl
luke-kiernan: Yeah...the amount of types floating around is a contributing factor to code complexity...I'll keep that in mind, look for opportunities to do stuff like that.

src/devices_models/devices/common/objective_function/market_bid.jl
jd-lara: you can pass the kwarg `console_level = Logging.Info` to build to see the logs in the console.

PR #1402, opened 2025-10-08
fix HVDC test

luke-kiernan:
Fix one of the 2 commented-out tests in `test/test_power_flow_in_the_loop.jl`. I had to remove the power comparison at the `from` bus of the HVDC line: the test used to make that bus PV (despite no generation), but now PF rejects such a system as bad input.

### DIFF
--- a/test/test_power_flow_in_the_loop.jl
+++ b/test/test_power_flow_in_the_loop.jl
@@ -160,8 +160,54 @@ end
     # reductions. we'd have to pass kwargs all the way down to add_power_flow_data!.
 end
 
-#=
-# the way this test removes components is problematic.
+"""
+Remove components from a system while preserving time series data on other components.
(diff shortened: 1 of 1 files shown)


test/test_power_flow_in_the_loop.jl
```diff
@@ -160,8 +160,54 @@ end
...
 end
 
-#=
-# the way this test removes components is problematic.
+"""
+Remove components from a system while preserving time series data on other components.
+(There ought to be a better way to do this than this roundabout method, but I have yet to find one:
+simply removing the time series on the components to be removed is not sufficient.)
+"""
+function remove_components!(sys::System, components::Vector{PSY.StaticInjection})
```
jd-lara, line 168: @daniel-thom should we enable something like this in PSY when we remove components with time series.

PR #1405, opened 2025-10-13
Mb/subnetwork bug fix

m-bossart:
Two changes

1. `get_available_components` filters based on the availability of `ACBus`
2. Gets network reduction data from the Ybus when building a copperplate model with multiple subsystems because the network reduction is used when assigning buses to subnetworks.

### DIFF
--- a/src/utils/powersystems_utils.jl
+++ b/src/utils/powersystems_utils.jl
@@ -44,6 +44,9 @@ function get_available_components(
     end
 end
 
+_filter_function(x::PSY.ACBus) =
+    PSY.get_bustype(x) != PSY.ACBusTypes.ISOLATED && PSY.get_available(x)
+
 function get_available_components(
     model::NetworkModel,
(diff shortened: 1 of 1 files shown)


src/utils/powersystems_utils.jl
```diff
@@ -51,7 +51,7 @@ function get_available_components(
 )
     subsystem = get_subsystem(model)
     return PSY.get_components(
-        x -> PSY.get_bustype(x) != PSY.ACBusTypes.ISOLATED,
+        x -> PSY.get_bustype(x) != PSY.ACBusTypes.ISOLATED && PSY.get_available(x),
```
jd-lara: when the filter functions get this complex, it is better to define it and pass it like 

```Julia
_filter_function(x) = whatever(x)
function get_available_components(
...
return  get_components(_filter_function, ...)
end
```

PR #1408, opened 2025-10-15
bug fix

m-bossart:
Bug fix for : https://github.com/NREL-Sienna/PowerSimulationsDecomposition.jl/issues/12
Once the outage generator was found, all subsequent multipliers were set to zero when building the active power balance expressions.


m-bossart:
image

PR #1416, opened 2025-10-22
Refactoring and new features for SCUC models

SebastianManriqueM:
This PR is addressing:

- Refactoring of thermal g-1 formulation to transform into "ContingencyReserveWithDeliverabilityConstraints". All the code related to the old Thermal formulation was removed.

- Adding capability of G-n SCUC formulations to run with Copperplate and AreaBalance.

- Addressing past comments from other PR from Matt and Gabriel.

- Improving testing for SCUC models

QUESTION:
Does it makes sense call the formulation "RampReserveWithDeliverabilityConstraints" for CopperPlate Model?, should we use a different name?.

GAPS:

- Only "RampReserveWithDeliverabilityConstraints" is running with CopperPlate and AreaBalance. I'd like your feedback if everything looks good, to proceed also with ContingencyReserveWithDeliverabilityConstraints

- There is still pending a refactoring N-1 formulation so it becomes a branch formulation instead of a network formulation.

NOTE:
I ran all the test locally and those are passing.


jd-lara:
closing in favor of #1462 and #1463

Issue #1421, opened 2025-10-28
MethodError at DecisionModel build


PabloBotinGP:
**Describe the bug**
StorageDispatchWithReserves formulation fails during DecisionModel build with a MethodError for missing _consider_parameter method. The error occurs when PowerSimulations tries to process market bid parameters for EnergyReservoirStorage devices using StorageDispatchWithReserves formulation.

The specific error is:
MethodError: no method matching _consider_parameter(::PowerSimulations.DecrementalCostAtMinParameter, ::PowerSimulations.OptimizationContainer, ::DeviceModel{EnergyReservoirStorage, StorageDispatchWithReserves})

**To Reproduce**
System Setup:

Julia Version 1.11.7 (arm64-apple-darwin24.0.0)
PowerSimulations v0.30.1 (psy5 branch)
StorageSystemsSimulations v0.12.0 (psy5 branch)
PowerSystems v4.6.2 (psy5 branch)

Steps to reproduce:

Load a PowerSystems System with EnergyReservoirStorage components
Create a storage device model using StorageDispatchWithReserves formulation:

using PowerSystems
using PowerSimulations  
using StorageSystemsSimulations
using Xpress

# Load system (ensure it contains EnergyReservoirStorage devices)
sys_path = joinpath(@__DIR__, "..", "DA_sys.json")
sys = System(sys_path)

# Define Storage Model with StorageDispatchWithReserves
storage_model = DeviceModel(
    EnergyReservoirStorage,
    StorageDispatchWithReserves;
    attributes=Dict(
        "reservation" => true,
        "energy_target" => false,
        "cycling_limits" => false,
        "regularization" => true,
    ),
)

# Create Unit Commitment template
template_uc = template_unit_commitment(;
    network = NetworkModel(CopperPlatePowerModel))

# Set device models
set_device_model!(template_uc, ThermalStandard, ThermalBasicUnitCommitment)
set_device_model!(template_uc, ThermalMultiStart, ThermalBasicUnitCommitment)
set_device_model!(template_uc, RenewableDispatch, RenewableFullDispatch)
set_device_model!(template_uc, HydroDispatch, HydroDispatchRunOfRiver)
set_device_model!(template_uc, PowerLoad, StaticPowerLoad)
set_device_model!(template_uc, storage_model)  # This causes the error

# Set up solver and decision model
solver = optimizer_with_attributes(Xpress.Optimizer, "MIPRELSTOP" => 0.1)
problem = DecisionModel(
[...]

PR #1422, opened 2025-10-29
Lk/issue 1421 hotfix

luke-kiernan:
Hotfix for issue #1421.

When I run the PSI tests, I get 2 failures and 1 error, but it looks like all of them are a consequence of a data issue: `ArgumentError: Found more than one matching time series: 2` at the `"ImportExportSource Source With CopperPlate and TimeSeries"` test.
```julia
    @test build!(model; output_dir = mktempdir()) == PSI.ModelBuildStatus.BUILT # fails
    @test solve!(model) == PSI.RunStatus.SUCCESSFULLY_FINALIZED # fails

    res = OptimizationProblemResults(model) # errors
```

### DIFF
--- a/src/devices_models/devices/common/objective_function/market_bid.jl
+++ b/src/devices_models/devices/common/objective_function/market_bid.jl
@@ -80,41 +80,26 @@ _has_parameter_time_series(::StartupCostParameter, device::PSY.StaticInjection) 
 _has_parameter_time_series(::ShutdownCostParameter, device::PSY.StaticInjection) =
     is_time_variant(PSY.get_shut_down(PSY.get_operation_cost(device)))
 
-_has_parameter_time_series(::IncrementalCostAtMinParameter, device::PSY.StaticInjection) =
+_has_parameter_time_series(
+    ::T,
+    device::PSY.StaticInjection,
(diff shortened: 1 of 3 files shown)


src/devices_models/devices/common/objective_function/market_bid.jl
```diff
@@ -48,22 +48,23 @@ _consider_parameter(
     ::DeviceModel{T, D},
 ) where {T, D} = has_container_key(container, StopVariable, T)
 
+# FIXME storage doesn't currently have an OnVariable. should it have one?
```
jd-lara, line 51: Not at the moment, we can this later as we update the formulation. @rodrigomha we can open an issue to include this if needed to support min charge/min discharge

jd-lara (commented):
I'll merge to fix the error and come back to the TODOs later

src/devices_models/devices/common/objective_function/market_bid.jl
luke-kiernan: Sure. It might not need an `OnVariable`. I'll see what makes sense when I revisit the storage MBC stuff, after I finish up the load MBC stuff.

PR #1423, opened 2025-10-29
Decremental MarketBidCost v2

luke-kiernan:
Continuation of PR NREL-Sienna/PowerSimulations.jl#1400, but now I have write permissions so I don't need to open it from a fork.

### DIFF
--- a/src/core/optimization_container.jl
+++ b/src/core/optimization_container.jl
@@ -2078,8 +2078,8 @@ function get_time_series_initial_values!(
     )
     ts_values = IS.get_time_series_values(
         component,
-        forecast,
-        initial_time;
+        forecast;
+        start_time = initial_time,
         len = length(time_steps),
         ignore_scaling_factors = true,
     )
--- a/src/parameters/add_parameters.jl
+++ b/src/parameters/add_parameters.jl
@@ -401,7 +401,7 @@ _get_max_tranches(data::AbstractDict) = maximum(_get_max_tranches.(values(data))
 
 # Iterate through all periods of a piecewise time series and return the maximum number of tranches
 function get_max_tranches(device::PSY.Device, piecewise_ts::IS.TimeSeriesKey)
-    data = PSY.get_data(PSY.get_time_series(device, piecewise_ts, nothing, nothing))
+    data = PSY.get_data(PSY.get_time_series(device, piecewise_ts))
     max_tranches = _get_max_tranches(data)
     return max_tranches
 end
--- a/test/test_market_bid_cost.jl
+++ b/test/test_market_bid_cost.jl
@@ -441,14 +441,18 @@ end
     # enough to change the decisions that form the correct solution
 
     # Scenario 1: hot and warm starts
-    c_sys5_pglib0a = create_multistart_sys(false, 1.0, 7.45; add_ts = false)
-    c_sys5_pglib1a = create_multistart_sys(false, 1.0, 7.45)
-    c_sys5_pglib2a = create_multistart_sys(true, 1.0, 7.45)
+    load_mult_a = 1.0
+    therm_mult_a = 7.4
+    c_sys5_pglib0a = create_multistart_sys(false, load_mult_a, therm_mult_a; add_ts = false)
+    c_sys5_pglib1a = create_multistart_sys(false, load_mult_a, therm_mult_a)
+    c_sys5_pglib2a = create_multistart_sys(true, load_mult_a, therm_mult_a)
 
     # Scenario 2: hot and cold starts
-    c_sys5_pglib0b = create_multistart_sys(false, 1.05, 7.4; add_ts = false)
-    c_sys5_pglib1b = create_multistart_sys(false, 1.05, 7.4)
-    c_sys5_pglib2b = create_multistart_sys(true, 1.05, 7.4)
+    load_mult_b = 1.05
+    therm_mult_b = 7.4
+    c_sys5_pglib0b = create_multistart_sys(false, load_mult_b, therm_mult_b; add_ts = false)
+    c_sys5_pglib1b = create_multistart_sys(false, load_mult_b, therm_mult_b)
+    c_sys5_pglib2b = create_multistart_sys(true, load_mult_b, therm_mult_b)
 
     test_generic_mbc_equivalence(c_sys5_pglib0a, c_sys5_pglib1a; multistart = true)
     test_generic_mbc_equivalence(c_sys5_pglib0b, c_sys5_pglib1b; multistart = true)


src/devices_models/devices/common/objective_function/market_bid.jl
```diff
@@ -221,45 +224,42 @@ end
...
     device::Union{PSY.RenewableDispatch, PSY.Storage},
-    model,
 )
     startup = PSY.get_start_up(PSY.get_operation_cost(device))
     apply_maybe_across_time_series(device, startup) do x
         if x != PSY.single_start_up_to_stages(0.0)
-            println(
-                "Nonzero startup cost detected for renewable generation or storage device $(get_name(device)).",
-            )
+            #println(
```
GabrielKS, line 231: TODO clean up (same below)

src/devices_models/devices/common/objective_function/market_bid.jl
```diff
@@ -442,7 +485,10 @@ function _add_pwl_constraint!(
     # time-variable P1 is problematic, so for now we require P1 to be constant. Thus we can
     # just look up what it is currently fixed to and use that here without worrying about
     # updating.
-    if _include_min_gen_power_in_constraint(component, U())
+    if _include_constant_min_gen_power_in_constraint(component, U(), D())
+        # TODO this seems kind of redundant with the
```
GabrielKS, line 489: with the what?

src/parameters/update_cost_parameters.jl
```diff
@@ -214,8 +222,13 @@ _index_into_param(cost_data, ::T) where {T <: Union{StartVariable, MultiStartVar
     start_up_cost(cost_data, T())
 _index_into_param(cost_data, ::VariableType) = cost_data
 
+get_update_multiplier(::DecrementalCostAtMinParameter) = -1.0
```
GabrielKS, line 225: Hmm. Why is this update-specific? Can we encode this detail in one place that gets used for both the initial population and the updating?

src/utils/time_series_utils.jl
```diff
@@ -47,6 +47,8 @@ _validate_eltype(
         )
     end
 
+# FIXME this is messing with test_device_load_constructors.jl:72
```
GabrielKS, line 50: TODO before merge?

test/test_utils/add_market_bid_cost.jl
```diff
@@ -0,0 +1,320 @@
...
+    end
+    add_mbc_inner!(sys, active_components; incr_curve = incr_curve, decr_curve = decr_curve)
+end
+
+"""
+Get a deterministic or DeterministicSingleTimeSeries time series from the system.
+"""
+function get_deterministic_ts(sys::PSY.System)
+    for device in get_components(PSY.Device, sys)
+        if has_time_series(device, Union{DeterministicSingleTimeSeries, Deterministic})
```
GabrielKS, line 71:
```suggestion
        if has_time_series(device, AbstractDeterministic)
```

test/test_utils/add_market_bid_cost.jl
```diff
@@ -0,0 +1,320 @@
...
+            for key in PSY.get_time_series_keys(device)
+                ts = get_time_series(device, key)
+                if ts isa DeterministicSingleTimeSeries || ts isa Deterministic
```
GabrielKS, line 74:
```suggestion
                if ts isa AbstractDeterministic
```

test/test_utils/mbc_simulation_utils.jl
```diff
@@ -0,0 +1,574 @@
...
+        InterruptiblePowerLoad => PowerLoadInterruption,
+        RenewableDispatch => RenewableFullDispatch,
+        # I include this file in the tests of SSS and HPS, which error on these formulations.
+        # HydroDispatch => HydroCommitmentRunOfRiver,
+        # EnergyReservoirStorage => StorageDispatchWithReserves,
+    )
+
+# debugging code for inspecting objective functions -- ignore
+# TODO LK group by folders.
+const DOWNLOADS = joinpath(homedir(), "Downloads")
```
GabrielKS, line 19: TODO remove before merge

test/test_utils/mbc_simulation_utils.jl
```diff
@@ -0,0 +1,574 @@
...
+            filename = filename,
+            is_decremental = is_decremental,
+            device_to_formulation = device_to_formulation,
+        )
+    end
+
+    # TODO make this more general as to which variables we're reading.
+    # e.g. hydro.
+
+    # TODO test slopes, breakpoints too once we are able to write those
```
GabrielKS, line 279: @luke-kiernan not the top priority but FYI we are able to write these now, so this can now be done

test/test_utils/mbc_system_utils.jl
```diff
@@ -0,0 +1,468 @@
...
+                therm_price_mult * get_slopes(pwl),
+            )
+            set_variable!(op_cost, CostCurve(new_pwl))
+        else
+            error("Unhandled operation cost variable type $(typeof(get_variable(op_cost)))")
+        end
+    end
+end
+
+tweak_for_startup_shutdown!(sys::System) = tweak_system!(sys::System, 0.8, 1.0, 1.0)
```
GabrielKS, line 129: Seems wise to put all the fiddly empirical numbers either (a) in one place that people know to look or (b) close to the tests that fail/error if they're wrong

GabrielKS (commented):
Miscellaneous small comments, not a full review yet.

test/test_utils/add_market_bid_cost.jl
GabrielKS: actually it looks like this function is unused, if so delete

test/test_utils/mbc_simulation_utils.jl
```diff
@@ -0,0 +1,574 @@
...
+    total1 =
+        [only(@combine(df, :total = sum(:value)).total) for df in values(ground_truth_1)]
+    total2 =
+        [only(@combine(df, :total = sum(:value)).total) for df in values(ground_truth_2)]
+    if !is_decremental
+        ground_truth_diff = total2 .- total1  # How much did the cost increase between simulation 1 and simulation 2 for each step
+    else
+        # objective = cost - benefit. higher load prices => more willing to pay, more benefit.
+        # so we get an extra negative sign, since we're increasing benefit, not cost.
+        ground_truth_diff = total1 .- total2
```
GabrielKS, line 430: Any reason you're doing this here rather than changing how the ground truth is calculated? `cost_due_to_time_varying_mbc` is already being passed `is_decremental`, it seems incorrect for it to not account for this there.

test/test_utils/mbc_simulation_utils.jl
```diff
@@ -0,0 +1,574 @@
...
+        vcat(values(read_variable_dict(res, var_name, gentype))...)
+        @rsubset(:name == unit_name)
+        @combine(:value = sum(:value))
+    end
+    return df[1, 1]
+end
+
+function build_generic_mbc_model(sys::System;
+    multistart::Bool = false,
+    standard::Bool = false,
```
GabrielKS, line 102: Pretty cryptic, here and elsewhere. `model_thermal_standard`? Same to a lesser extent for `multistart`.

test/test_utils/mbc_simulation_utils.jl
```diff
@@ -0,0 +1,574 @@
...
+                        TimeSeries.values(ii_comp),
+                    ),
+                )
+            end
+        end
+    end
+    # NOTE this could be rewritten nicely using PowerAnalytics
+    # Select component based on comp_type - fallback to legacy behavior if needed
+    sel = make_selector(T, comp_name)
+    @assert !isnothing(first(get_components(sel, sys)))
```
GabrielKS, line 314:
```suggestion
    @assert !isnothing(get_component(T, sys, comp_name))
```
big fan of `ComponentSelector`s but there's no point if you're only gonna use it once, immediately

test/test_utils/mbc_simulation_utils.jl
GabrielKS: -> https://github.com/NREL-Sienna/PowerSimulations.jl/issues/1429

jd-lara (commented):
I'll have to merge to make the HPS test pass too after I merged the MBC branch there.

test/test_utils/mbc_simulation_utils.jl
luke-kiernan: Sure. Let me know once you have an implementation for import/export and I'll add it elsewhere too

src/devices_models/devices/common/objective_function/market_bid.jl
luke-kiernan: Sure, I can remove those commented out `println`'s. Is there a plan to do something about these "invisible warning" messages? imo the very minimum would be a post-build summary ("there were [n] warnings, see [log file]")

src/devices_models/devices/common/objective_function/market_bid.jl
luke-kiernan: With the already existing `_include_min_gen_power_in_constraint`. It occurred to me that I could just use that one function and check for the presence of an `OnVariable`. I'll switch to that approach--it would've prevented a bug I spent most of Friday hunting down--and remove the comment.

test/test_utils/add_market_bid_cost.jl
luke-kiernan: Yeah I didn't realize the `get_forecast_{horizon/interval/etc}(sys)` functions existed when I wrote this.

test/test_utils/mbc_system_utils.jl
luke-kiernan: Are there other fiddyly empirical numbers besides there? I could move the startup/shutdown one to be before the relevant test, but the test for the other one is in the middle of a long `@testset` block.

test/test_utils/mbc_simulation_utils.jl
luke-kiernan: Hmm...I find things clearer as-is. `cost_due_to_time_varying_mbc` calculates exactly that: with decremental, it's the relationship between the MBC cost and the objective function that changes, not the cost itself.

src/parameters/update_cost_parameters.jl
luke-kiernan: In the initial objective function, the decremental vs incremental behavior is determined by the `is_decremental::Bool` argument to `add_pwl_term!` at `market_bid.jl:680`. That then acts as a selector for functions like `_get_pwl_data`, `get_offer_curves_maybe_decremental`.

In the update, the decremental vs incremental behavior happens across 2 layers: `handle_variable_cost_parameter` calls `update_variable_cost!`. There's `_get_parameter_field` in the first and `get_update_multiplier` in the second.

Ideas:
1. Switch to using `_get_parameter_field` inside `add_pwl_term!`: dispatch on parameter type not bool. [Very doable, would reduce code duplication, but maybe less readable than what we have right now.]
2. Incorporate the `get_update_multiplier` sign into the `parameter_multiplier` arrays. [Will require more care, to make sure that's ok with the other places the same parameter multiplier is used.]
3. Add a bool argument `is_decremental` to `update_variable_cost!`. [Cheeky way of fixing 2.]

PR #1427, opened 2025-11-01
Support time-variable `ImportExportCost`

GabrielKS:
Here we add time-variable incremental and decremental support to `Source`s with `ImportExportCost`s. This is very similar to `MarketBidCost`, but with the very convenient guarantee that both incremental and decremental cost curves start at (0 power, 0 cost).

Depends on https://github.com/NREL-Sienna/PowerSystems.jl/pull/1560
Branches off https://github.com/NREL-Sienna/PowerSimulations.jl/pull/1428, which branches off https://github.com/NREL-Sienna/PowerSimulations.jl/pull/1423

### DIFF
--- a/src/simulation/simulation_results.jl
+++ b/src/simulation/simulation_results.jl
@@ -249,7 +249,7 @@ function get_emulation_problem_results(
 end
 
 function _populate_system_in_results!(
-    results::SimulationResults,
+    results::SimulationProblemResults,
     populate_system::Bool,
     populate_units::Union{IS.UnitSystem, String, Nothing},
 )


src/devices_models/devices/common/objective_function/market_bid.jl
```diff
@@ -148,34 +156,104 @@ function validate_mbc_breakpoints_slopes(device::PSY.StaticInjection, decrementa
...
+            decremental,
+            is_ts,
+            x,
+            device_name,
+            p1,
+        )
+    end
+end
+
+function _do_specific_occ_validation(
```
luke-kiernan, line 183: Why not just call it `_occ_validation`? The `_do` (and to a lesser extent, the `specific`) seems redundant.

test/test_utils/mbc_simulation_utils.jl
```diff
@@ -2,9 +2,10 @@
 # If you make changes, run those tests too!
 const TIME1 = DateTime("2024-01-01T00:00:00")
 test_path = mktempdir()
+const FormulationDict = Dict{Type{<:PSY.Device}, Type{<:PSI.AbstractDeviceFormulation}}
```
luke-kiernan, line 5: Thanks! This type abbreviation was badly needed. (Not sure why I didn't think it...)

src/devices_models/devices/common/objective_function/market_bid.jl
```diff
@@ -282,42 +360,42 @@ function validate_mbc_component(
...
 
 # Slope and breakpoint validations are done together, nothing to do here
-validate_mbc_component(
+validate_occ_component(
     ::AbstractPiecewiseLinearSlopeParameter,
     device::PSY.StaticInjection,
 ) = nothing
 
-function _process_market_bid_parameters_helper(
+function _process_occ_parameters_helper(
```
rodrigomha, line 392: I think we can add a small comment here that explains that validate the data and finally add the parameters

src/parameters/update_cost_parameters.jl
```diff
@@ -71,15 +71,15 @@ _get_parameter_field(
...
+    get_input_offer_curves(args...; kwargs...)
 
 _maybe_tuple(::StartupCostParameter, value) = Tuple(value)
 _maybe_tuple(::ShutdownCostParameter, value) = value
 _maybe_tuple(::AbstractCostAtMinParameter, value) = value
 
 function handle_variable_cost_parameter(
     param::Union{StartupCostParameter, ShutdownCostParameter, AbstractCostAtMinParameter},
-    op_cost::PSY.MarketBidCost,
+    op_cost::OfferCurveCost,
```
rodrigomha: I'm guessing OfferCurveCost is the supertype of MarketBid and ImportExport right?

GabrielKS: yeah -> https://github.com/NREL-Sienna/PowerSystems.jl/pull/1568

src/devices_models/devices/common/objective_function/market_bid.jl
GabrielKS: Done

src/devices_models/devices/common/objective_function/market_bid.jl
GabrielKS: Some validations are already performed in the body of `validate_occ_breakpoints_slopes`, this is for the ones that are different depending on whether it's an `ImportExportCost` or a `MarketBidCost`. But I'll come up with a better name

PR #1433, opened 2025-11-04
load constructor test fixed

luke-kiernan:
Despite going through every place we `throw` something here in PSI, I couldn't figure out exactly which error was being tested. So I switched the test to match the behavior. edit: that's usually an iffy thing to do, but here, it feels fine. The important thing is that an error is thrown

### DIFF
--- a/src/utils/time_series_utils.jl
+++ b/src/utils/time_series_utils.jl
@@ -47,8 +47,6 @@ _validate_eltype(
         )
     end
 
-# FIXME this is messing with test_device_load_constructors.jl:72
-# the offer curves are nothing, so an error gets thrown here, when it expects a later error.
 function _validate_eltype(::Type{T}, component::PSY.Component, element, msg = "") where {T}
     component_name = get_name(component)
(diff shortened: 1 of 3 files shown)


src/utils/time_series_utils.jl
```diff
@@ -47,8 +47,6 @@ _validate_eltype(
         )
     end
 
-# FIXME this is messing with test_device_load_constructors.jl:72
-# the offer curves are nothing, so an error gets thrown here, when it expects a later error.
 function _validate_eltype(::Type{T}, component::PSY.Component, element, msg = "") where {T}
```
jd-lara, line 50: what's element?

luke-kiernan: Either the field itself [if the field that we're checking isn't a time series] or an instance in the time series [if it is one].

luke-kiernan: Could be a tuple [start up costs], a float, or a `CostCurve` instance.

jd-lara (approved):

PR #1455, opened 2025-11-13
Fix cost tracking and ramp constraints for thermal generators

taranraj123:
Fixes some bugs in thermal generator cost calculations and ramp constraint formulation.

The main issues were:
  - VOM costs weren't being scaled by time step duration, causing incorrect costs for sub-hourly resolutions (15-min, 30-min intervals)
  - Startup and shutdown costs were being calculated but never added to the ProductionCostExpression for tracking
  - The linear ramp constraints are tighter and use the UC commitments, if available. 

## Changes

**Cost tracking fixes**:

  - Multiply VOM costs by `dt = resolution / MILLISECONDS_IN_HOUR` to properly scale with time resolution
  - Uncommented `add_to_expression!` calls for startup costs and shutdown costs so they are accounted for in the ProductionCostExpression

**Ramp constraint improvements**:

The existing ramp constraints for thermal dispatch used ActivePowerRangeExpression, which includes a term Pmax * On[t]. This causes the constraints to become incorrectly relaxed when units are committed ON by the upstream unit commitment problem.

Example: Consider a unit with Pmax=1212 MW and ramp_limit=450 MW/period. When the unit is ON (On[t]=1), the existing constraint becomes:

P[t] - Pmax*On[t] - (P[t-1] - Pmax*On[t-1]) ≤ ramp_limit
→ P[t] - P[t-1] ≤ 450 + 1212 = 1662 MW

The constraint is supposed to limit ramp rate to 450 MW, but instead allows changes up to 1662 MW.

The Proposed Fix: The new implementation uses the commitment status (y[t] ∈ {0,1}) directly as a parameter with big-M formulation:

P[t] - P[t-1] ≤ ramp_limit + Pmax*(2 - y[t-1] - y[t])

When the unit stays ON (y[t-1]=1, y[t]=1): P[t] - P[t-1] ≤ 450 + 1212*0 = 450 MW

When the unit is OFF or transitioning: the big-M term relaxes the constraint appropriately.

## Testing

Added tests to validate:
  - VOM costs scale correctly between hourly and 30-minute resolutions
  - Startup/shutdown costs appear in ProductionCostExpression
  - Ramp constraints work properly with UC feedforward. Updated constraint count expectations in thermal generation tests to match the new ramp constraint structure.

Note: Commented out one pre-existing broken test in test_market_bid_cost.jl (multistart scenario) - this was already failing on main and is unrelated to these changes.


taranraj123:
> The changes to the cost look ok to me. I need to look into more details about the ramp change.

Thank you @jd-lara.

src/devices_models/devices/common/rateofchange_constraints.jl
```diff
@@ -219,6 +219,98 @@ function add_linear_ramp_constraints!(
...
+    end
+
+    time_steps = get_time_steps(container)
+    variable = get_variable(container, U(), V)
+    ramp_devices = _get_ramp_constraint_devices(container, devices)
+    minutes_per_period = _get_minutes_per_period(container)
+    IC = _get_initial_condition_type(T, V, W)
+    initial_conditions_power = get_initial_condition(container, IC(), V)
+    ic_power_by_name =
+        Dict(get_component_name(ic) => get_value(ic) for ic in initial_conditions_power)
```
jd-lara: You don't need the initial condition by name here adding another loop. You can index directly into the initial conditions container inside the loop in 268.

jd-lara (changes requested):
The PR logic looks ok to me but there is an unnecessary Dict creation in the logic to fix the ramp constraints.

src/devices_models/devices/common/rateofchange_constraints.jl
taranraj123: Done. Thanks for the review. Please let me know if there is anything else

Issue #944, opened 2023-03-14
Security constrained optimal power flow


hei06j:
**Is your feature request related to a problem? Please describe.**
I want to represent N-1 (N-k) security constrained optimal power flow (unit commitment, economic dispatch) with PTDF formulation against branch and generator contingencies. A two stage mathematical problem and/or decomposition techniques can be used to improve computation efficiency.

**Describe the solution you'd like**
The N-1 (N-K) secure network with preventive actions can tolerate the branch and generator contingencies considered in the contingency list.

Corrective actions such as re-dispatch, transformer OLTC, and line switching could also be added.

jd-lara:
closed via #1567

PR #1456, opened 2025-11-14
Sm/restore_scuc_psy5

SebastianManriqueM:
This PR restores DLRs, G-1 and N-1 models considering redesign of Network Reductions in PNM.

# ISSUES TO CLOSE
THis PR closses the following issues:
- https://github.com/NREL-Sienna/PowerOperationsModels.jl/issues/51
- https://github.com/NREL-Sienna/PowerSimulations.jl/issues/944

# Notes:

- G-1 models RampReserveWithDeliverabilityConstraints and ContingencyReserveWithDeliverabilityConstraints are compatible with PTDFPowerModel, AreaPTDF, AreaBalance and CopperPlate

- N-1 model was refactored as a branch formulation. This opens the oportunity for AC modeling of the network but considering post contingency in DC model.

- The network formulation for N-1 was removed.

- Legacy code from G-1 models as Generation formulations were removed.

- Dynamic Line ratings were restored.

# Pending:
- I'm still not sure how to handle DLRs with 3W transformers and monitored lines


src/PowerSimulations.jl
```diff
@@ -369,7 +367,7 @@ export ActivePowerInTimeSeriesParameter
 export ReactivePowerTimeSeriesParameter
 export DynamicBranchRatingTimeSeriesParameter
 export FuelCostParameter
-export PostContingencyDynamicBranchRatingTimeSeriesParameter
+export PostContingencyDynamicBranchRatingTimeSeriesParameter#TODO review idf this should remain
```
m-bossart: Just flagging this, if it is unused lets comment or remove for now

src/core/expressions.jl
```diff
@@ -23,6 +23,7 @@ struct PTDFBranchFlow <: ExpressionType end
 struct PostContingencyBranchFlow <: PostContingencyExpressions end
 struct PostContingencyActivePowerGeneration <: PostContingencyExpressions end
 struct PostContingencyNodalActivePowerDeployment <: PostContingencyExpressions end
+struct PostContingencyAreaActivePowerDeployment <: PostContingencyExpressions end
```
m-bossart, line 26: Why do we need a different expression for nodal and area active power deployment?

src/core/formulations.jl
```diff
@@ -146,6 +139,10 @@ Branch formulation for PhaseShiftingTransformer flow control
 """
 struct PhaseAngleControl <: AbstractBranchFormulation end
 
+abstract type AbstractSecurityConstrainedStaticBranch <: AbstractBranchFormulation end
+
+struct SecurityConstrainedStaticBranch <: AbstractSecurityConstrainedStaticBranch end
```
m-bossart, line 147: Add docstring following the other examples

src/core/formulations.jl
```diff
@@ -250,7 +247,13 @@ struct RangeReserve <: AbstractReservesFormulation end
 """
 Struct for to add reserves to be larger than a specified requirement and map how those should be allocated and deployed considering generators outages
 """
-struct RangeReserveWithDeliverabilityConstraints <:
+struct RampReserveWithDeliverabilityConstraints <:
+       AbstractSecurityConstrainedReservesFormulation end
+
+"""
+Struct for to add reserves implicit reserve requirement from generation outages and map how those should be allocated and deployed in each outage at each time-step
```
m-bossart, line 257: Typo: "Struct for to" -> "Struct to...". A few more cases of this above as well.

src/core/network_reductions.jl
```diff
@@ -139,3 +139,30 @@ function get_branch_argument_constraint_axis(
...
+            continue
+        end
+
+        name = get_branch_argument_constraint_axis(
+            net_reduction_data,
+            reduced_branch_tracker,
+            branch_type,
+            U,
+        )
+        push!(names, name)
```
m-bossart: I would suggest changing the variable names to be more descriptive in this function (I probably wrote this originally). Maybe "names" -> "all_names", "name" -> "names_by_type"

src/core/optimization_container.jl
```diff
@@ -666,12 +629,12 @@ function initialize_system_expressions!(
...
     system::PSY.System,
     bus_reduction_map::Dict{Int64, Set{Int64}},
-) where {T <: Union{AreaPTDFPowerModel, SecurityConstrainedAreaPTDFPowerModel}}
+) where {T <: AreaPTDFPowerModel}
     areas = get_available_components(network_model, PSY.Area, system)
     if isempty(areas)
         throw(
             IS.ConflictingInputsError(
-                "AreaPTDFPowerModel/SecurityConstrainedAreaPTDFPowerModel doesn't support systems with no Areas",
+                "AreaPTDFPowerModel doesn't support systems with no Areas",
```
m-bossart: "AreaPTDFPowerModel doesn't support systems without Areas"

src/devices_models/devices/AC_branches.jl
```diff
@@ -458,6 +458,139 @@ function get_min_max_limits(
...
+    ::Type{<:PostContingencyConstraintType},
+    ::Type{<:AbstractBranchFormulation},
+)
+    if PSY.get_flow_limits(device).to_from != PSY.get_flow_limits(device).from_to
+        @warn(
+            "Flow limits in Line $(PSY.get_name(device)) aren't equal. The minimum will be used in formulation $(T)"
+        )
+    end
+    limit = min(
+        PSY.get_rating(device),
```
m-bossart: Shouldn't this be `rating_b`?

src/devices_models/devices/AC_branches.jl
```diff
@@ -458,6 +458,139 @@ function get_min_max_limits(
...
+        PSY.get_flow_limits(device).from_to,
+    )
+    minmax = (min = -1 * limit, max = limit)
+    return minmax
+end
+
+"""
+Min and max limits for Abstract Branch Formulation
+"""
+function get_scuc_min_max_limits(
```
m-bossart: I think the name of this function should be changed because the functionality has nothing to do with UC. What about `get_emergency_min_max_limits`?

src/devices_models/devices/AC_branches.jl
```diff
@@ -458,6 +458,139 @@ function get_min_max_limits(
...
+    end
+    return (min = -1 * PSY.get_rating_b(device), max = PSY.get_rating_b(device))
+end
+
+function get_scuc_min_max_limits(
+    ::PSY.PhaseShiftingTransformer,
+    ::Type{PhaseAngleControlLimit},
+    ::Type{PhaseAngleControl},
+) #  -> Union{Nothing, NamedTuple{(:min, :max), Tuple{Float64, Float64}}}
+    return (min = -π / 2, max = π / 2)
```
m-bossart: We can just call `get_min_max_limits` here so it is obvious they are the same

src/devices_models/devices/static_injection_security_constrained_models.jl
```diff
@@ -1,355 +1,698 @@
...
-            base_name = "$(var_type)_$(D)_{$(name), $(o), $(t)}",
-        )
-        # TODO: Contingencies Implement method to get max change depending on the device and formulation
-        JuMP.set_upper_bound(variable[name, o, t], PSY.get_max_active_power(d))
-        JuMP.set_lower_bound(variable[name, o, t], 0.0)
-    end
+    for outage in associated_outages
+        outage_name = string(IS.get_uuid(outage))
+        associated_devices =
+            PSY.get_associated_components(sys, outage; component_type = PSY.Generator)
```
m-bossart, line 34: What about non-generator static injectors (e.g. Storage)?

m-bossart: Should this be more general? This appears in a bunch of places

src/network_models/security_constrained_models.jl
```diff
@@ -1,117 +1,379 @@
...
-    ::NetworkModel{<:AbstractPTDFModel},
-)
-    if PSY.get_rating_b(branch) === nothing
-        @warn "Branch $(get_name(branch)) has no 'rating_b' defined. Post-contingency limit is going to be set using normal-operation rating.
-            \n Consider including post-contingency limits using set_rating_b!()."
-        return (min = -1 * PSY.get_rating(branch), max = PSY.get_rating(branch))
-    end
-    return (min = -1 * PSY.get_rating_b(branch), max = PSY.get_rating_b(branch))
-end
+# TODO - error if a the outage is associated with a reduced branch in N-1 formulation
```
m-bossart: We should open an issue to deal with this case. This should be checked and handled automatically.

test/test_security_constrained_models.jl
```diff
@@ -362,11 +383,4 @@ end
             10000,
         )
     end
-    # SecurityConstrainedPTDF input Error testing
```
m-bossart, line 365: What was this testing for? Is there a reason to remove it?

test/test_static_injection_security_constrained_models.jl
```diff
@@ -98,12 +756,1170 @@
...
+        "ActivePowerTimeSeriesParameter__PowerLoad";
+        table_format = TableFormat.WIDE,
+    )
+    thermal_gen = read_variable(
+        results,
+        "ActivePowerVariable__ThermalStandard";
+        table_format = TableFormat.WIDE,
+    )
+
+    zone_1_load = sum(eachcol(load[!, ["Bus4_1", "Bus3_1", "Bus2_1"]]))
```
m-bossart, line 1884: Can we add add these checks related to the area balance to a helper function similar to `compare_outage_power_and_deployed_reserves`

test/test_static_injection_security_constrained_models.jl
```diff
@@ -1,21 +1,512 @@
...
+            res = OptimizationProblemResults(ps_model)
+            compare_outage_power_and_deployed_reserves(
+                sys,
+                res,
+                reserve_up)
+        end
+    end
+end
+
+@testset "Test if G-n with Ramp reserve deliverability constraints builds when there is a device without set_device_model!()" begin
```
m-bossart, line 363: Can you add a comment explaining why we need this test? It isn't clear to me

src/core/expressions.jl
SebastianManriqueM: PostContingencyNodalActivePowerDeployment is for PTDF and PostContingencyAreaActivePowerDeployment is for AreaBalance model. I was trying to be consistent with the architecture we have

src/core/formulations.jl
SebastianManriqueM: Done: """"
Branch formulation for PSY.ACTransmission Branches that allows to model security-constrained UC/ED by adding outage supplemental attributes to the branches.
""""

src/devices_models/devices/AC_branches.jl
SebastianManriqueM: I agreee and thanks for the suggestion!

src/devices_models/devices/AC_branches.jl
SebastianManriqueM: absolutely

src/devices_models/devices/AC_branches.jl
SebastianManriqueM: indeed

src/devices_models/devices/static_injection_security_constrained_models.jl
SebastianManriqueM: The Storage example is a good one. I think the easiest way would be making an abstract type with the devices type we support. In StaticInjection for example this formulation is not intended (yet) to support InterconnectingConverter for instance. I'll let this for now

src/network_models/security_constrained_models.jl
SebastianManriqueM: I added issue https://github.com/NREL-Sienna/PowerSimulations.jl/issues/1458, and deleted the todo

test/test_static_injection_security_constrained_models.jl
SebastianManriqueM: Got it!. We added this test because at the beginning we were iterating through all the devices types in the system. However, it does not mean that the user set the device model to all the branches. If the user does not set the device model, then it will result in an error because the variables for that branch type were not created (Yonghong uses this a lot)

jd-lara:
closing in favor of https://github.com/NREL-Sienna/PowerSimulations.jl/pull/1462 and https://github.com/NREL-Sienna/PowerSimulations.jl/pull/1463

PR #1475, opened 2025-12-10
fix should_export_dual call

luke-kiernan:
Function called is [here](https://github.com/NREL-Sienna/PowerSimulations.jl/blob/main/src/simulation/simulation_results_export.jl#L138): no type signature, but 2nd arg is called `tstamp`. Without it, I get an error about trying to call `isless(Int, DateTime)`.

### DIFF
--- a/src/core/store_common.jl
+++ b/src/core/store_common.jl
@@ -50,7 +50,7 @@ function write_model_dual_results!(
         write_result!(store, model_name, key, index, update_timestamp, data)
 
         if export_params !== nothing &&
-           should_export_dual(export_params[:exports], index, model_name, key)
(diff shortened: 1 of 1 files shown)


luke-kiernan:
One thing I'm puzzled about: why wasn't this caught by our test cases?

jd-lara (approved):

luke-kiernan:
Seems like we just need better testing all-around of the emulation model code. The thing that triggers it is [this line](https://github.com/NREL-Sienna/PowerSimulations.jl/blob/main/src/operation/emulation_model.jl#L634), where the index is of int type.

The same bug also occurs in several other functions on in this same file. Will open a PR fixing that and adding a test later today.

PR #1476, opened 2025-12-10
fix a silly typo

luke-kiernan:


GabrielKS (approved):
Sure, looks fine — troubling that the tests didn't catch this though….

Issue #1474, opened 2025-12-10
Exports with emulation model


luke-kiernan:
**Describe the bug**
If I take the basic setup of `test/performance/performance_test.jl` and add an export (in the same vein as `test/test_utils/run_simulation.jl`), then I get errors.

**To Reproduce**
Run the below using the environment inside the `test` folder.
<details>
  <summary>Full script</summary>

```julia
using PowerSimulations
const PSI = PowerSimulations
using PowerSystems
const PSY = PowerSystems
using Logging
using PowerSystemCaseBuilder
using PowerNetworkMatrices
using HydroPowerSimulations
using HiGHS
using Dates
using PowerFlows

function set_device_models!(template::ProblemTemplate, uc::Bool = true)
    if uc
        # unique to UC
        set_device_model!(template, ThermalMultiStart, ThermalStandardUnitCommitment)
        set_device_model!(template, ThermalStandard, ThermalStandardUnitCommitment)
        set_device_model!(template, HydroDispatch, FixedOutput)
    else
        # unique to ED
        set_device_model!(template, ThermalMultiStart, ThermalBasicDispatch)
        set_device_model!(template, ThermalStandard, ThermalBasicDispatch)
        set_device_model!(template, HydroDispatch, HydroDispatchRunOfRiver)
    end

    set_device_model!(template, RenewableDispatch, RenewableFullDispatch)
    set_device_model!(template, PowerLoad, StaticPowerLoad)
    set_device_model!(template, DeviceModel(Line, StaticBranch))
    set_device_model!(template, Transformer2W, StaticBranchUnbounded)
    set_device_model!(template, TapTransformer, StaticBranchUnbounded)
    set_service_model!(
        template,
        ServiceModel(VariableReserve{ReserveUp}, RangeReserve),
    )
    set_service_model!(
        template,
        ServiceModel(VariableReserve{ReserveDown}, RangeReserve),
    )
    return template
end

# based on test/test_utils/run_simulation.jl and test/performance/performance_test.jl
function run_rts(
    pf::Union{<:PowerFlows.PowerFlowEvaluationModel, Nothing},
    export_path::String
)
    sys_rts_da = build_system(PSISystems, "modified_RTS_GMLC_DA_sys")
    sys_rts_rt = build_system(PSISystems, "modified_RTS_GMLC_RT_sys")
    sys_rts_realization = build_system(PSISystems, "modified_RTS_GMLC_realization_sys")

[...]

PR #1480, opened 2025-12-11
Lk/emulation bugfixes

luke-kiernan:
More emulation related bugfixes. Adding test cases for these things is blocked by issue #1474


GabrielKS (approved):
LGTM, should test these when possible but the bugfixes are obvious enough that I'm okay approving without testing

PR #1481, opened 2025-12-11
clean up testing warnings

jd-lara:


luke-kiernan (commented):
When I run the tests locally, I get one failure in the "Test simulation partitions" tests, saying "UndefVarError: `fast_ipopt_optimizer` not defined in `Main`," followed by a comment about world age saying the binding might be too new. I'm used to seeing those messages due to `Revise` issues, but after closing the REPL and re-running, I got the same error again...not sure if that's a new error, or one that was present previously.

PR #1482, opened 2025-12-12
add slack to staticbranchbounds

jd-lara:


test/test_device_branch_constructors.jl
```diff
@@ -80,6 +78,26 @@ end
...
+    set_device_model!(template, DeviceModel(Line, StaticBranchBounds; use_slacks = true))
+    set_device_model!(
+        template,
+        DeviceModel(MonitoredLine, StaticBranchBounds; use_slacks = true),
+    )
+    model_m = DecisionModel(template, system; optimizer = HiGHS_optimizer)
+    @test build!(model_m; output_dir = mktempdir(; cleanup = true)) ==
+          PSI.ModelBuildStatus.BUILT
+
+    @test !check_variable_bounded(model_m, FlowActivePowerVariable, Line)
```
m-bossart: I'm not sure I understand this. If we use `StaticBranchBounds` with `use_slacks = true`, in the current implementation there is no bound on the flow variable. Why would the slacks be used if the flow variable is unbounded. How is this set up different than `StaticBranchUnbounded`?

m-bossart: I was expecting the variable to be bounded and the slacks to be there to prevent infeasibility.

PR #1483, opened 2025-12-12
manage pretty tables two versions

jd-lara:


jd-lara:
@copilot find the equivalent in Pretty Tables v3 to the keyword show_header which was supported in v2. Review the updating guidelines.

PR #1487, opened 2025-12-16
Events cleanup

m-bossart:
- Uses a dedicated system in case builder instead of making system modifications in the tests.
- Enables hydro outages 
Requires:

- https://github.com/NREL-Sienna/PowerSystemsTestData/pull/116
- https://github.com/NREL-Sienna/PowerSystemCaseBuilder.jl/pull/179

### DIFF
--- a/src/core/definitions.jl
+++ b/src/core/definitions.jl
@@ -161,6 +161,3 @@ Base.convert(::Type{SimulationBuildStatus}, val::String) =
 Base.convert(::Type{ModelBuildStatus}, val::String) = get_enum_value(ModelBuildStatus, val)
 Base.convert(::Type{RunStatus}, val::String) = get_enum_value(RunStatus, val)
 Base.convert(::Type{SOSStatusVariable}, x::String) = get_enum_value(SOSStatusVariable, x)
-
-const EVENTS_COMPATIBLE_INJECTORS =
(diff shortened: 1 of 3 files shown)


src/core/definitions.jl
```diff
@@ -162,6 +162,5 @@ Base.convert(::Type{ModelBuildStatus}, val::String) = get_enum_value(ModelBuildS
 Base.convert(::Type{RunStatus}, val::String) = get_enum_value(RunStatus, val)
 Base.convert(::Type{SOSStatusVariable}, x::String) = get_enum_value(SOSStatusVariable, x)
 
-#TODO - extend to Hydro once outages are made in HydroPowerSimulations
 const EVENTS_COMPATIBLE_INJECTORS =
```
jd-lara: Having a Union of these many types is not a good idea. We need to rethink the use of this constant

jd-lara: Change this to 


```suggestion
const EVENTS_COMPATIBLE_INJECTORS = Set{}
```

and in the code where you check do 

```Julia 
if !(dtype \in EVENTS_COMPATIBLE_INJECTORS)
            continue
        end
```

jd-lara (changes requested):
The use of a union isn't correct in that case.

jd-lara (approved):

PR #1492, opened 2025-12-16
low-hanging fruit for build _impl! precompilation time

luke-kiernan:
Results of spending an hour or so with `SnoopCompile.jl`: along with IS PR 523, these take the precompilation time for a super basic UC `build_impl!` from 3.1 s to 2.6 s.


luke-kiernan:
Huh what? The precompile time is the same or slightly worse. Weird.

Issue #1493, opened 2025-12-17
`RenewableDispatch` with quadratic cost curve


luke-kiernan:
**If this is a question, something isn't working or an idea please start a Q&A discussion in the [Discussion tab](https://github.com/NREL-Sienna/PowerSimulations.jl/discussions)**
Open a bug report only if you can provide the details below

**Describe the bug**
Quadratic cost curves on `RenewableDispatch` components are broken. The code [here](https://github.com/NREL-Sienna/PowerSimulations.jl/blob/main/src/devices_models/devices/common/objective_function/quadratic_curve.jl#L48) fails to take into account the fact that `RenewableDispatch` components don't have `get_active_power_limits`. Same issue at line 77 of the same file.

**To Reproduce**
```julia
using PowerSimulations
using PowerSystemCaseBuilder
using PowerSystems
const PSY = PowerSystems
using PowerNetworkMatrices
using TimeSeries
using HydroPowerSimulations
using HiGHS

sys_uc = build_system(PSITestSystems, "c_sys5_hy_uc"; add_single_time_series=true)
rd1 = first(get_components(RenewableDispatch, sys_uc))
quadratic_cost = PSY.CostCurve(PSY.QuadraticCurve(0.02, 1.5, 0.0))
rd1_cost = get_operation_cost(rd1)
set_variable!(rd1_cost, quadratic_cost)
transform_single_time_series!(
    sys_uc,
    Hour(1),  # horizon
    Hour(1),   # interval
);
ptdf = PTDF(sys_uc)

template_uc = ProblemTemplate(NetworkModel(PTDFPowerModel; use_slacks=true, duals=[CopperPlateBalanceConstraint]))

set_device_model!(template_uc, ThermalStandard, ThermalBasicUnitCommitment)
set_device_model!(template_uc, RenewableDispatch, RenewableFullDispatch)
set_device_model!(template_uc, HydroDispatch, HydroDispatchRunOfRiver)
set_device_model!(template_uc, PowerLoad, StaticPowerLoad)
set_device_model!(template_uc, Line, StaticBranch)
set_device_model!(template_uc, Transformer2W, StaticBranch)
 
solver_highs = optimizer_with_attributes(HiGHS.Optimizer, "mip_rel_gap" => 0.5)
 
problem_uc = DecisionModel(
    template_uc,
    sys_uc;
    optimizer=solver_highs,
    name="UC"
)

build!(problem_uc, output_dir=mktempdir())
```

[...]

PR #1494, opened 2025-12-17
fix issue 1493

luke-kiernan:
resolve issue #1493: replace `get_max_active_power` with `get_min_max_limits`, which then requires passing the device formulation `U` as well.

### DIFF
--- a/src/devices_models/devices/common/objective_function/quadratic_curve.jl
+++ b/src/devices_models/devices/common/objective_function/quadratic_curve.jl
@@ -41,11 +41,12 @@ end
 function _add_quadraticcurve_variable_cost!(
     container::OptimizationContainer,
     ::T,
+    ::U,
     component::PSY.Component,
     proportional_term_per_unit::Vector{Float64},
     quadratic_term_per_unit::Vector{Float64},
-) where {T <: VariableType}
-    lb, ub = PSY.get_active_power_limits(component)
+) where {T <: VariableType, U <: AbstractDeviceFormulation}
+    lb, ub = get_min_max_limits(component, ActivePowerVariableLimitsConstraint, U)
     for t in get_time_steps(container)
         _check_quadratic_monotonicity(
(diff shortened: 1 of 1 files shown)


rodrigomha:
I like this approach. A more repetitious solution would be to just make the method for quadratic curve for RenewableDispatch as:

```julia
function _add_quadraticcurve_variable_cost!(
    ::T,
    component::PSY.RenewableDispatch,
    proportional_term_per_unit::Vector{Float64},
    quadratic_term_per_unit::Vector{Float64},
) where {T <: VariableType}
    lb = 0.0
    ub = PSY.get_max_active_power(component)
    for t in get_time_steps(container)
        _check_quadratic_monotonicity(
            PSY.get_name(component),
            quadratic_term_per_unit[t],
            proportional_term_per_unit[t],
            lb,
            ub,
        )
        _add_quadraticcurve_variable_term_to_model!(
            container,
            T(),
            component,
            proportional_term_per_unit[t],
            quadratic_term_per_unit[t],
            t,
        )
    end
    return
end

function _add_quadraticcurve_variable_cost!(
    ::T,
    component::PSY.RenewableDispatch,
    proportional_term_per_unit::Float64,
    quadratic_term_per_unit::Float64,
) where {T <: VariableType}
    lb = 0.0
    ub = PSY.get_max_active_power(component)
    _check_quadratic_monotonicity(PSY.get_name(component),
        quadratic_term_per_unit,
        proportional_term_per_unit,
        lb,
        ub,
    )
    for t in get_time_steps(container)
        _add_quadraticcurve_variable_term_to_model!(
            container,
            T(),
            component,
            proportional_term_per_unit,
            quadratic_term_per_unit,
            t,
        )
    end
    return
end
```

@jd-lara Do you think we should expand the method to use the formulation and we can use the min_max methods?

jd-lara (approved):

PR #1499, opened 2025-12-22
perf annotations and small type inference fixes

luke-kiernan:


src/devices_models/devices/common/add_variable.jl
```diff
@@ -38,6 +38,7 @@ function add_variables!(
     contributing_devices::Union{Vector{V}, IS.FlattenIteratorWrapper{V}},
     formulation::AbstractReservesFormulation,
 ) where {T <: VariableType, U <: PSY.AbstractReserve, V <: PSY.Component}
+    # PERF: compilation hotspot. Switch to TSC.
```
GabrielKS, line 41: What is `TSC`?

src/devices_models/devices/common/objective_function/market_bid.jl
```diff
@@ -391,15 +391,20 @@ validate_occ_component(
...
 function _process_occ_parameters_helper(
     ::P,
     container::OptimizationContainer,
     model,
     devices,
 ) where {P <: ParameterType}
-    validate_occ_component.(Ref(P()), devices)
-    if _consider_parameter(P(), container, model)
-        ts_devices = filter(device -> _has_parameter_time_series(P(), device), devices)
+    param_instance = P()
```
GabrielKS, line 401: Are you sure it's not free to create singleton instances of `P`?

src/devices_models/devices/common/add_variable.jl
luke-kiernan: `TypeSortedCollection`: it's a container that allows for heterogeneous types, such that you can `map`, `mapreduce`, etc. over the container in a type-stable fashion.

src/devices_models/devices/common/objective_function/market_bid.jl
luke-kiernan: iirc somehow the problem was the `Ref` plus the broadcasting created type instability. However, I accepted CoPilot's suggestion without thorough testing: perhaps simply replacing `Ref(P())` [mutable] with `(P(), )` [immutable] would've been sufficient.

jd-lara (approved):

PR #1512, opened 2026-01-08
Sm/restore_dlrs

SebastianManriqueM:
This PR restore DLR capabilities.
Please merge first this branch https://github.com/NREL-Sienna/PowerSimulations.jl/tree/mb/remove-old-sc-code of PR https://github.com/NREL-Sienna/PowerSimulations.jl/pull/1509


src/devices_models/devices/AC_branches.jl
```diff
@@ -316,17 +316,197 @@ function get_min_max_limits(
...
+    if PSY.has_time_series(branch, ts_type, ts_name)
+        branch_dlr_params = get_parameter_column_refs(param_container, get_name(branch))
+        return branch_dlr_params[t] * mult[ci_name, t]
     end
-    return device_dlr_params
+
+    return PSY.get_rating(branch)
+end
+
+#TODO sm/further discuss this approach which will cause negligible capacity increasing when more than 2 circuits are in parallel
```
jd-lara: @SebastianManriqueM we need to open an issue about this one to track it but lets remove the the TODO

src/devices_models/devices/AC_branches.jl
```diff
@@ -316,17 +316,197 @@ function get_min_max_limits(
...
+"""
+    get_equivalent_dynamic_branch_rating(param_container::ParameterContainer, bp::PNM.BranchesParallel{<:PSY.ACTransmission}, ts_name::String, ts_type::DataType, t::Int, ci_name::String, mult)
+
+Calculate the total rating for branches in parallel that contain Dynamic Branch Rating Time Series.
+For parallel circuits, the rating is the sum of individual ratings divided by the number of circuits.
+This provides a conservative estimate that accounts for potential overestimation of total capacity.
+"""
+function get_equivalent_dynamic_branch_rating(
```
jd-lara, line 351: @m-bossart don't we have a method in PNM for this?

src/devices_models/devices/AC_branches.jl
```diff
@@ -316,17 +316,197 @@ function get_min_max_limits(
...
+    ) / length(bp.branches)
+end
+
+"""
+    get_equivalent_dynamic_branch_rating(param_container::ParameterContainer, bs::PNM.BranchesSeries, ts_name::String, ts_type::DataType, t::Int, ci_name::String, mult)
+
+Calculate the rating for branches in series that contain Dynamic Branch Rating Time Series.
+For series circuits, the rating is limited by the weakest link: Rating_total = min(Rating1, Rating2, ..., Ratingn)
+"""
+function get_equivalent_dynamic_branch_rating(
```
jd-lara, line 379: same here

test/test_utils/add_components_to_system.jl
```diff
@@ -0,0 +1,87 @@
...
+        g = PSY.get_g(ac_transmission),
+        services = PSY.get_services(ac_transmission),
+        ext = PSY.get_ext(ac_transmission))
+
+    #Set ratings the half so the case remains equivalent to the original
+    set_rating!(ac_transmission, rating_new)
+    add_component!(sys, ac_transmission_copy)
+end
+
+function add_equivalent_ac_transmission_with_parallel_circuits!(
```
jd-lara, line 29: @m-bossart same here, we have these as methods in PNM

src/devices_models/devices/AC_branches.jl
SebastianManriqueM: > @m-bossart don't we have a method in PNM for this?

@jd-lara: The current implementation of this requires pass 
``` 
ts_name::String,
ts_type::DataType
```

to get the dlr time series. It would be wierd for me pull the parameter containers of PSI from PNM, but open to discuss it @m-bossart

src/devices_models/devices/AC_branches.jl
SebastianManriqueM: TODO removed and issue added: https://github.com/NREL-Sienna/PowerSimulations.jl/issues/1529

src/devices_models/devices/AC_branches.jl
m-bossart: We do have `get_equivalent_rating(::BranchesParallel)` and `get_equivalent_rating(::BranchesSeries)` in PNM. One possibility is to get a single multiplier from the timeseries, and then call the PNM methods and return the product. We need to think about the edge cases here because the time series are associated with a branch and a rating is associated with the double circuit. My suggestion is the following:
1. If only one circuit has a DLR timeseries, associate that multiplier with the total rating.
2. If both circuits have DLR timeseries, confirm that they are the same multiplier and use with the total rating.
3. If two circuits have different DLR timeseries, throw an error.

test/test_utils/add_components_to_system.jl
m-bossart: This is a test utility function, I'm not sure we have something like this in PNM

src/devices_models/devices/AC_branches.jl
m-bossart: @jd-lara @SebastianManriqueM  this is directly relevant to the issues @juflorez is seeing

PR #1514, opened 2026-01-08
Support 3d variables in EmulationModel

daniel-thom:


GabrielKS (commented):
I recommend using the functionality in `src/utils/indexing.jl`, such as `assign_maybe_broadcast!`, rather than manually branching on number of dimensions all the time. That is how I handle all the 3D parameters stuff internally and it leads to lots less duplicate code and doesn't seem to be horrible performance-wise. My draft PR https://github.com/NREL-Sienna/PowerSimulations.jl/pull/1518 is an example of how to do that with results — that should probably be subsumed within this.

jd-lara:
This PR needs testing with the IEC curves.

jd-lara:
closed in favor of #1539

PR #1518, opened 2026-01-14
Fix straggling 3D results issue

GabrielKS:
Trying to write 3D parameter results in this particular place was failing:

Screenshot 2026-01-13 at 5 02 45 PM

I fixed it in the same way I've been handling this kind of thing elsewhere (which required a broadening in the typing of `assign_maybe_broadcast!` since `HDF5.Dataset` isn't an `AbstractArray` — probably that code gets inlined anyway). I'm not totally sure why this current simulation is triggering this but none of my tests did. Draft because I haven't written tests yet.

### DIFF
--- a/src/simulation/hdf_simulation_store.jl
+++ b/src/simulation/hdf_simulation_store.jl
@@ -1070,7 +1070,7 @@ function _write_dataset!(
     array::Vector{Float64},
     index::EmulationModelIndexType,
 )
-    dataset[index, :] = array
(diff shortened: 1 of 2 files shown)


GabrielKS:
@annacasavant has a simulation that needs something like this feature, @m-bossart @daniel-thom if this gets consolidated into https://github.com/NREL-Sienna/PowerSimulations.jl/pull/1514 it'd be good to verify that it supports her use case

PR #1536, opened 2026-02-12
Sm/restore_dlr_up

SebastianManriqueM:
This PR restore DLR capabilities.


src/devices_models/devices/utils_dlrs.jl
```diff
@@ -0,0 +1,192 @@
...
+Calculate the total rating for PSY.ACTransmission branches that contain Dynamic Branch Rating Time Series.
+"""
+function get_equivalent_dynamic_branch_rating(
+    param_container::ParameterContainer,
+    branch::U,
+    ts_name::String,
+    ts_type::DataType,
+    t::Int,
+    ci_name::String,
+    mult,
```
jd-lara, line 17: Why is mult untyped?

src/devices_models/devices/AC_branches.jl
```diff
@@ -380,13 +380,37 @@ function add_constraints!(
         slack_ub = get_variable(container, FlowActivePowerSlackUpperBound(), T)
         slack_lb = get_variable(container, FlowActivePowerSlackLowerBound(), T)
     end
+
+    has_dlr_ts =
```
jd-lara, line 384: The implementation of the parameter addition should go in the constructor. Follow the pattern from ThermalStandard 

```Julia
if haskey(get_time_series_names(device_model), DynamicBranchRatingTimeSeriesParameter)
        add_parameters!(container, ActivePowerTimeSeriesParameter, devices, device_model)
    end
```

src/devices_models/devices/AC_branches.jl
```diff
@@ -450,13 +474,37 @@ function add_constraints!(
...
```
jd-lara, line 478: same here

src/devices_models/devices/utils_dlrs.jl
jd-lara, line 1: This file should go in the utils folder and it is likely part of PSY utils or PNM utils

src/devices_models/devices/utils_dlrs.jl
```diff
@@ -0,0 +1,192 @@
...
+) where {U <: PSY.ACTransmission}
+    if PSY.has_time_series(branch, ts_type, ts_name)
```
jd-lara, line 19: You are doing this check every time. This function is likely unnecessary if you add the parameter for the time series first in the constructor and then just use the parameter in the constraint construction

src/devices_models/devices/utils_dlrs.jl
```diff
@@ -0,0 +1,192 @@
...
+function get_equivalent_dynamic_branch_rating(
+    param_container::ParameterContainer,
+    bp::PNM.BranchesParallel{<:PSY.ACTransmission},
+    ts_name::String,
+    ts_type::DataType,
+    t::Int,
+    ci_name::String,
+    mult,
+)
+    return sum(
```
jd-lara, line 43: Same here

src/devices_models/devices/AC_branches.jl
SebastianManriqueM: I think I get it... let me see

PR #1549, opened 2026-02-20
add default time series names for sources

luke-kiernan:


test/test_device_source_constructors.jl
jd-lara, line 1: why did you disable all the tests with time series?

luke-kiernan: For `Source` components, the defaults include both `max_active_power_out` and `max_active_power_in`: the old code throws an error whenever we have one but not the other.

Also, slightly confused: I haven't disabled any tests here.

PR #1550, opened 2026-02-20
test breakpoints and slopes in `run_mbc_sim`

luke-kiernan:
Close the PSI equivalent of POM issue [30](https://github.com/NREL-Sienna/PowerOperationsModels.jl/issues/30).

(I haven't ported these tests over to POM yet because much of it is multi-timestep. I'll fix it over there for single problems in a separate PR.)


test/test_utils/mbc_simulation_utils.jl
jd-lara, line 1: Why are there tests being called in the utils file?

luke-kiernan: Yeah, it's mildly awkward. The 3-way split of PSI that we're working on should enable us to write unit tests instead of these messy, complex integration tests, so imo there's bigger fish to fry.

PR #1559, opened 2026-03-06
DLRs

m-bossart:
This PR adds support for **Dynamic Line Ratings (DLR)** in DC-PF models using the PTDF and VirtualPTDF network formulations. DLR allows branch flow limits to vary over time through a time-series multiplier on the static thermal rating, enabling tighter (or looser) limits at each dispatch interval.

** IMPORTANT: 

- We need to make PF compatible with the released version of PNM.
- Merge First "Add and fix methods for handling DLRs" in PNM: https://github.com/NREL-Sienna/PowerNetworkMatrices.jl/pull/266.
---

# Overview of Changes

## Split `BranchReductionOptimizationTracker` into typed variable and parameter dicts

**File:** `src/core/network_reductions.jl`

**Why:** The original `BranchReductionOptimizationTracker` had a single `argument_dict` used to track both JuMP decision variables (branch flows) and parameters (DLR factors) for reduced/parallel branches. This caused type ambiguity when storing `JuMP.VariableRef` objects vs. `Float64` parameter values — the two have fundamentally different semantics and need different storage types.

**What changed:** The struct was split into two separate dictionaries:

```julia
mutable struct BranchReductionOptimizationTracker
    variable_dict::Dict{
        Type{<:ISOPT.VariableType},
        Dict{Tuple{Int,Int}, Vector{JuMP.VariableRef}},
    }
    parameter_dict::Dict{
        Type{<:ISOPT.ParameterType},
        Dict{Tuple{Int,Int}, Vector{Union{Float64, JuMP.VariableRef}}},
    }
    constraint_dict::...
    ...
end
```

- `variable_dict` tracks `JuMP.VariableRef` flow variables, keyed by `VariableType` and arc `(from_bus, to_bus)`.
- `parameter_dict` tracks DLR parameter values, keyed by `ParameterType` and arc. Values are `Float64` for standard solves and `JuMP.VariableRef` (JuMP parameters) for recurrent/rolling-horizon solves.

Two new typed lookup functions replace the old generic one:

- `search_for_reduced_branch_variable!(tracker, arc, T)` — looks up or registers a variable entry.
- `search_for_reduced_branch_parameter!(tracker, arc, T)` — looks up or registers a parameter entry.

Both return `(has_entry::Bool, tracker_vector)` where `has_entry = true` means another branch type already populated this arc (shared parallel arc), so the caller should reuse the stored value rather than creating a new one.

## Correct parameter sharing across parallel branches of **different** types

**Files:** `src/parameters/add_parameters.jl`, `src/core/parameters.jl`

**Why:** In a network reduction, a `BranchesParallel{PSY.ACTransmission}([line, monitored_line])` is registered under *both* `Line` and `MonitoredLine` in the `name_to_arc_map`. When `_add_time_series_parameters!` is called once for `Line` and once for `MonitoredLine`, without coordination both would try to create independent JuMP parameter objects for the same arc's DLR time series — resulting in duplicated parameters pointing to the same underlying data.

**What changed:** `_add_time_series_parameters!` for `ACTransmission + AbstractPTDFModel` now calls `search_for_reduced_branch_parameter!` before creating any parameter:

- **First branch type to process the arc** (`has_entry = false`): Creates the JuMP parameter via `set_parameter!`, then stores the resulting `JuMP.VariableRef` (or `Float64` for non-recurrent builds) in the tracker.
- **Subsequent branch types for the same arc** (`has_entry = true`): Reads the already-stored value from the tracker and calls `set_parameter!` with that existing `VariableRef` or `Float64` — no new parameter is created.

A new overload of `set_parameter!` was added to `src/core/parameters.jl` to support passing an already-created `JuMP.VariableRef` directly (bypassing the `add_jump_parameter` call that would create a duplicate):

```julia
function set_parameter!(
    container::ParameterContainer,
    ::JuMP.Model,
    parameter::JuMP.VariableRef,
    ixs...,
)
    assign_maybe_broadcast!(get_parameter_array(container), parameter, ixs)
    return
end
```

---

## Deterministic build order: DLR-models process arcs before static models

**File:** `src/core/optimization_container.jl` (`build_impl!`)

**Why:** `BranchModelContainer` is a `Dict{Symbol, DeviceModelForBranches}`, which has non-deterministic iteration order in Julia. In the mixed `BranchesParallel{Line, MonitoredLine}` case, the arc belongs to both types' maps. The `get_branch_argument_constraint_axis` function uses a shared `arc_tuples_with_constraints::Set` to prevent multiple branch types from each creating a constraint for the same arc. Whichever type is processed **first** claims the arc.

If `MonitoredLine` (configured without DLR) happened to be processed before `Line` (configured with DLR) due to dict ordering, it would claim the arc with a static `FlowRateConstraint` — and then when `Line`'s DLR-aware `add_flow_rate_constraint_with_parameters!` ran, the arc was already taken and no DLR constraint was ever built. This was a **silent correctness bug**: the model solved without error, but ignored the DLR time series entirely.

**What changed:** Both the `ArgumentConstructStage` and `ModelConstructStage` loops in `build_impl!` now iterate a sorted list of branch models, with DLR-enabled models (those with non-empty `get_time_series_names(b)`) sorted first:

```julia
sorted_branch_models = sort(
    collect(values(template.branches));
    by = b -> isempty(get_time_series_names(b)) ? 1 : 0,
)
# Both loops use sorted_branch_models instead of values(template.branches)
```

This guarantees DLR-aware models always claim shared arcs before static models, regardless of dict hash order.

---

### DIFF
--- a/src/core/network_model.jl
+++ b/src/core/network_model.jl
@@ -270,6 +270,9 @@ function instantiate_network_model!(
         )
     elseif model.reduce_radial_branches
         @info "Applying radial reduction"
+        if !isempty(irreducible_buses)
+            @warn "Irreducible buses identified due to DLRs. The reduction of any radial branch between 2 irreducible buses wil be ignored"
+        end
         ybus =
             PNM.Ybus(
                 sys;
@@ -365,6 +368,9 @@ function instantiate_network_model!(
             )
         elseif model.reduce_radial_branches
             @info "Applying radial reduction"
+            if !isempty(irreducible_buses)
+                @warn "Irreducible buses identified due to DLRs. The reduction of any radial branch between 2 irreducible buses wil be ignored"
+            end
             ptdf = PNM.VirtualPTDF(
                 sys;
                 network_reductions = PNM.NetworkReduction[PNM.RadialReduction(;


src/core/network_model.jl
```diff
@@ -217,22 +249,27 @@ function instantiate_network_model!(
     end
     if model.reduce_radial_branches && model.reduce_degree_two_branches
         @info "Applying both radial and degree two reductions"
+        irreducible_buses = _get_irreducible_buses_due_to_dlrs(sys, model, branch_models)
         ybus = PNM.Ybus(
             sys;
             network_reductions = PNM.NetworkReduction[
-                PNM.RadialReduction(),
-                PNM.DegreeTwoReduction(),
+                PNM.RadialReduction(; irreducible_buses = collect(irreducible_buses)),
```
m-bossart: Don't need to add irreducible buses to radial reduciton, only degree two

src/core/network_reductions.jl
```diff
@@ -1,7 +1,8 @@
 mutable struct BranchReductionOptimizationTracker
     variable_dict::Dict{
-        Type{<:ISOPT.VariableType},
-        Dict{Tuple{Int, Int}, Vector{JuMP.VariableRef}},
+        Type{<:Union{ISOPT.VariableType, ISOPT.ParameterType}},
+        Any,  # VariableType entries: Dict{Tuple{Int,Int}, Vector{VariableRef}}
```
m-bossart: The parameters should not be in the tracker as a Float. They need to be a parameter. There might be an issue somewhere else

src/core/network_reductions.jl
```diff
@@ -87,13 +132,13 @@ function get_branch_argument_variable_axis(
     return collect(keys(name_axis))
 end
 
-function get_branch_argument_variable_axis(
+#= function get_branch_argument_variable_axis(
```
m-bossart, line 186: I think this function is unneeded which is why I commented it. We should delete it though

src/parameters/add_parameters.jl
```diff
@@ -208,6 +226,91 @@ end
...
+        end
+        device_with_time_series = PNM.get_device_with_time_series(reduction_entry, ts_type, ts_name)
+        ts_uuid = string(IS.get_time_series_uuid(ts_type, device_with_time_series, ts_name))
+
+        has_entry, tracker_container = search_for_reduced_branch_argument!(
+            reduced_branch_tracker,
+            arc,
+            T,
+        )
+        @show typeof(tracker_container)
```
m-bossart: Just flagging to remove after debugging

test/test_network_constructors_with_dlr.jl
```diff
@@ -0,0 +1,326 @@
...
+        for (ix, add_parallel_line_name) in enumerate(parallel_lines_names_to_add)
+            sys = PSB.build_system(PSITestSystems, "c_sys5")
+            line_to_add_parallel = get_component(Line, sys, add_parallel_line_name)
+            add_equivalent_ac_transmission_with_parallel_circuits!(
+                sys,
+                line_to_add_parallel,
+                PSY.Line,
+            )
+
+            add_dlr_to_system_branches!(
```
m-bossart: Not defined

src/core/network_model.jl
```diff
@@ -205,6 +205,38 @@ function _get_filters(branch_models::BranchModelContainer)
...
 
+function _get_irreducible_buses_due_to_dlrs(
+    sys::PSY.System,
+    network_model::NetworkModel,
+    branch_models::BranchModelContainer,
+)
+    @debug "Identifying buses that are irreducible due to dynamic line ratings"
+    irreducible_buses = Set{Int64}()
+    for branch_type in network_model.modeled_ac_branch_types
+        device_model = branch_models[Symbol(branch_type)]
```
jd-lara, line 216: we should break out this function into a separate function that dispatches into the branch type

jd-lara: This can be more effective if we check that the models have the time series attribute and skip those that don't have it

src/core/network_model.jl
```diff
@@ -205,69 +205,41 @@ function _get_filters(branch_models::BranchModelContainer)
...
             ],
         )
     elseif model.reduce_radial_branches
         @info "Applying radial reduction"
         ybus =
-            PNM.Ybus(sys; network_reductions = PNM.NetworkReduction[PNM.RadialReduction()])
+            PNM.Ybus(
+                sys;
+                network_reductions = PNM.NetworkReduction[PNM.RadialReduction(;
+                    irreducible_buses = irreducible_buses,
```
m-bossart, line 280: We should remove passing irreducible buses to the radial reduction. It is ok to reduce a DLR if it is radial.

src/core/network_model.jl
```diff
@@ -205,69 +205,41 @@ function _get_filters(branch_models::BranchModelContainer)
...
     end
     if model.reduce_radial_branches && model.reduce_degree_two_branches
         @info "Applying both radial and degree two reductions"
-        irreducible_buses = _get_irreducible_buses_due_to_dlrs(sys, model, branch_models)
         ybus = PNM.Ybus(
             sys;
             network_reductions = PNM.NetworkReduction[
-                PNM.RadialReduction(; irreducible_buses = collect(irreducible_buses)),
-                PNM.DegreeTwoReduction(; irreducible_buses = collect(irreducible_buses)),
+                PNM.RadialReduction(; irreducible_buses = irreducible_buses),
```
m-bossart, line 267: same here

src/core/network_model.jl
```diff
@@ -323,37 +297,41 @@ function instantiate_network_model!(
...
+
         if model.reduce_radial_branches && model.reduce_degree_two_branches
             @info "Applying both radial and degree two reductions"
-            irreducible_buses =
-                _get_irreducible_buses_due_to_dlrs(sys, model, branch_models)
             ptdf = PNM.VirtualPTDF(
                 sys;
                 network_reductions = PNM.NetworkReduction[
-                    PNM.RadialReduction(; irreducible_buses = collect(irreducible_buses)),
+                    PNM.RadialReduction(; irreducible_buses = irreducible_buses),
```
m-bossart, line 363: same here

src/core/network_model.jl
```diff
@@ -323,37 +297,41 @@ function instantiate_network_model!(
...
                     ),
                 ],
             )
         elseif model.reduce_radial_branches
             @info "Applying radial reduction"
             ptdf = PNM.VirtualPTDF(
                 sys;
-                network_reductions = PNM.NetworkReduction[PNM.RadialReduction()],
+                network_reductions = PNM.NetworkReduction[PNM.RadialReduction(;
+                    irreducible_buses = irreducible_buses,
```
m-bossart, line 377: same here

src/operation/operation_model_interface.jl
```diff
@@ -377,16 +377,61 @@ function serialize_optimization_model(model::OperationModel)
...
+    end
+    return collect(irreducible_buses)
+end
+
 function instantiate_network_model!(model::OperationModel)
     template = get_template(model)
     network_model = get_network_model(template)
     branch_models = get_branch_models(template)
     number_of_steps = get_time_steps(get_optimization_container(model))[end]
+    irreducible_buses_dlrs = _get_irreducible_buses_due_to_dlrs(
```
m-bossart: Why not do this inside of the inner `instantiate_network_model!`. With the current implementation, we go through these loops to determine irreducible buses even if we will never need that information (e.g. if the network formulation is Copper plate).

src/core/network_model.jl
SebastianManriqueM: > We should remove passing irreducible buses to the radial reduction. It is ok to reduce a DLR if it is radial.

I don't think this affect too much the purpose of the reductions.  but I think it has possible use cases:
I'm thinking in a case where you are trying to reduce unserved energy of a load connected through a radial line by using DLRs, or reduce curtailment of renewables.

src/operation/operation_model_interface.jl
SebastianManriqueM: > Why not do this inside of the inner `instantiate_network_model!`. With the current implementation, we go through these loops to determine irreducible buses even if we will never need that information (e.g. if the network formulation is Copper plate).

thats true

src/core/network_model.jl
m-bossart: When you use radial reductions, the user is saying that they don't want to model radial lines. This is inherently contradictory to the idea of trying to model a DLR on a radial line to reduce unserved energy of a load. I do think the user should be responsible in this case to not do both simultaneously.

SebastianManriqueM: My point is that we can reduce all the other radial lines, but no the radial(s) lines that has DLRs. So I don't think that is contradictory

jd-lara: The behavior should be a warning that a line with a DLR has been reduced in the radial reduction but prioritize the reduction.

jarry7: Physical significance should likely take precedence over software intent/performance.  

The DLR on a radial is deemed important by the engineer/researcher because he specifically asked for it to be modelled with a DLR.  In other words - constraints on monitored elements should be preserved in any network equivalent to maintain the validity of the constrained optimization.

The engineer's/researcher's intent may be contradictory if we assume that they want the entire network reduced (vs the subset that exclude when DLRs are modelled on those radial branches).  I think that is a bad assumption.

In this case, network reduction may still be desired on other radials where DLRs do not exist.  Where DLRs do exist on radial branches (load pockets, remote generation interconnection), we warn that a radial line has a DLR on it (and will be kept as a DLR branch).  The rest of the network reduction continues as expected.

jd-lara: The problem with these types of situation is making implicit assumptions about the intent of the user and allowing silent behavior that can be detrimental to performance. If the preference is to do keep a radial line with a DLR for whatever reason then it should be added to the exceptions explicitly or throw an error instead of a warning. 

We have seen issues like this in the case of adding a DLR on a single line to a double circuit.

jarry7: I don't think we are assuming anything.  

The user was very clear on the need for (1) DLR on the radial branch and (2) Radial network reductions.  So I think we should implement (1) on the branches requested and (2) on the subset of the network's radial branches.

Otherwise, it sounds like we are making the subjective and silent decision (not the user) by reducing every radial and giving a warning on not implementing DLRs vs implementing the DLRs on branches requested and reducing a subset of radials that exclude DLRs.

jd-lara: The comments here show that the suggestion is to warn (not silent) or throw error.

SebastianManriqueM: Ok, I'm understanding that I should prioritize reductions and warn if there is a line with DLR that was reduced. Is that correct? @jd-lara

jd-lara: we need to define adequately the behavior, we haven't defined the desired outcome as per the chat with @jarry7

SebastianManriqueM: Sounds good. When we have a consensus, just let me know and I'll be happy to implement the missing parts. :)

jarry7: The desired outcome (my subjective view):
- [ ] Represent DLRs as requested by the user (incl. on radials)
- [ ] Reduce network on all radials excl. those where DLRs were requested
- [ ] Warn the user that radial reductions will not happen on the DLR branch(es)

jd-lara: In order to move forward let's implement the behavior that @jarry7 is asking for. I hope we don't come to regret it

SebastianManriqueM: I think that`s the current implementation

PR #1564, opened 2026-03-19
Mb/cost expression refactor

### LINKED ISSUE #1534: Refactor ProductionCostExpression
Refactor the `ProductionCostExpression` in PSI which currently includes all of the costs aggregated into a single expression. The goal is to create new expressions which represent the constituent types of costs (e.g. FuelCost, VOMCost, StartupCost, ShutdownCost)

m-bossart:


src/devices_models/device_constructors/thermalgeneration_constructor.jl
```diff
@@ -104,6 +104,11 @@ function construct_device!(
 
     add_expressions!(container, FuelConsumptionExpression, devices, device_model)
     add_expressions!(container, ProductionCostExpression, devices, device_model)
+    add_expressions!(container, FuelCostExpression, devices, device_model)
```
jd-lara: I am confused about this design. It forces us to make a new loop over the components for every expression type

test/test_device_thermal_generation_constructors.jl
```diff
@@ -117,6 +117,163 @@ const TIME1 = DateTime("2024-01-01T00:00:00")
         # because it includes the startup cost in addition to generation costs
         @test cost_with_startup > cost_no_startup
     end
+
+   @testset "Test cost expression decomposition and nonnegativity (ThermalBasicUnitCommitment)" begin
+        for (i, cost_reference, thermal_formulation) in test_cases 
+            if thermal_formulation == ThermalBasicUnitCommitment
```
m-bossart: Why is this check here? I think it is always true.

test/test_device_thermal_generation_constructors.jl
```diff
@@ -117,6 +117,163 @@ const TIME1 = DateTime("2024-01-01T00:00:00")
         # because it includes the startup cost in addition to generation costs
         @test cost_with_startup > cost_no_startup
     end
+
+   @testset "Test cost expression decomposition and nonnegativity (ThermalBasicUnitCommitment)" begin
```
m-bossart: Related to co-pilot's suggestion of not re-solving all of the test cases. Lets just include these checks in the existing test above that already loops trough all of the test cases.

src/devices_models/device_constructors/thermalgeneration_constructor.jl
m-bossart: What is the alternative? This is already the case for most of the objective function code right? Where we add components individually in separate loops.

m-bossart: Should `ProductionCostExpression` be built as the sum of the component expressions? Instead of adding the same terms to both `ProductionCostExpression` and the constituent expression (e.g. `FuelCostExpression`, `VOMCostExpression`, etc.)

m-bossart:
@ChenHaotianC Now that this branch is up to date with main, can you please re-run the benchmarking on EI for this branch and for main, including multiple builds to get more consistent timing.

ChenHaotianC:
> @ChenHaotianC Now that this branch is up to date with main, can you please re-run the benchmarking on EI for this branch and for main, including multiple builds to get more consistent timing.

@m-bossart I ran the tests on EI for this branch and for main, using 5 build! runs. Based on the EI test, the build times for this branch are about the same as main.

Below is the table with the 5 build times.

```
time of   build! | main     | Mb/cost   expression refactor
               1 | 487.8348 | 436.0467
               2 | 422.3756 | 429.9026
               3 | 457.9066 | 428.5894
               4 | 454.0667 | 447.0854
               5 | 437.3533 | 473.0421
         average | 451.9074 | 442.93324
```

m-bossart:
I think the main outstanding issue here is the handling of the curtailment costs. Essentially, should curtailment cost be a constituent cost of production cost?

haotianc3:
> I think the main outstanding issue here is the handling of the curtailment costs. Essentially, should curtailment cost be a constituent cost of production cost?

I think curtailment cost shouldn't be a constituent of ProductionCostExpression.
A few reasons:
1. In how other PCM tools define it (PLEXOS, GenX, PyPSA, Dispa-SET), "production cost" means real actual costs — fuel, VOM, startup, no-load. Curtailment is either a modeling penalty or an opportunity cost, not money that actually changes hands. Rolling it into ProductionCostExpression would make it not directly comparable to those tools.
2. The existing -p^re term in RenewableFullDispatch already pushes against curtailment on its own. If CurtailmentCostExpression also gets folded into ProductionCostExpression, we'd end up with a negative "production cost" for renewables, which is going to confuse anyone reading the output.

ChenHaotianC (approved):
The testing/docstring updates, abstraction cleanup, and fuel-curve handling all look good. No further comments from me.

m-bossart:
This looks good to me also

jd-lara (approved):

PR #1567, opened 2026-03-20
SCUC N-1 Implementation

SebastianManriqueM:
# Summary
This PR introduces and refines the N-1 security-constrained branch flow construction for PTDF-based network models.

**This has branch has merged the dlr branch into main**

# What was implemented
- Added post-contingency branch flow expression construction for security-constrained AC transmission devices under PTDF formulations.
- Refactored the post-contingency flow builder into two logical stages:
  - Outer function handles outage-level preparation and validation.
  - Inner helper handles branch-type iteration and expression assembly (including threaded task execution).

- Reordered the main execution flow to be outage-first, which avoids repeating outage-specific preprocessing work for every monitored branch type.
- Added explicit validation of outage arc availability in the LODF outage axis before indexing.
- Added a clear error when an outage arc is not found in LODF, with guidance to:
remove the outage, or
mark outage arc buses as irreducible in the network model.

@jd-lara @m-bossart. I think making the outages work even for branches that were reduced seems to need some other implementations in the reduction mapings

### DIFF
--- a/src/devices_models/device_constructors/thermalgeneration_constructor.jl
+++ b/src/devices_models/device_constructors/thermalgeneration_constructor.jl
@@ -1,3 +1,13 @@
+# TODO: Security constrained models implement the correct functions for the model
+function has_security_arguments(device_model::DeviceModel)::Bool
+    return true
+end
+
+# TODO: Security constrained models implement the correct functions for the model
+function has_security_model(device_model::DeviceModel)::Bool
+    return true
+end
+
 @inline function _handle_common_thermal_parameters!(
     container::OptimizationContainer,
     devices::IS.FlattenIteratorWrapper{T},


m-bossart:
Similar to DLRs, we need to agree upon the desired interaction of the N-1 formulation and radial reductions. I would assume we want the same general behavior that we agreed upon for DLRs--- i.e. if an outage is to be modeled on a radial branch and radial reductions are enabled, the branch with the outage should be considered an exception and not reduced.

Unlike for DLRs, we should be able to still have degree two reductions if one of the branches is outaged, and we consider the outage to impact the full BranchesSeries reduction object.

m-bossart:
> # Summary
> This PR introduces and refines the N-1 security-constrained branch flow construction for PTDF-based network models.
> 
> **This has branch has merged the dlr branch into main**
> 
> # What was implemented
> * Added post-contingency branch flow expression construction for security-constrained AC transmission devices under PTDF formulations.
> * Refactored the post-contingency flow builder into two logical stages:
>   
>   * Outer function handles outage-level preparation and validation.
>   * Inner helper handles branch-type iteration and expression assembly (including threaded task execution).
> * Reordered the main execution flow to be outage-first, which avoids repeating outage-specific preprocessing work for every monitored branch type.
> * Added explicit validation of outage arc availability in the LODF outage axis before indexing.
> * Added a clear error when an outage arc is not found in LODF, with guidance to:
>   remove the outage, or
>   mark outage arc buses as irreducible in the network model.
> 
> @jd-lara @m-bossart. I think making the outages work even for branches that were reduced seems to need some other implementations in the reduction mapings

Can you explain what additional mapping you need?

jd-lara:
@SebastianManriqueM and @m-bossart to generalize the N-1 and N-x correctly we need to use this https://github.com/NREL-Sienna/PowerNetworkMatrices.jl/pull/270 that calculates the flow gates

PR #1568, opened 2026-03-22
performance updates 

jd-lara:
This is a big PR with some attempts to get performance improvements and avoid loops with some consolidation. Proudly this is a 100% hand made PR

### DIFF
--- a/src/operation/decision_model.jl
+++ b/src/operation/decision_model.jl
@@ -586,10 +586,10 @@ function handle_initial_conditions!(model::DecisionModel{<:DecisionProblem})
 end
 
 function _make_device_cache(
+    filter_function::Function,
     devices::IS.FlattenIteratorWrapper{T},
     check_components::Bool,
     sys::PSY.System,
-    filter_function::Function,
 ) where {T <: PSY.Device}
     device_cache = sizehint!(Vector{T}(), length(devices))
     for device in devices
@@ -602,10 +602,10 @@ function _make_device_cache(
 end
 
 function _make_device_cache(
+    ::Nothing,
     devices::IS.FlattenIteratorWrapper{T},
     check_components::Bool,
     sys::PSY.System,
-    ::Nothing,
 ) where {T <: PSY.Device}
     device_cache = sizehint!(Vector{T}(), length(devices))
     for device in devices
@@ -627,6 +627,6 @@ function make_device_cache!(
     devices = PSY.get_components(T, system; subsystem_name = subsystem)
     filt_func = get_attribute(model, "filter_function")
     model.device_cache =
-        _make_device_cache(devices, check_components, system, filt_func)
+        _make_device_cache(filt_func, devices, check_components, system)
     return
 end


.claude/claude.md
```diff
@@ -0,0 +1,146 @@
...
+## Core Architecture
+
+### Operation Models
+
+The central abstraction is `OperationModel`, with two concrete types:
+
+  - **`DecisionModel{M <: DecisionProblem}`** — Solves optimization problems over a specified horizon (e.g., 24h unit commitment, 1h economic dispatch). Contains a `ProblemTemplate`, an `OptimizationContainer` (JuMP model wrapper), and a `System` from PowerSystems.jl.
+  - **`EmulationModel{M <: EmulationProblem}`** — Simulates real-time operation with a single time-step horizon. Used for AGC, reserve deployment, and similar fast-timescale problems.
+
+Built-in problem types: `GenericOpProblem`, `UnitCommitmentProblem`, `EconomicDispatchProblem`, `AGCReserveDeployment`.
```
m-bossart, line 16: AGCReserveDeployment not included yet?

.claude/claude.md
```diff
@@ -0,0 +1,146 @@
...
+set_device_model!(template, ThermalStandard, ThermalBasicUnitCommitment)
+set_device_model!(template, RenewableDispatch, RenewableFullDispatch)
+set_service_model!(template, VariableReserve{ReserveUp}, RangeReserve)
+```
+
+### Device, Service, and Network Models
+
+These types bind a PowerSystems component type to a formulation:
+
+  - **`DeviceModel{D <: PSY.Device, B <: AbstractDeviceFormulation}`** — Specifies how a device type is modeled. The formulation determines which variables, constraints, and parameters are added. Also carries feedforward specifications, time series mappings, and attributes.
```
m-bossart, line 33: The feedforwards are specified outside of the `DeviceModel`?

src/operation/operation_model_serialization.jl
```diff
@@ -38,15 +37,6 @@ end
 function serialize_problem(model::OperationModel; optimizer = nothing)
     # A PowerSystem cannot be serialized in this format because of how it stores
```
m-bossart: Is this comment still relevant if we are no longer serializing the system?

src/operation/template_validation.jl
```diff
@@ -0,0 +1,108 @@
...
+    model_has_branch_filters = false
+    branch_keys_to_delete = Symbol[]
+    for (k, device_model) in model.template.branches
+        make_device_cache!(device_model, system, get_check_components(settings))
+        if isempty(get_device_cache(device_model))
+            @info "The system data doesn't include Branches of type $(k), consider changing the models in the template" _group =
+                LOG_GROUP_MODELS_VALIDATION
+            push!(branch_keys_to_delete, k)
+        else
+            push!(network_model.modeled_ac_branch_types, get_component_type(device_model))
```
m-bossart, line 97: `modeled_ac_branch_types` should be only those branches that are modeled and <: ACTransmission. It seems like in this logic two terminal HVDC would also be included which could cause failures downstream.

test/test_utils/mock_operation_models.jl
```diff
@@ -133,6 +133,7 @@ function mock_construct_device!(
     template = PSI.get_template(problem)
     PSI.finalize_template!(template, PSI.get_system(problem))
     PSI.validate_time_series!(problem)
+    #PSI.validate_template(problem)
```
m-bossart, line 136: Should be uncommented for validation to be run in test?

m-bossart (commented):
What is the logic for removing the `system_to_file` as a user defined option?

.claude/claude.md
jd-lara: we probably want to remove that from here and it is likely the implementation will need to be finalized in POM

test/test_utils/mock_operation_models.jl
jd-lara: It fails on the mocks but still runs in the DeviceModel tests.

src/operation/template_validation.jl
jd-lara: True, I need to make this correction.

src/operation/operation_model_serialization.jl
jd-lara: no, this is an open question for me to retain or not this capability. So, TBD

.claude/claude.md
jd-lara: yes, in the Sequence

src/operation/decision_model.jl
```diff
@@ -617,3 +614,49 @@ function handle_initial_conditions!(model::DecisionModel{<:DecisionProblem})
     end
     return
 end
+
+function _make_device_cache(
+    devices::IS.FlattenIteratorWrapper{T},
+    check_components::Bool,
+    sys::PSY.System,
+    filter_function::Function,
```
daniel-thom: Technically, the function should be the first arg. But it's an internal function and I'm not sure I care.

src/operation/template_validation.jl
jd-lara: This is incorrect. 

```
""" Supertype for all Two Terminal HVDC transmission devices between AC Buses. Not to be confused with [DCBranch](@ref)"""
abstract type TwoTerminalHVDC <: ACBranch end
```

so it won't be an issue

test/test_utils/mock_operation_models.jl
daniel-thom, line 136: Looks like you forgot something.

.gitignore
```diff
@@ -1,4 +1,7 @@
 test_simulation_results/*
+# Claude
+settings.json
```
daniel-thom, line 3: What is this?

jd-lara: Its to prevent accidental checkin of the local claude settings

src/operation/decision_model.jl
jd-lara: This is a change I can make

daniel-thom: Are you planning to make this change?

test/test_utils/mock_operation_models.jl
jd-lara: This is on purpose. For the mock we don't use it

src/operation/decision_model.jl
jd-lara: Done [f9c019f](https://github.com/NREL-Sienna/PowerSimulations.jl/pull/1568/commits/f9c019fea92caff9f34619e4e1a8a4463623e3a8)

test/test_utils/mock_operation_models.jl
daniel-thom: We leave code commented out?

PR #1569, opened 2026-03-22
change serialization to hdf logic

jd-lara:
This PR is meant to always serialize the systems to the HDF. I need to test it on large systems to check the impact and add an opt out approach. 

The motivation for this is to prevent the users having to always call set_system! in the results objects when post processing and making it easier to automate that process. 

In combination with #1568 that removes the serialization of the system should be a clean option. One test is expected to fail in relation to the de-serialization of problems which is a feature no one really uses.

### DIFF
--- a/src/simulation/hdf_simulation_store.jl
+++ b/src/simulation/hdf_simulation_store.jl
@@ -883,7 +883,7 @@ function _deserialize_attributes!(store::HdfSimulationStore)
         end
 
         store.optimizer_stats_datasets[model_name] = problem_group[OPTIMIZER_STATS_PATH]
-        store.optimizer_stats_write_index[model_name] = 0
+        store.optimizer_stats_write_index[model_name] = 1
     end
 
     em_group = _get_emulation_model_path(store)
--- a/src/simulation/simulation.jl
+++ b/src/simulation/simulation.jl
@@ -1172,6 +1172,7 @@ end
 function _serialize_systems_to_json(sim::Simulation)
     simulation_models = get_models(sim)
     results = Dict{String, String}()
+    @debug Threads.threadid() "Serializing systems to JSON in parallel with model building"
     for dm in get_decision_models(simulation_models)
         sys = get_system(dm)
         uuid = string(IS.get_uuid(sys))


src/simulation/simulation.jl
```diff
@@ -614,13 +615,44 @@ function _build!(
...
+    TimerOutputs.@timeit BUILD_PROBLEMS_TIMER "Initialize Problem Storage" begin
+        open_store(HdfSimulationStore, get_store_dir(sim), "w") do store
+            set_simulation_store!(sim, store)
+            try
+                _initialize_problem_storage!(sim)
+                if store_systems_in_results
+                    # Fetch pre-computed JSON from the parallel task and write to HDF5 store.
+                    serialized = fetch(serialization_task)
+                    for (uuid, json_text) in serialized
+                        write_system_json!(store, uuid, json_text)
```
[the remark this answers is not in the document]
jd-lara: @daniel-thom this should be addressed with the fix to the time series to prevent passing multiple systems just to handle the different transformations.

src/simulation/hdf_simulation_store.jl
```diff
@@ -857,7 +883,7 @@ function _deserialize_attributes!(store::HdfSimulationStore)
         end
 
         store.optimizer_stats_datasets[model_name] = problem_group[OPTIMIZER_STATS_PATH]
-        store.optimizer_stats_write_index[model_name] = 0
+        store.optimizer_stats_write_index[model_name] = 1
```
daniel-thom: What was happening here?

src/simulation/simulation.jl
```diff
@@ -972,9 +993,24 @@ function _execute!(
     execution_order = get_execution_order(sim)
     steps = get_steps(sim)
     num_executions = steps * length(execution_order)
-    store_params =
-        _initialize_problem_storage!(sim, cache_size_mib, min_cache_flush_size_mib)
-    status = RunStatus.RUNNING
+    store = get_simulation_store(sim)
+    # For InMemorySimulationStore, initialize storage here since it doesn't persist from build.
```
daniel-thom, line 952: Why is this?

src/simulation/simulation.jl
```diff
@@ -972,9 +993,24 @@ function _execute!(
...
+            something(min_cache_flush_size_mib, MIN_CACHE_FLUSH_SIZE_MiB),
+        )
+    else
+        rules = CacheFlushRules(;
+            max_size = something(cache_size_mib, DEFAULT_SIMULATION_STORE_CACHE_SIZE_MiB) * MiB,
+            min_flush_size = trunc(
+                something(min_cache_flush_size_mib, MIN_CACHE_FLUSH_SIZE_MiB) * MiB,
+            ),
+        )
+        set_cache_flush_rules!(store, rules)
```
daniel-thom, line 967: I may not be following, but didn't we already configure this in `initialize_problem_storage!`?

src/simulation/simulation.jl
```diff
@@ -1133,35 +1169,43 @@ function execute!(sim::Simulation; kwargs...)
...
-                        # Partitioned simulations serialize the systems once during build.
-                        _serialize_systems_to_store!(store, sim)
-                    end
                     log_cache_hit_percentages(store)
                 catch e
                     set_simulation_status!(sim, RunStatus.FAILED)
                     @error "simulation failed" exception = (e, catch_backtrace())
                 end
             end
+            # Delete incomplete store file after it has been closed
```
daniel-thom: Aren't there cases where a developer will need the store in order to debug the failure?

src/simulation/simulation.jl
```diff
@@ -614,13 +615,44 @@ function _build!(
...
+        # Systems are read-only during building, so this is safe.
+        serialization_task = Threads.@spawn _serialize_systems_to_json(sim)
+    end
+
+    try
+        _build_decision_models!(sim)
+        _build_emulation_model!(sim)
+    catch
+        if store_systems_in_results
+            Base.throwto(serialization_task, InterruptException())
```
daniel-thom: Should we be using this function?
```
help?> Base.throwto
  │ Warning
  │
  │  The following bindings may be internal; they may change or be removed in future versions:
  │
  │    •  Base.throwto

  No documentation found for private binding Base.throwto.
```

src/simulation/simulation.jl
jd-lara: I am allowing to overwrite the cache configuration if the users passes the kwargs at the simulation phase. We can move it to the build phase too.

src/simulation/simulation.jl
jd-lara: we never kept it, before if the model build failed it would never persist.

src/simulation/simulation.jl
jd-lara: Good catch, I didn't see that.

src/simulation/hdf_simulation_store.jl
jd-lara: I actually don't recall why we had that 0 there.

.claude/Sienna.md
```diff
@@ -0,0 +1,174 @@
...
+
+#### Unnecessary allocations
+
+  - Use views instead of copies (`@view`, `@views`)
+  - Pre-allocate arrays instead of `push!` in loops
+  - Use in-place operations (functions ending with `!`)
+
+#### Captured variables
+
+Avoid closures that capture variables causing boxing. Pass variables as function arguments instead.
```
daniel-thom: This statement needs some nuance. If the captured variable does not change its type, especially if it is not reassigned, no boxing occurs.
https://docs.julialang.org/en/v1/manual/performance-tips/#man-performance-captured

.claude/Sienna.md
```diff
@@ -0,0 +1,174 @@
...
+
+#### Splatting penalty
+
+Avoid splatting (`...`) in performance-critical code.
```
daniel-thom: If the length of the container is known, such as for a Tuple, this should be ok, right? I'm pretty sure we do this today.

.claude/Sienna.md
```diff
@@ -0,0 +1,174 @@
...
+
+#### Abstract return types
+
+Avoid returning `Union` types or abstract types.
```
daniel-thom: More nuance is needed. It is ok to return a Union of something and Nothing. A factory function will return one of multiple subtypes of an abstract type. Maybe add a statement about performance critical code?

src/simulation/hdf_simulation_store.jl
```diff
@@ -739,6 +756,15 @@ function serialize_system!(store::HdfSimulationStore, sys::PSY.System)
     return
 end
 
+function write_system_json!(store::HdfSimulationStore, uuid::String, json_text::String)
+    root = store.file[HDF_SIMULATION_ROOT_PATH]
+    systems_group = _get_group_or_create(root, "systems")
+    if !haskey(systems_group, uuid)
```
daniel-thom, line 762: So, you can't overwrite an existing UUID? Is there a way to delete the existing entry? Log a warning message? Someone will eventually be surprised by this.

src/simulation/hdf_simulation_store.jl
daniel-thom: This is in the de-serialization / read-only path. We will never write, so the value does not really matter. However, I think that we should leave it as 0 (invalid index), not set (empty dict), or go find the proper write index. 1 is valid but incorrect.

src/simulation/simulation.jl
```diff
@@ -614,22 +598,40 @@ function _build!(
...
+        serialization_task = Threads.@spawn _serialize_systems_to_json(sim)
+    end
+
+    try
+        _build_decision_models!(sim)
+        _build_emulation_model!(sim)
+    catch
+        # Let the serialization task finish naturally — it's read-only and harmless.
+        # The fetch below won't be reached, so the task result is discarded on exit.
+        rethrow()
```
daniel-thom: The try/catch is no longer doing anything.

src/simulation/simulation.jl
daniel-thom: I do not think that is true. This would be a change.

PR #1572, opened 2026-03-23
Fix multiple offshore island issues

rodrigomha:


jd-lara:
@m-bossart does this get fixed in PNM 0.20?

jd-lara:
closed in favor of #1605

PR #1573, opened 2026-03-24
add fixedoutput for sources

m-bossart:


rodrigomha (changes requested):
Overall looks good, but some tests are broken for reading variable results.

PR #1578, opened 2026-03-30
Support adding multiple cost expressions at once

ChenHaotianC:
Refactor add_expressions! and relevant functions to add multiple cost expressions

### DIFF
--- a/src/devices_models/device_constructors/thermalgeneration_constructor.jl
+++ b/src/devices_models/device_constructors/thermalgeneration_constructor.jl
@@ -103,12 +103,25 @@ function construct_device!(
     )
 
     add_expressions!(container, FuelConsumptionExpression, devices, device_model)
-    add_expressions!(container, ProductionCostExpression, devices, device_model)
-    add_expressions!(container, FuelCostExpression, devices, device_model)
-    add_expressions!(container, StartUpCostExpression, devices, device_model)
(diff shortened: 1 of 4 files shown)


src/devices_models/device_constructors/thermalgeneration_constructor.jl
```diff
@@ -103,12 +103,25 @@ function construct_device!(
...
+        ),
+        devices,
+        device_model,
+    )
+    # add_expressions!(container, ProductionCostExpression, devices, device_model)
+    # add_expressions!(container, FuelCostExpression, devices, device_model)
+    # add_expressions!(container, StartUpCostExpression, devices, device_model)
+    # add_expressions!(container, ShutDownCostExpression, devices, device_model)
+    # add_expressions!(container, FixedCostExpression, devices, device_model)
+    # add_expressions!(container, VOMCostExpression, devices, device_model)
```
m-bossart, line 124: Remove commented code when committing

m-bossart (commented):
This PR adds all of the the expression containers at once, but it does not address the concern of repetition when actually building the cost expressions. It also treats the total cost production cost expression and the constituent parts separately, meaning the user is forced to add the cost appropriate to each. This is prone to error. The design should enforce that the ProductionCostExpression is by definition the sum of the constituent parts.

PR #1579, opened 2026-03-30
SCUC branch formulation using MODF

SebastianManriqueM:
Here is a first implementation for this.

I think the ContingencySpec in PNM greatly simplified this work and the handling regarding the reductions.

- We need to merge first the MODF PNM branch so this test pass


src/PowerSimulations.jl
```diff
@@ -431,6 +432,7 @@ export PTDF
 export VirtualPTDF
 export LODF
 export VirtualLODF
+export VirtualMODF
 import InfrastructureSystems: @assert_op, TableFormat, list_recorder_events, get_name
```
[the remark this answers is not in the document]
SebastianManriqueM: @copilot apply changes based on [this feedback](https://github.com/NREL-Sienna/PowerSimulations.jl/pull/1579#discussion_r3012671336)

src/core/network_reductions.jl
```diff
@@ -235,3 +235,26 @@ function get_branch_argument_constraint_axis(
...
+            net_reduction_data,
+            reduced_branch_tracker,
+            branch_type,
+            U,
+        )
+        push!(all_names, names_by_type)
+    end
+
+    return reduce(vcat, all_names)
+end
```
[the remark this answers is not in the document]
SebastianManriqueM: @copilot apply changes based on [this feedback](https://github.com/NREL-Sienna/PowerSimulations.jl/pull/1579#discussion_r3012671357)

src/devices_models/devices/ac_transmission_security_constrained_models.jl
```diff
@@ -83,6 +83,51 @@ function get_emergency_min_max_limits(
...
+        else
+            @warn "Outage $(uuid) associated with $(V) is not registered on the MODF matrix; skipping its post-contingency constraints." maxlog=1
+        end
+    end
+    sort!(ids)
+    return ids
+end
+
 """
 Add branch post-contingency rate limit constraints for ACBranch considering LODF and Security Constraints
```
m-bossart: Update docs to MODF (not LODF)

src/core/network_model.jl
```diff
@@ -28,6 +28,8 @@ Establishes the NetworkModel for a given PowerModels formulation type.
     PTDF/VirtualPTDF matrix produced by PowerNetworkMatrices (optional).
 - `LODF_matrix::Union{PNM.PowerNetworkMatrix, Nothing}` = nothing
```
m-bossart: Do we still need the LODF in the network model? I don't think it is ever used

src/core/network_model.jl
```diff
@@ -58,6 +60,7 @@ mutable struct NetworkModel{T <: PM.AbstractPowerModel}
     use_slacks::Bool
     PTDF_matrix::Union{Nothing, PNM.PowerNetworkMatrix}
     LODF_matrix::Union{Nothing, PNM.PowerNetworkMatrix}
+    MODF_matrix::Union{Nothing, PNM.VirtualMODF}
```
m-bossart, line 63: Compute the MODF when constructing the network if it is not passed by the user (same as PTDF?)

src/devices_models/devices/ac_transmission_security_constrained_models.jl
```diff
@@ -0,0 +1,399 @@
...
+    device_model::DeviceModel{V, U},
+    network_model::NetworkModel{X},
+) where {
+    T <: PostContingencyEmergencyFlowRateConstraint,
+    V <: PSY.ACTransmission,
+    U <: AbstractSecurityConstrainedStaticBranch,
+    X <: AbstractPTDFModel,
+}
+    time_steps = get_time_steps(container)
+    modf_matrix = get_MODF_matrix(network_model)
```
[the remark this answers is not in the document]
m-bossart: This is related to the comment above. Instead of erroring, we can compute the matrix internally.

src/devices_models/devices/ac_transmission_security_constrained_models.jl
```diff
@@ -0,0 +1,400 @@
...
+
+    outage_ids = _post_contingency_outage_ids(sys, modf_matrix, V, device_model)
+    isempty(outage_ids) && return
+
+    net_reduction_data = network_model.network_reduction
+    all_branch_maps_by_type = PNM.get_all_branch_maps_by_type(net_reduction_data)
+    reduced_branch_tracker = get_reduced_branch_tracker(network_model)
+    modeled_branch_types = network_model.modeled_ac_branch_types
+
+    branches_names = get_branch_argument_constraint_axis(
```
m-bossart: I think there are some high level issues with how the constraints are built. 

Right now, for each device type we look at the associated outages and then build post-contingency constraints for all other modeled ac branch types. I think there are two issues here: 

1. When we find an outage, we are building post-contingency constraints for all modeled branch types. I think we should restrict to only devices modeled with a security constrained formulation (e.g. `SecurityConstrainedStaticBranch`). 
2. I don't think we handle properly the case where a branch type has an outage but is not modeled with the security constrained formulation. Then we will never build post-contingency constraints based on that outage.  

We should discuss and align on the expected behavior for these combinations of outage/formulation.

src/devices_models/devices/ac_transmission_security_constrained_models.jl
```diff
@@ -0,0 +1,399 @@
...
+        net_reduction_data,
+        reduced_branch_tracker,
+        modeled_branch_types,
+        PostContingencyEmergencyFlowRateConstraint,
+    )
```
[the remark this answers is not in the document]
m-bossart: Related to the comment above.

test/test_ac_transmission_security_constrained_models.jl
```diff
@@ -0,0 +1,460 @@
...
+        psi_checksolve_test(
+            ps_model,
+            [MOI.OPTIMAL, MOI.ALMOST_OPTIMAL],
+            test_obj_values[sys],
+            10000,
+        )
+    end
+end
+
+@testset "Security Constrained branch formulation Network DC-PF with PTDF/MODF Model" begin
```
m-bossart: Mostly repetitive with the previous testset. Only difference is PTDF vs VirtualPTDF. I don't think this test adds much. Similar for testing these three systems in each of the testsets.

test/test_ac_transmission_security_constrained_models.jl
```diff
@@ -0,0 +1,460 @@
...
+            continue # skipping test for c_sys14_dc as Highs takes so long to find optimal solution
+        end
+        psi_checksolve_test(
+            ps_model,
+            [MOI.OPTIMAL, MOI.ALMOST_OPTIMAL],
+            test_obj_values[sys],
+            10000,
+        )
+    end
+end
```
m-bossart, line 405: Would like to see more targeted testing that things are working as expected. For example, that the constraint tracking is working properly as expected when we have reductions.

m-bossart:
Another design question: How to handle dimensionality mismatch between outages for the flow-gate case where outage-monitored line combinations can be different

src/devices_models/devices/ac_transmission_security_constrained_models.jl
m-bossart: Based on a discussion with @SebastianManriqueM I think that this is already implemented correctly

m-bossart:
> Another design question: How to handle dimensionality mismatch between outages for the flow-gate case where outage-monitored line combinations can be different

@jd-lara Thoughts on this? One option would be to build the container to include all combinations of outages x components x time and then only fill the appropriate constraints based on the flow gates. Doesn't seem like the cleanest option though.

test/test_ac_transmission_security_constrained_models.jl
SebastianManriqueM: Sounds good. Also, virtualPTDF with threads on PTDF constraints and MODF constraints apparently is causing some issues with the KLU solver for factorization of MODF in PNM: ```test_ac_transmission_security_constrained_models: ┌ Error: DecisionModel Build Failed
│   exception =
│    ArgumentError: Invalid Status
│    Stacktrace:
│      [1] kluerror
│        @ C:\Users\smachado\.julia\packages\KLU\Eh2kE\src\KLU.jl:52 [inlined]
│      [2] kluerror
│        @ C:\Users\smachado\.julia\packages\KLU\Eh2kE\src\KLU.jl:60 [inlined]
│      [3] solve!(klu::KLU.KLUFactorization{Float64, Int64}, B::Vector{Float64}; check::Bool)
│        @ KLU C:\Users\smachado\.julia\packages\KLU\Eh2kE\src\KLU.jl:675
│      [4] solve!
│        @ C:\Users\smachado\.julia\packages\KLU\Eh2kE\src\KLU.jl:670 [inlined]
│      [5] _solve_factorization
│        @ C:\Users\smachado\.julia\packages\PowerNetworkMatrices\Vl79O\src\virtual_ptdf_calculations.jl:297 [inlined]
│      [6] _compute_woodbury_factors(mat::VirtualMODF{Tuple{Vector{Tuple{Int64, Int64}}, Vector{Int64}}, Tuple{Dict{Tuple{Int64, Int64}, Int64}, Dict{Int64, Int64}}}, modifications::Tuple{ArcModification})
│        @ PowerNetworkMatrices C:\Users\smachado\.julia\packages\PowerNetworkMatrices\Vl79O\src\woodbury_kernel.jl:122
│      [7] _get_woodbury_factors(vmodf::VirtualMODF{Tuple{Vector{Tuple{Int64, Int64}}, Vector{Int64}}, Tuple{Dict{Tuple{Int64, Int64}, Int64}, Dict{Int64, Int64}}}, mod::NetworkModification)```

test/test_ac_transmission_security_constrained_models.jl
SebastianManriqueM: not sure what you mean by "constraint tracking"?

SebastianManriqueM:
@jd-lara @m-bossart 

Test are passing on my local
image

However, I do have some wierd behavior. some times I got errors related to the KLU solver in PNM.

AI suggest that using threads for both PTDF flow expressions and MODF post contingency flow expressions + KLU solver might be causing issues, specially in windows (I'll solve Matts comments and dig further into this): 

## Why Windows Is Less Stable
The important distinction is that there appear to be two related failure modes in the same code path:

1. A recoverable numerical failure, where KLU returns an invalid/singular solve status and Julia raises `ArgumentError: Invalid Status`.
2. A non-recoverable native failure, where the process crashes inside `libklu.dll` before Julia can surface a normal exception.

The most likely explanation is that the Virtual PTDF/MODF path is combining:

- contingencies that can make the linear solve singular or near-singular, and
- multi-threaded expression construction over shared matrix/factorization state.

Two places in the PowerSimulations code are relevant:

- PTDF branch flow expression construction in `src/devices_models/devices/AC_branches.jl`
- post-contingency MODF expression construction in `src/devices_models/devices/ac_transmission_security_constrained_models.jl`

Both use `Threads.@spawn` while reading PTDF/MODF columns. Those reads ultimately call into `PowerNetworkMatrices`, which uses KLU-backed factorization solves to materialize virtual PTDF/MODF entries lazily.

That matters because KLU factorization objects typically maintain internal mutable work buffers during `solve!`. If multiple tasks read through the same virtual matrix object concurrently, the underlying KLU state may be reused in a way that is not thread-safe. On some platforms this may only surface as a bad return status; on Windows it can instead manifest as an access violation in the native library.

So the Windows instability is not just that the matrix is singular. The stronger issue is:

- singular or ill-conditioned contingency data stresses the solve path, and
- concurrent access to a shared native factorization appears to make failure handling unsafe on Windows.

This also explains why the behavior looked non-deterministic during reproduction: the exact symptom depended on timing and which thread first touched the problematic factorization state.

## Evidence For That Interpretation
Observed stack traces on Windows showed:

- crash in `klu_l_lsolve` inside `libklu.dll`
- calls from `KLU.solve!`
- calls from `PowerNetworkMatrices._solve_factorization`
- calls from `VirtualPTDF` / `VirtualMODF` indexing
- calls from the spawned branch-expression builders in PowerSimulations

This is stronger evidence than a plain numerical infeasibility because the crash happens in native code, not in JuMP or model assembly logic.

The threading hypothesis is also consistent with the surrounding implementation:

- `AC_branches.jl` computes PTDF flow expressions in spawned tasks
- `ac_transmission_security_constrained_models.jl` computes post-contingency MODF expressions in spawned tasks
- both paths read lazily-evaluated PTDF/MODF data rather than precomputed dense arrays

That means a single test can concurrently trigger many KLU solves against shared virtual matrix state.

jd-lara:
The tests are passing pending the release of PFs

docs/src/api/PowerSimulations.md
```diff
@@ -419,6 +425,7 @@ DCCurrentBalance
 PostContingencyGenerationBalanceConstraint
 PostContingencyActivePowerVariableLimitsConstraint
 PostContingencyActivePowerReserveDeploymentVariableLimitsConstraint
+PostContingencyEmergencyFlowRateConstraint
```
m-bossart: For naming, I prefer just `PostContingencyFlowRateConstraint`. I think it is more consistent with the parameter naming (`PostContingencyBranchRatingTimeSeriesParameter` and `BranchRatingTimeSeriesParameter`)

src/core/constraints.jl
```diff
@@ -264,7 +264,23 @@ The specified constraint is formulated as:
...
+each monitored arc m, the post-contingency flow constraint is:
+
+```math
+\\begin{align*}
+&  -R_m \\leq \\sum_{j} \text{PTDF}^\\text{post}_{m,c}[j] \\cdot p_j \\leq R_m,\\quad \\forall t \\in \\{1,\\dots, T\\} \\\\
+&  \\text{PTDF}^{\\text{post}}_{m,c} = vmodf[m, c] \\\\
+\\end{align*}
+```
+"""
+struct PostContingencyEmergencyFlowRateConstraint <: PostContingencyConstraintType end
```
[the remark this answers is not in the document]
m-bossart: Not needed because this was not actually supported previously.

src/core/device_model.jl
```diff
@@ -38,6 +39,12 @@ feedforward to enable passing values between operation model at simulation time
   - `duals::Vector{DataType} = Vector{DataType}()`: use to pass constraint type to calculate the duals. The DataType needs to be a valid ConstraintType
   - `time_series_names::Dict{Type{<:TimeSeriesParameter}, String} = get_default_time_series_names(D, B)` : use to specify time series names associated to the device`
   - `attributes::Dict{String, Any} = get_default_attributes(D, B)` : use to specify attributes to the device
+  - `outages::AbstractVector{<:PSY.Outage} = PSY.Outage[]` : N-1 contingencies to model
+    when the formulation is security-constrained. The constructor records the user's UUID
+    selection; template validation later expands each entry to its per-type set of monitored
```
m-bossart: "The constructor records the user's UUID selection" - not sure what this means, I think it should be clarified in the docstring

src/core/network_model.jl
```diff
@@ -74,7 +77,7 @@ mutable struct NetworkModel{T <: PM.AbstractPowerModel}
         ::Type{T};
         use_slacks = false,
         PTDF_matrix = nothing,
-        LODF_matrix = nothing,
+        MODF_matrix = nothing,
```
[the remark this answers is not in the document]
m-bossart: Again I don't think this is necessary in this case because N-1 code has not been previously supported in the released code.

src/core/network_model.jl
```diff
@@ -144,43 +197,95 @@ function _get_filters(branch_models::BranchModelContainer)
...
+                    ),
+                )
+            end
+            _push_component_buses!(irreducible_buses, component)
+        end
+    end
+    return
+end
+
+function _push_component_buses!(buses::Set{Int64}, branch::PSY.Branch)
```
m-bossart, line 355: This looks like it will fail for three winding transformers. Do we have a check somewhere for not supporting three winding transformers + outages?

src/core/network_model.jl
```diff
@@ -192,7 +297,7 @@ function instantiate_network_model!(
     if isempty(model.subnetworks)
         model.subnetworks = PNM.find_subnetworks(sys)
     end
-    irreducible_buses = _get_irreducible_buses_due_to_dlrs(
+    irreducible_buses = _get_irreducible_buses_due_to_monitored_components(
         sys,
         model,
         branch_models,
```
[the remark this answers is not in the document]
m-bossart: I think the current implementation that always leaves monitored components un-reduced is fine. I think the workflow of passing the matrices to the constructor should be discouraged (and maybe not supported) so it is easier to make the correct simplifications internally to the build. There is not a good performance case for computing the matrices external to the simulation build.

src/core/network_model.jl
```diff
@@ -144,43 +197,95 @@ function _get_filters(branch_models::BranchModelContainer)
...
-            push!(irreducible_buses, bus_to)
-            push!(irreducible_buses, bus_from)
+            _push_component_buses!(irreducible_buses, branch)
         end
     end
-    return collect(irreducible_buses)
+    return
+end
+
+function _add_outage_monitored_irreducible_buses!(
```
m-bossart, line 311: This code pushes the buses of any monitored components to `irreducible_buses`. But shouldn't we also be pushing the buses of the outaged component?

src/core/optimization_container.jl
```diff
@@ -746,9 +746,9 @@ function build_impl!(
         )
     end
 
-    # # Sort branch models so that those with time series parameters (e.g. DLR) come first.
-    # # This ensures that DLR-aware constraint builders claim shared arcs before static builders,
-    # # preventing static constraints from overriding DLR constraints for parallel branches of
+    # # Sort branch models so that those with time series parameters (e.g. TimeSeriesBound) come first.
```
m-bossart: What is `TimeSeriesBound`?  I think this should just give the specific parameter name: (e.g. `BranchRatingTimeSeriesParameter`)

src/core/optimization_container.jl
```diff
@@ -746,9 +746,9 @@ function build_impl!(
...
+    # # This ensures that TimeSeriesBound-aware constraint builders claim shared arcs before static builders,
+    # # preventing static constraints from overriding TimeSeriesBound constraints for parallel branches of
```
m-bossart: Same as above.

src/devices_models/devices/AC_branches.jl
```diff
@@ -361,6 +428,49 @@ function get_min_max_limits(
     )
 end
 
+"""
+Flow limits for an AC transmission reduction entry, honoring the DeviceModel's
+`PARALLEL_BRANCH_MAX_RATING_KEY` attribute for parallel groups. Use this in
+constraint construction where the line's DeviceModel is available; non-line
+formulations should continue to use `get_min_max_limits`.
```
m-bossart: "Use this in constraint construction where the line's DeviceModel is available; non-line formulations should continue to use `get_min_max_limits`." -> I'm confused about what this means. When is the device model not available?

src/devices_models/devices/AC_branches.jl
```diff
@@ -595,17 +705,15 @@ function add_flow_rate_constraint_with_parameters!(
...
-    )
-
     all_branch_maps_by_type = PNM.get_all_branch_maps_by_type(net_reduction_data)
 
+    # `get_constraint_map_by_type[FlowRateConstraint][T]` drops arcs already
+    # claimed by another branch type's static FlowRateConstraint pass, which
+    # would silently skip the TS bound on shared parallel arcs. Walk
+    # `name_to_arc_map[T]` directly, matching `_add_time_series_parameters!`.
+    name_to_arc_map = PNM.get_name_to_arc_map(net_reduction_data, T)
+    branch_names = collect(keys(name_to_arc_map))
```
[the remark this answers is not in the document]
m-bossart: I think this is right, I'm not sure why we would want to not add to the tracker in this case.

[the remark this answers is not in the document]
m-bossart: This seems like an edge case that we are still not guaranteed to handle correctly: Mixed type parallel branch where one of hte branches has a rating time series parameter.

src/operation/template_validation.jl
```diff
@@ -1,3 +1,45 @@
...
+    for (_, device_model) in branch_models
+        _any_component_has_branch_rating_ts(device_model) || continue
+        D = get_component_type(device_model)
+        B = get_formulation(device_model)
+        if B <: StaticBranchUnbounded
+            @warn "BranchRatingTimeSeriesParameter time series attached to \
+                   $(D) components will be ignored: $(B) does not enforce flow \
+                   limits."
+        elseif B <: StaticBranch || B <: AbstractSecurityConstrainedStaticBranch
+            continue
```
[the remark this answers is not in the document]
m-bossart: I think this is a good point that the network model should be checked here for compatibility to give a meaningful error message

src/operation/template_validation.jl
```diff
@@ -103,6 +145,160 @@ function validate_template_impl!(model::OperationModel)
...
+
+# Multi-dispatch outage subtype tag, avoids inline `isa` in the auto-discover
+# path. Default catch-all is `false`; specialized to `true` for `PlannedOutage`.
+_is_planned_outage(::PSY.PlannedOutage) = true
+_is_planned_outage(::PSY.Outage) = false
+
+"""
+Populate `device_model.outages` for every security-constrained branch device
+model in the template, in a single pass over the system's outage supplemental
+attributes. `DeviceModel{D, SC}` claims an outage iff `D` is among the types
```
m-bossart: Should write out explicitly that SC is a security constrained formulation.

src/operation/template_validation.jl
```diff
@@ -103,6 +145,160 @@ function validate_template_impl!(model::OperationModel)
...
+- If `m.outages` is non-empty when this runs, the user explicitly listed UUIDs
+  via the constructor kwarg. Restrict to those UUIDs only; warn for any
+  user-listed UUID that produced no `D`-type entry.
+- If `m.outages` is empty, auto-discover. Honor `"include_planned_outages"`
+  on `m`'s attributes (default `false`) — `PlannedOutage`s are skipped on the
+  auto-discover path unless the attribute is `true`. Explicit user selection
+  bypasses this filter (listing a planned outage explicitly is intent).
+
+Empty `monitored_components` on an outage is treated as "monitor nothing" — a
+warning is emitted. A monitored component whose type is not modeled by the
```
m-bossart: Specify here explicitly what the behavior is when this occurs (i.e. no constraints are built for that monitored component that is not in the template)

src/operation/template_validation.jl
```diff
@@ -103,6 +145,160 @@ function validate_template_impl!(model::OperationModel)
...
+template is reported once per type with the offending outage UUIDs.
+"""
```
m-bossart, line 273: This docstring has a lot of good clarifications about the use and expected behavior of the outage modeling. I wonder if we can have some of this content in a more public facing docstring. I'm not sure where it belongs, possibly in `SecurityConstrainedStaticBranch`? Especially the part about the outaged (not monitored) component needing the security constrained formulation.

src/PowerSimulations.jl
```diff
@@ -436,11 +438,12 @@ import PowerSystems as PSY
 import InfrastructureSystems as IS
 import PowerFlows
 import PowerNetworkMatrices as PNM
-import PowerNetworkMatrices: PTDF, VirtualPTDF, LODF, VirtualLODF
+import PowerNetworkMatrices: PTDF, VirtualPTDF, LODF, VirtualLODF, VirtualMODF
```
m-bossart: We no longer need to import/export `LODF` and `VirtualLODF`?

test/test_network_constructors_with_branch_rating_time_series.jl
```diff
@@ -376,47 +396,47 @@ end
...
     ]
-    branches_dlr = ["1", "2", "6"]
-    dlr_factors = vcat([fill(x, 6) for x in [0.99, 0.98, 1.0, 0.95]]...)
+    branches_with_rating_ts = ["1", "2", "6"]
+    rating_factors = vcat([fill(x, 6) for x in [0.99, 0.98, 1.0, 0.95]]...)
 
-    test_obj_values = [356577.0, 279735.0, 241293.703]
-    parallel_lines_names_to_add = ["1", "2", "3"]#Add parallel lines in lines with and without DLRs
+    test_obj_values = [243859.89, 243884.35, 243877.86]
+    parallel_lines_names_to_add = ["1", "2", "3"]#Add parallel lines in lines with and without TimeSeriesBounds like TimeSeriesBound
```
m-bossart: confusing comment

test/test_network_constructors_with_branch_rating_time_series.jl
```diff
@@ -376,47 +396,47 @@ end
...
     n_steps = 2
     test_results_slacks = Dict(
         1 => [456, 0, 288, 288, 24],
         2 => [456, 0, 288, 288, 24],
-        3 => [408, 0, 264, 264, 24],
+        3 => [408, 0, 288, 288, 24],
```
m-bossart, line 426: What is driving these value changes?

test/test_ac_transmission_security_constrained_models.jl
```diff
@@ -0,0 +1,876 @@
...
+        # Monitoring `all_branches` would pin every bus as irreducible and
+        # cancel the reduction the test exists to exercise.
+        for line_name in lines_outages[sys]
+            component = get_component(ACTransmission, sys, line_name)
+            transition_data = GeometricDistributionForcedOutage(;
+                mean_time_to_recovery = 10,
+                outage_transition_probability = 0.9999,
+                monitored_components = [component],
+            )
+            add_supplemental_attribute!(sys, component, transition_data)
```
SebastianManriqueM, line 319: I feel this test is kind of trivial, we are adding the monitored line to the same outed lin, so the post contingency flow through the outed line will be always zero

src/operation/template_validation.jl
SebastianManriqueM: I think that the default should be either all the branches are monitored (this is the classic scuc without "flowgates") rather than nothing

SebastianManriqueM (commented):
Generally speaking I dont have anything to add to Matt comments rather than:
- The test that we had where the outed component is the same monitored component is kind of trivial (I added a non-trivial test)
- Just a suggestion that probably requires more discussion and each one might have his own preference but if the user does not select any monitored component in the outage, I think the default should be consider all the branches (Traditional scuc "without flowgates")

src/core/network_model.jl
jd-lara: resolved

src/core/network_model.jl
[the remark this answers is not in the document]
jd-lara: I agree with this but it should be discussed at the Meeting first. The only rationale for precomputing the matrix is passing it to multiple models. That should be a follow up PR where we actually calculate just the the ABA factorization and calculate PTDF or MODF on top accordingly

src/operation/template_validation.jl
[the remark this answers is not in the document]
jd-lara: I added compatibility to the other formulations in this PR. The comment doesn't apply

src/operation/template_validation.jl
jd-lara: no, it should be nothing. It makes no sense to let the user end up with a huge problem if they do n-1 everything vs everything

test/performance/performance_test.jl
```diff
@@ -63,7 +63,6 @@ function copy_single_time_series_across_systems!(dst_sys::System, src_sys::Syste
     )
         dst_component =
             PSY.get_component(typeof(src_component), dst_sys, PSY.get_name(src_component))
```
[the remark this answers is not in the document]
jd-lara: This is an incorrect suggestion. At this point the code, it should throw an error if the component isn't found. This can mask an issue earlier in the code.

src/core/network_model.jl
jd-lara: Added and explicit check to throw an error

src/devices_models/devices/AC_branches.jl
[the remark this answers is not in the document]
jd-lara: This should already be handled in the parameter addition not here.

src/core/network_model.jl
jd-lara, line 1: @m-bossart and @SebastianManriqueM the changes here are 100% complicated to handle all the combinations of time series, monitored components and outage modeling. It need a careful review

src/operation/template_validation.jl
```diff
@@ -1,3 +1,71 @@
+function _any_component_has_branch_rating_ts(device_model::DeviceModel, sys::PSY.System)
+    ts_name = get(
```
jd-lara: drive me crazy when claude uses these patterns.

src/operation/template_validation.jl
```diff
@@ -1,3 +1,71 @@
...
+
+function _check_branch_rating_time_series_formulation!(
+    branch_models::Dict,
+    sys::PSY.System,
+)
+    for (_, device_model) in branch_models
+        _any_component_has_branch_rating_ts(device_model, sys) || continue
+        D = get_component_type(device_model)
+        B = get_formulation(device_model)
+        if B <: StaticBranchUnbounded
```
jd-lara: Should error here

src/devices_models/devices/AC_branches.jl
```diff
@@ -701,13 +843,33 @@ function add_constraints!(
     if use_slacks
         slack_ub = get_variable(container, FlowActivePowerSlackUpperBound(), B)
     end
+
+    has_ts_rating =
+        haskey(get_time_series_names(device_model), BranchRatingTimeSeriesParameter)
+    branches_with_ts = Set{String}()
+    if has_ts_rating
```
jd-lara: This is a really silly way to check for the parameter time series

src/simulation/simulation_events.jl
```diff
@@ -22,14 +26,12 @@ function extend_event_parameters!(simulation::Simulation, event_model)
     sequence = get_sequence(simulation)
     sim_state = get_simulation_state(simulation)
     em_model = get_emulation_model(get_models(simulation))
+    isnothing(em_model) && return
```
jd-lara: This is useless since it fails earlier

jd-lara: emulator can't be nothing 
```julia
em_model = get_emulation_model(get_models(simulation))
    isnothing(em_model) && error(
        "Simulation has events configured but no emulation model; events require an emulator",
    )
```

docs/src/explanation/branch_rating_limits.md
```diff
@@ -0,0 +1,188 @@
...
+This page explains how PowerSimulations derives a flow **rating** for an AC
+transmission element and how that rating is turned into optimization
+constraints. It covers every reduction type produced by
+`PowerNetworkMatrices` (PNM), the formulations that consume the rating, the
+difference between linear and nonlinear network models, the parallel-branch
+aggregation attribute and its defaults, and the branch-rating time series.
+
+For the per-formulation constraint algebra (variable names, slacks, objective
+terms) see the [`PowerSystems.Branch` Formulations](@ref) page in the
+Formulation Library. This page is the conceptual companion: *where the number
```
m-bossart: "Where the constraint branch flow limit on the right-hand side comes from"

src/devices_models/devices/ac_transmission_security_constrained_models.jl
```diff
@@ -0,0 +1,814 @@
...
+) where {T <: PSY.ACTransmission}
+    monitored = _monitored_component_names(device_model, T)
+    monitored_devices = [d for d in devices if PSY.get_name(d) in monitored]
+    isempty(monitored_devices) && return
+    add_parameters!(
+        container,
+        PostContingencyBranchRatingTimeSeriesParameter,
+        monitored_devices,
+        device_model,
+    )
```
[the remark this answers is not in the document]
jd-lara: This seems incorrect, the use of the reductions mapping in handled inside the add_parameters call

src/operation/template_validation.jl
```diff
@@ -1,3 +1,76 @@
+function _any_component_has_branch_rating_ts(device_model::DeviceModel, sys::PSY.System)
+    haskey(get_time_series_names(device_model), BranchRatingTimeSeriesParameter) ||
+        return false
+    ts_name = get_time_series_names(device_model)[BranchRatingTimeSeriesParameter]
```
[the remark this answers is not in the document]
jd-lara: Confirm the check that prevents the addition of time series for `StaticBranchBounds` is used later and the case for `StaticBranchUnbounded` is just ignored

src/core/device_model.jl
```diff
@@ -38,6 +39,14 @@ feedforward to enable passing values between operation model at simulation time
   - `duals::Vector{DataType} = Vector{DataType}()`: use to pass constraint type to calculate the duals. The DataType needs to be a valid ConstraintType
   - `time_series_names::Dict{Type{<:TimeSeriesParameter}, String} = get_default_time_series_names(D, B)` : use to specify time series names associated to the device`
   - `attributes::Dict{String, Any} = get_default_attributes(D, B)` : use to specify attributes to the device
+  - `outages::AbstractVector{<:PSY.Outage} = PSY.Outage[]` : N-1 contingencies to model
+    when the formulation is security-constrained. The constructor stores the
+    `IS.get_uuid(outage)` of each entry as a key in the model's
+    `outages::Dict{UUID, Dict{DataType, Set{String}}}` field with empty inner maps;
+    template validation later fills the inner maps with the per-type set of monitored
+    component names that each outage carries. An empty default triggers auto-discovery
+    of every outage in the system that monitors at least one component of type `D`.
```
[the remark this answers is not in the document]
jd-lara: This comment is incorrect.

test/test_events.jl
```diff
@@ -237,6 +237,47 @@ end
...
+    # to skip RenewableDispatch in `extend_event_parameters!`. The existing
+    # "Renewable outage" testset only checks the emulator; this asserts the
+    # outage is also projected into the D2 *decision* model's status/power.
+    dates_ts = collect(
+        DateTime("2024-01-01T00:00:00"):Hour(1):DateTime("2024-01-02T23:00:00"),
+    )
+    outage_data = fill!(Vector{Int64}(undef, 48), 0)
+    outage_data[3] = 1
+    outage_data[10:11] .= 1
+    outage_data[23:22] .= 1
```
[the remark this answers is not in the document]
m-bossart: This should be changed to 22:23, but the test is fine because it is just comparing that status and active power variable are matching.

docs/src/explanation/branch_rating_limits.md
```diff
@@ -0,0 +1,191 @@
...
+
+The active-power flow is bounded symmetrically by the aggregated rating:
+
+```math
+-R \;\le\; f_{b,t} \;\le\; R
+```
+
+applied through `FlowRateConstraint` (PTDF and DC) or, for
+`StaticBranchBounds`, directly as variable bounds on the flow variable.
+`MonitoredLine` is the exception — it carries explicit, possibly asymmetric
```
m-bossart: We should describe here how this exception interacts with the content above about using `branch_rating` to derive a single value and then `min_max_flow_limits` to make the symmetrical limits. Its not clear how this happens when you have for example a MonitoredLine with asymmetrical flow limits that is within a reduction.

docs/src/explanation/branch_rating_limits.md
```diff
@@ -0,0 +1,191 @@
...
+    emergency rating for parallel groups, and the type-aware emergency rating
+    otherwise.
+
+Support and validation:
+
+  - Branch rating time series are only supported with `StaticBranch` and
+    `AbstractSecurityConstrainedStaticBranch`. Any other formulation —
+    including `StaticBranchUnbounded`, which enforces no flow limits — raises a
+    `ConflictingInputsError` rather than silently ignoring the time series.
+  - At build a sanity check warns if the normalized series scaled by the
```
m-bossart, line 167: I wonder if it makes sense to have some threshold. I think a slight derating would be possible in practical applications.

docs/src/explanation/branch_rating_limits.md
```diff
@@ -0,0 +1,191 @@
...
+
+## [MonitoredLine](@id monitored_line_rating)
+
+`PSY.MonitoredLine` does not use the symmetric `branch_rating` path because it
+carries explicit `flow_limits` that can be tighter than, and asymmetric to, its
+rating:
+
+  - General rate limit: ``\min(\text{rating},\ \text{flow\_limits.from\_to},\ \text{flow\_limits.to\_from})``,
+    applied symmetrically. A warning is emitted when the from-to and to-from
+    limits differ (the minimum is then used).
```
m-bossart: This is confusing to me. We seem to imply that we can have asymmetric limits above, but then this seems like the asymmetry is ignored and a warning is issued.

src/core/formulations.jl
```diff
@@ -136,6 +136,62 @@ struct StaticBranchBounds <: AbstractBranchFormulation end
...
+Concretely, for every monitored ACTransmission branch and every claimed
+outage, this formulation adds:
+
+```
+-rate_emergency ≤ post_contingency_flow ≤ rate_emergency
+```
+
+where `post_contingency_flow` is derived from the modification factors
+(MODF) provided by `PowerNetworkMatrices` and `rate_emergency` comes from
+the branch's `rating` (or `rating_b` when a `PostContingencyBranchRatingTimeSeriesParameter`
```
m-bossart: This docstring is misleading. The `rate_emergency` should come from `rating_b` even when the rating is not time varying.

src/core/optimization_container.jl
```diff
@@ -746,10 +746,10 @@ function build_impl!(
         )
     end
 
-    # # Sort branch models so that those with time series parameters (e.g. DLR) come first.
-    # # This ensures that DLR-aware constraint builders claim shared arcs before static builders,
-    # # preventing static constraints from overriding DLR constraints for parallel branches of
-    # # different types sharing the same arc.
+    # # Sort branch models so that those with time series parameters (e.g. BranchRatingTimeSeriesParameter)
```
m-bossart: I think this entire comment is no longer relevant and should be deleted (also the commented sort code below)

src/devices_models/devices/AC_branches.jl
```diff
@@ -27,8 +27,8 @@ get_parameter_multiplier(::UpperBoundValueParameter, ::PSY.ACTransmission, ::Abs
 
 get_variable_multiplier(::PhaseShifterAngle, d::PSY.PhaseShiftingTransformer, ::PhaseAngleControl) = 1.0/PSY.get_x(d)
 
-get_multiplier_value(::AbstractDynamicBranchRatingTimeSeriesParameter, d::PSY.ACTransmission, ::StaticBranch) = PSY.get_rating(d)
-get_multiplier_value(::AbstractDynamicBranchRatingTimeSeriesParameter, d::PNM.BranchesParallel, ::StaticBranch) = PNM.get_equivalent_rating(d)
+get_multiplier_value(::AbstractBranchRatingTimeSeriesParameter, d::PSY.ACTransmission, ::StaticBranch) = PSY.get_rating(d)
+get_multiplier_value(::AbstractBranchRatingTimeSeriesParameter, d::PSY.ACTransmission, ::AbstractSecurityConstrainedStaticBranch) = PSY.get_rating(d)
```
m-bossart, line 31: Shouldn't this be `rating_b`?

src/devices_models/devices/AC_branches.jl
```diff
@@ -267,98 +316,46 @@ end
...
-            max = PSY.get_rating_secondary(transformer),
-        )
-    elseif winding_number == 3
-        limits = (
-            min = -1 * PSY.get_rating_tertiary(transformer),
-            max = PSY.get_rating_tertiary(transformer),
-        )
-    end
-    return limits
+function branch_rating(entry, ::DeviceModel)
```
m-bossart, line 329: `entry::ACTransmission` ?

src/devices_models/devices/AC_branches.jl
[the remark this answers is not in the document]
m-bossart: But don't we want to avoid the repeated constraint? I think this is correct below where we are using the tracker but we might end up with an extra entry in the constraint container that never gets filled below. I think using the tracker to define the container name axes was correct.

src/devices_models/devices/ac_transmission_security_constrained_models.jl
```diff
@@ -0,0 +1,869 @@
...
+    _assign_container!(container.constraints, ConstraintKey(T, V, meta), cons_container)
+    return cons_container
+end
+
+"""
+For each outage in `device_model.outages`, resolve every monitored component
+(across every monitored type) to its arc in the active reduction graph —
+using `component_to_reduction_name_map` as a redirect when the monitored name
+is an individual component that was reduced into a representative (e.g. `"3"`,
+`"3_copy"` → `"3double_circuit"`). Duplicate arcs within an outage are
```
m-bossart: The example with the names is just confusing, I would remove

src/devices_models/devices/ac_transmission_security_constrained_models.jl
```diff
@@ -0,0 +1,869 @@
...
+where each inner tuple is `(monitored_type, container_name, arc, reduction_kind)`.
+The `monitored_type` is needed by callers to fetch the correct post-contingency
+flow variable for that component.
+"""
+function _resolve_monitored_arcs(
+    device_model::DeviceModel,
+    net_reduction_data::PNM.NetworkReductionData,
+)
+    name_to_arc_maps = PNM.get_name_to_arc_maps(net_reduction_data)
+    c2r_maps = PNM.get_component_to_reduction_name_map(net_reduction_data)
```
m-bossart: spell out variable name for consistency

src/devices_models/devices/ac_transmission_security_constrained_models.jl
```diff
@@ -0,0 +1,869 @@
...
+
+    resolved = _resolve_monitored_arcs(device_model, net_reduction_data)
+
+    con_lb = _add_post_contingency_sparse_constraints!(container, T, V; meta = "lb")
+    con_ub = _add_post_contingency_sparse_constraints!(container, T, V; meta = "ub")
+
+    expressions = get_expression(container, PostContingencyBranchFlow(), V)
+    jump_model = get_jump_model(container)
+
+    # Multi-component outage dedup: if another SC DeviceModel already added
```
m-bossart, line 375: Should this functionality for checking for already existing constraints be included in the branch tracker? It is very similar functionally to what we do for non-emergency constraints.

src/devices_models/devices/ac_transmission_security_constrained_models.jl
```diff
@@ -0,0 +1,869 @@
...
+    registered_contingencies = PNM.get_registered_contingencies(modf_matrix)
+
+    net_reduction_data = network_model.network_reduction
+    resolved = _resolve_monitored_arcs(model, net_reduction_data)
+
+    expression_container = _add_post_contingency_sparse_expression!(
+        container, T, V, resolved, time_steps,
+    )
+
+    # Multi-component outage dedup: an outage attached to N>1 component types
```
m-bossart, line 483: What is the intended behavior when an outage is associated with two different component types but only one of them has a security constrained formulation? I would think we shouldn't silently only add one of the outages but rather warn or error.

src/simulation/simulation_events.jl
```diff
@@ -21,16 +25,14 @@ end
...
     sim_state = get_simulation_state(simulation)
+    # `apply_simulation_events!` already errored if there is no emulation
+    # model, and it is the only caller — `em_model` cannot be nothing here.
     em_model = get_emulation_model(get_models(simulation))
     model_name = get_name(em_model)
     for (event_uuid, device_type_maps) in
         event_model.attribute_device_map[model_name]
         sim_time = get_current_time(simulation)
         for (dtype, device_names) in device_type_maps
-            if dtype == PSY.RenewableDispatch
```
m-bossart, line 30: Do you recall why this was in here?

m-bossart:
When we move this to POM we should try to make the testing of these three interacting features and the (many) combinations/edge cases that might arise more systematic and comprehensive:

1. Time varying branch ratings
2. Network reductions 
3. Outages/ Security constrained formulations.

Here are some initial notes on relevant combinations to test (many of these are already included, but just need to be systematized so we don't end up with a test file that is many thousands of lines...)
image

src/simulation/simulation_events.jl
jd-lara: I determined this was left over code that we never removed

src/devices_models/devices/ac_transmission_security_constrained_models.jl
jd-lara: we have flow gates for mixed transformers and lines... this is the best idea I could come up with

docs/src/explanation/branch_rating_limits.md
jd-lara: We can extend after this work is done

m-bossart:
> When we move this to POM we should try to make the testing of these three interacting features and the (many) combinations/edge cases that might arise more systematic and comprehensive:
> 
> 1. Time varying branch ratings
> 2. Network reductions
> 3. Outages/ Security constrained formulations.
> 
> Here are some initial notes on relevant combinations to test (many of these are already included, but just need to be systematized so we don't end up with a test file that is many thousands of lines...) image

One more case to test: Flow gate with mixed components outaged where only one of the component types has a security constrained formulation. I think at least we need a warning in this case that the outage being modeled does not match the flowgate specification.

m-bossart (approved):

m-bossart:
Looks good pending the tests

jd-lara (approved):

PR #1611, opened 2026-05-13
deserialize system in results

m-bossart:
By default deserializes the system for each problem from the simulation store

### DIFF
--- a/src/simulation/simulation_results.jl
+++ b/src/simulation/simulation_results.jl
@@ -103,6 +103,11 @@ function SimulationResults(
         container_key_lookup = get_container_key_lookup(store)
         for (name, problem_params) in sim_params.decision_models_params
             name = string(name)
+            system = if has_system(store, get_system_uuid(problem_params))
+                deserialize_system(store, get_system_uuid(problem_params))
+            else
+                nothing
+            end
             problem_result = SimulationProblemResults(
                 DecisionModel,
                 store,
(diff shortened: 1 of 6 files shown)


src/simulation/simulation_results.jl
```diff
@@ -103,26 +103,39 @@ function SimulationResults(
         container_key_lookup = get_container_key_lookup(store)
         for (name, problem_params) in sim_params.decision_models_params
             name = string(name)
+            system = if has_system(store, get_system_uuid(problem_params))
```
m-bossart, line 106: Should the HDFSimulationStore always have the system? If so this should error and not just return nothing?

jd-lara:
### Code review

Found 1 issue:

1. `has_system` (and the existing `deserialize_system`) call `_get_group_or_create(root, "systems")`, which invokes `HDF5.create_group` when the `"systems"` group is absent. The new file-path `SimulationResults` constructor opens the store in read-only mode (`open_store(HdfSimulationStore, simulation_store_path, "r")`), so this throws an HDF5 error for any store that does not already have a `"systems"` group — for example, a simulation built with `store_systems_in_results=false`, or an older store written before that group was introduced. Before this PR the same pattern in `deserialize_system` was only reachable via an explicit `get_system!` call; this PR moves it onto the default load path, so any such store will fail to load via `SimulationResults(path)`. A `haskey(root, "systems")` guard (returning `false`/erroring as appropriate) avoids the write attempt on a read-only handle.

https://github.com/Sienna-Platform/PowerSimulations.jl/blob/640652b5ecd424381f15b31e506ae69cd3bfef04/src/simulation/hdf_simulation_store.jl#L767-L784

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

jd-lara (changes requested):
In this PR any store without a "systems" group (e.g. built with `store_systems_in_results=false`) will fail to load.


  Suggested change in `src/simulation/hdf_simulation_store.jl`

```julia
  function has_system(store::HdfSimulationStore, uuid::Base.UUID)
      root = store.file[HDF_SIMULATION_ROOT_PATH]
      haskey(root, "systems") || return false
      return haskey(root["systems"], string(uuid))
  end

  function deserialize_system(store::HdfSimulationStore, uuid::Base.UUID)
      root = store.file[HDF_SIMULATION_ROOT_PATH]
      uuid_str = string(uuid)
      if !haskey(root, "systems") || !haskey(root["systems"], uuid_str)
          error("No system with UUID $uuid_str is stored")
      end
      json_text = HDF5.read(root["systems"][uuid_str])
      return PSY.from_json(json_text, PSY.System)
  end
```

m-bossart:
@jd-lara Are these test failures related to power models expected?

jd-lara:
> @jd-lara Are these test failures related to power models expected?

Yeah, I just rebased your branch to merge

PR #1613, opened 2026-05-19
network bugfixes

jd-lara:
These are bugs from trying to run the SCUC model in the CATS model


src/core/network_model.jl
```diff
@@ -612,6 +663,17 @@ function instantiate_network_model!(
                 model.network_reduction,
             )
         end
+        # Reconcile PTDF/MODF reductions before outage consolidation populates
+        # the branch maps.
+        if _reconcile_ptdf_modf_reduction!(model, sys)
+            model.network_reduction =
```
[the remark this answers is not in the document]
jd-lara: we will keep the existing behavior of reconciling the buses

docs/src/how_to/security_constrained_models.md
```diff
@@ -0,0 +1,204 @@
...
+### Gotcha 3 — Reported `PTDFBranchFlow` orientation under reductions
+
+When degree-two/radial reductions merge several branches into one equivalent,
+each member branch's `PTDFBranchFlow` is assembled from the *representative*
+arc's PTDF column. For a member whose native `from → to` arc is opposite the
+merged-path direction, the recorded flow was historically **sign-flipped**
+relative to its own orientation (and relative to
+`PowerFlowBranchActivePowerFromTo`).
+
+This is now corrected: each member's reported `PTDFBranchFlow` matches its
```
m-bossart, line 159: Aren't we still reporting PTDFBranchFlow by reduction entry. So this is necessarily false for some branches if all the branches in the chain are not oriented in the same direction?

src/core/network_model.jl
```diff
@@ -169,23 +169,64 @@ function _consolidate_device_model_outages_with_modf!(
...
     end
     return reductions
 end
 
+"""
+Buses retained by `nrd` (reduction representatives plus irreducible buses);
+this is the matrix's bus dimension.
+"""
+function _retained_buses(nrd::PNM.NetworkReductionData)
+    return union(Set(keys(PNM.get_bus_reduction_map(nrd))), PNM.get_irreducible_buses(nrd))
```
m-bossart: I'm wondering why we need this union? In what case would the irreducible buses not be included in the keys of the bus reduction map? And maybe we are just masking an issue with this.

src/core/network_model.jl
```diff
@@ -169,23 +169,64 @@ function _consolidate_device_model_outages_with_modf!(
...
+"""
+Rebuild PTDF and MODF onto the union of their retained buses when they diverge.
+Returns `true` if a rebuild happened; throws if one pass fails to converge them,
+since mismatched reductions break the nodal-balance vs. MODF-column dimensions.
+"""
+function _reconcile_ptdf_modf_reduction!(
+    model::NetworkModel{<:AbstractPTDFModel},
+    sys::PSY.System,
+)
+    ptdf_nrd = model.PTDF_matrix.network_reduction_data
```
m-bossart: use a getter

test/test_ac_transmission_security_constrained_models.jl
```diff
@@ -727,6 +681,117 @@ end
     @test n_checked >= 1
 end
 
+@testset "PTDFBranchFlow member orientation sign under network reduction" begin
```
m-bossart, line 684: Should this be moved out of the test file for security constrained?

src/devices_models/devices/common/add_to_expression.jl
```diff
@@ -2029,6 +1968,9 @@ function add_to_expression!(
                         net_reduction_data,
                     )
                 end
+                # flow_expression is native-oriented; re-apply the member sign
+                # so the interface total is orientation-invariant.
+                direction *= get_ptdf_orientation_sign(net_reduction_data, br_type, name)
```
m-bossart, line 1973: I don't understand this. In this case name will be the name of the reduction. And `get_ptdf_orientation_sign` is trying to match that name to one of the segments within the reduction.

test/test_ac_transmission_security_constrained_models.jl
```diff
@@ -727,6 +681,117 @@ end
...
+                expected = PSI.get_hinted_aff_expr(length(nz_idx))
+                for i in nz_idx
+                    JuMP.add_to_expression!(expected, ptdf_col[i], nodal_balance[i, t])
+                end
+                @test JuMP.isequal_canonical(pbf[name, t], expected)
+            end
+        end
+    end
+    @test n_checked >= 1
+end
```
m-bossart, line 761: I think this is missing a test for the case where direction -1 that things are built correctly within a simulation build.

src/devices_models/devices/AC_branches.jl
```diff
@@ -1024,24 +976,23 @@ function add_expressions!(
...
+                (e, catch_backtrace())
             rethrow()
         end
     end
     for task in tasks
         name, expressions = fetch(task)
-        branch_flow_expr[name, :] .= expressions
+        # ptdf[arc,:] uses the representative's orientation; flip series members
+        # back to native from→to. sign == 1.0 fast path skips a broadcast-multiply.
+        orientation_sign = get_ptdf_orientation_sign(net_reduction_data, B, name)
```
m-bossart, line 990: I don't understand the need for multiplying by orientation_sign. Orientation_sign is the orientation between a segment of the series reduction and the PTDF flow expression. In the container, name is the name of the reduction entry, so the sign of that flow should not depend on an individual segment?

docs/src/how_to/security_constrained_models.md
jd-lara: As we discussed this is fixed in the addition of the expressions

test/test_ac_transmission_security_constrained_models.jl
jd-lara: we can update in the move to POM cc. @acostarelli

test/test_ac_transmission_security_constrained_models.jl
jd-lara: we can update in the move to POM cc. @acostarelli

src/core/network_model.jl
```diff
@@ -197,8 +202,8 @@ function _reconcile_ptdf_modf_reduction!(
     model::NetworkModel{<:AbstractPTDFModel},
     sys::PSY.System,
 )
-    ptdf_nrd = model.PTDF_matrix.network_reduction_data
-    modf_nrd = get_MODF_matrix(model).network_reduction_data
+    ptdf_nrd = PNM.get_network_reduction_data(model.PTDF_matrix)
```
m-bossart, line 205: Use `get_PTDF_matrix`

src/core/network_model.jl
```diff
@@ -215,8 +220,8 @@ function _reconcile_ptdf_modf_reduction!(
     model.MODF_matrix =
         PNM.VirtualMODF(sys; tol = PTDF_ZERO_TOL, network_reductions = reductions)
 
-    if _retained_buses(model.PTDF_matrix.network_reduction_data) !=
-       _retained_buses(get_MODF_matrix(model).network_reduction_data)
+    if _retained_buses(PNM.get_network_reduction_data(model.PTDF_matrix)) !=
```
m-bossart, line 223: use get_PTDF_matrix

m-bossart (approved):
I left a few very minor comments about using getters. Pending the tests passing this looks good to me

PR #1624, opened 2026-06-04
fix issue 1623 by adding extra sign for special case

### LINKED ISSUE #1623: PowerFlow in the loop mismatch with Sources
When we have sources that inject or withdraw power with a fixed output formulation, the PF object is not updated properly. This leads to mismatches in the flows from the optimization problem and the PF. MWE below. If the source formulation is removed, the results match (as expected) 
MWE:
```
# Builds the c_sys5_uc 5-bus system, adds a Source to nodeC with separate in/out
# time series so the source exports to the grid in simulation step 1 and imports
# from it in step 2.  Runs a 2-step simulation with PTDFPowerModel and
# DCPowerFlow in the loop, filtering to only model line "1" (nodeA → nodeB).
# Finally reads PTDFBranchFlow and PowerFlowBranchActivePowerFromTo for that
# line and prints the element-wise comparison.

using PowerSystems
using PowerSimulations
using PowerFlows
using PowerNetworkMatrices
using PowerSystemCaseBuilder
using Dates
using HiGHS
using JuMP
using TimeSeries
using Logging

const PSI = PowerSimulations
const PSY = PowerSystems
import PowerSystemCaseBuilder: PSITestSystems

HiGHS_optimizer = JuMP.optimizer_with_attributes(
    HiGHS.Optimizer,
    "time_limit" => 100.0,
    "random_seed" => 12345,
    "log_to_console" => false,
)

# ── 5-bus system with single time series (needed for source TS timestamps) ──
sys = build_system(PSITestSystems, "c_sys5_uc"; add_single_time_series = true)

# ── add Source to nodeC ──
source = PSY.Source(;
    name = "interconnect",
    available = true,
    bus = get_component(ACBus, sys, "nodeC"),
    active_power = 0.0,
    reactive_power = 0.0,
    active_power_limits = (min = -1.0, max = 1.0),
    reactive_power_limits = (min = -1.0, max = 1.0),
    R_th = 0.01,
    X_th = 0.02,
    internal_voltage = 1.0,
    internal_angle = 0.0,
    base_power = 100.0,
)
add_component!(sys, source)

# ── build in/out TS that alternate between export and import across steps ──
# Borrow 48-hour timestamps from an existing load
load = first(get_components(PowerLoad, sys))
tstamp = TimeSeries.timestamp(
    get_time_series_array(SingleTimeSeries, load, "max_active_power"),
)

# Step 1 (hours 0–23): source provides power to grid (out active, in zero)
# Step 2 (hours 24–47): source takes power from grid (out zero, in active)
out_vals = vcat(fill(0.5, 24), fill(0.0, 24))
in_vals  = vcat(fill(0.0, 24), fill(0.3, 24))

add_time_series!(
    sys, source,
    SingleTimeSeries(
        "max_active_power_out",
        TimeArray(tstamp, out_vals);
        scaling_factor_multiplier = PSY.get_max_active_power,
    ),
)
add_time_series!(
    sys, source,
    SingleTimeSeries(
        "max_active_power_in",
        TimeArray(tstamp, in_vals);
        scaling_factor_multiplier = PSY.get_max_active_power,
    ),
)

# Convert all single time series to 24-h deterministic forecasts (2 steps)
transform_single_time_series!(sys, Hour(24), Hour(24))

# ── template: PTDFPowerModel + DCPowerFlow in the loop, one line only ──
target_line_name = "1"   # Line "1": nodeA → nodeB
@assert !isnothing(get_component(Line, sys, target_line_name))

network_model = NetworkModel(
    PTDFPowerModel;
    PTDF_matrix = PTDF(sys),
    power_flow_evaluation = DCPowerFlow(),
)

template = ProblemTemplate(network_model)
set_device_model!(template, ThermalStandard, ThermalBasicUnitCommitment)
set_device_model!(template, PowerLoad, StaticPowerLoad)
set_device_model!(template, DeviceModel(Source, FixedOutput))
set_device_model!(
    template,
    DeviceModel(
        Line,
        StaticBranch;
        attributes = Dict{String, Any}(
            "filter_function" => x -> get_name(x) == target_line_name,
        ),
    ),
)

uc_model = DecisionModel(
    template,
    sys;
    name = "UC",
    optimizer = HiGHS_optimizer,
    store_variable_names = true,
    optimizer_solve_log_print = false,
)

models   = SimulationModels(; decision_models = [uc_model])
sequence = SimulationSequence(;
    models = models,
    ini_cond_chronology = InterProblemChronology(),
)
sim = Simulation(;
    name = "pf_source_repro",
    steps = 2,
    models = models,
    sequence = sequence,
    simulation_folder = mktempdir(; cleanup = true),
)

@info "Building simulation..."
build_out = build!(sim; console_level = Logging.Warn)
@assert build_out == PSI.SimulationBuildStatus.BUILT "Build failed: $build_out"

@info "Executing simulation..."
exec_out = execute!(sim)
@assert exec_out == PSI.RunStatus.SUCCESSFULLY_FINALIZED "Execute failed: $exec_out"

# ── compare PTDFBranchFlow vs PowerFlowBranchActivePowerFromTo ──
sim_results = SimulationResults(sim)
uc_results  = get_decision_problem_results(sim_results, "UC")

ptdf_flows = read_realized_expression(
    uc_results,
    "PTDFBranchFlow__Line";
    table_format = PSI.TableFormat.WIDE,
)
pf_flows = read_realized_aux_variable(
    uc_results,
    "PowerFlowBranchActivePowerFromTo__Line";
    table_format = PSI.TableFormat.WIDE,
)

@assert target_line_name ∈ names(ptdf_flows) "Line '$target_line_name' missing from PTDFBranchFlow"
@assert target_line_name ∈ names(pf_flows) "Line '$target_line_name' missing from PowerFlowBranchActivePowerFromTo"

ptdf_vals = ptdf_flows[!, target_line_name]
pf_vals   = pf_flows[!, target_line_name]
diff      = abs.(ptdf_vals .- pf_vals)

println("\n=== Line '$target_line_name' (nodeA → nodeB) flow comparison ===")
println("PTDFBranchFlow:                   ", ptdf_vals)
println("PowerFlowBranchActivePowerFromTo: ", pf_vals)
println("|diff|:                           ", diff)
println("max |diff|: $(maximum(diff))  (expected ≈ 0.0)")

if maximum(diff) < 1e-3
    @info "PASS: PTDF and DC-PF flows agree (max |diff| = $(maximum(diff)))"
else
    @warn "FAIL: PTDF and DC-PF flows disagree" max_diff=maximum(diff) ptdf=ptdf_vals pf=pf_vals
end
```

luke-kiernan:
Address issue #1623: `ActivePowerIn` parameters have _negative_ values. Will need the same fix in IOM and/or POM.

Aside: this is one place where a having dedicated type for flow direction could help with code clarity.


jd-lara (changes requested):
This solution is pretty bad. We should be using the same construction as with the storage and pump hydro

luke-kiernan:
> This solution is pretty bad. We should be using the same construction as with the storage and pump hydro

I took a look at storage and hydro. As far as I can tell, they don't have this edge case where there's a time series parameter whose multiplier encodes direction. Storage devices have separate `{input/output}_active_power_limits`, while sources have just one `active_power_limits` that controls net (so negative minimum).

For variables, `get_variable_upper_bound` holds magnitude, while `get_variable_multiplier` holds sign. For parameters, `get_multiplier value` serves **both** roles: the signs are baked into the muliplier array, not determined by a separate dispatch. That's the fundamental problem here.

Options:
1. Change that design and split sign vs magnitude for parameters too. Would require significant refactoring to the `add_to_expression` machinery. [But maybe that should happen anyway...]
2. Leave parameter multipliers as-is. Wedge a special case into the PowerFlows.jl accounting.

Is there a middle of the road option here I'm overlooking?

jd-lara:
> > This solution is pretty bad. We should be using the same construction as with the storage and pump hydro
> 
> I took a look at storage and hydro. As far as I can tell, they don't have this edge case where there's a time series parameter whose multiplier encodes direction. Storage devices have separate `{input/output}_active_power_limits`, while sources have just one `active_power_limits` that controls net (so negative minimum).
> 
> For variables, `get_variable_upper_bound` holds magnitude, while `get_variable_multiplier` holds sign. For parameters, `get_multiplier value` serves **both** roles: the signs are baked into the muliplier array, not determined by a separate dispatch. That's the fundamental problem here.
> 
> Options:
> 
> 1. Change that design and split sign vs magnitude for parameters too. Would require significant refactoring to the `add_to_expression` machinery. [But maybe that should happen anyway...]
> 2. Leave parameter multipliers as-is. Wedge a special case into the PowerFlows.jl accounting.
> 
> Is there a middle of the road option here I'm overlooking?

This statement "As far as I can tell, they don't have this edge case where there's a time series parameter whose multiplier encodes direction." seems like a read flag on how the parameters are being used in the sources. I'll need to check with @m-bossart to understand what's going on

PR #1625, opened 2026-06-05
perf improvements

jd-lara:
This PR introduces a couple of small changes to handle the problem export and input data validation. These changes come after the feedback to flexibilize the exporting formats for the models


luke-kiernan (changes requested):
My 2 cents: why is `export_optimization_model` a string, not a scoped enum? Probably don't need a full abstract supertype with 3 separate subtypes--not many multiple dispatch opportunities here--but enum is nice in that it's self-documenting and correctness-enforcing.

luke-kiernan:
Requires PF PR [388](https://github.com/Sienna-Platform/PowerFlows.jl/pull/388), otherwise environment stuff fails.

PR #1633, opened 2026-06-14
sc slacks

jd-lara:
This PR solves an issue reported by @ChenHaotianC that the current code doesn't add the slacks to the post contingency flow constraints


test/test_device_branch_constructors.jl
```diff
@@ -925,3 +925,64 @@ end
...
+    model_default, _, status_default = _build_zib_monitored_line(false)
+    @test status_default == PSI.ModelBuildStatus.BUILT
+    @test !haskey(PSI.get_template(model_default).branches, :MonitoredLine)
+    container_default = PSI.get_optimization_container(model_default)
+    @test !PSI.has_container_key(
+        container_default,
+        PSI.FlowRateConstraint,
+        MonitoredLine,
+    )
+end
```
m-bossart, line 989: Add a test where `"model_all_branches" = false`, and there are multiple monitored lines in the system, only one of which is reduced due to being close to zero impedance. I believe that case will still fail with an un-informative error (won't find the arc key in the PTDF when building the matrices). We should at least try to catch this and give an informative message and suggest to the user that they can set the keyword if they want to force all of the monitored lines to be explicitly modeled.

m-bossart (changes requested):

src/core/network_model.jl
```diff
@@ -402,6 +431,84 @@ function _push_component_buses!(buses::Set{Int64}, device::PSY.StaticInjection)
...
+)
+    removed_arcs = PNM.get_removed_arcs(network_model.network_reduction)
+    isempty(removed_arcs) && return
+    for m in values(branch_models)
+        _warn_reduced_monitored_lines!(removed_arcs, m)
+    end
+    return
+end
+
+_warn_reduced_monitored_lines!(removed_arcs, ::DeviceModel) = nothing
```
m-bossart:
```suggestion
_warn_reduced_monitored_lines!(removed_arcs::Set{Tuple{Int, Int}}, ::DeviceModel) = nothing
```

src/core/network_model.jl
```diff
@@ -402,6 +431,84 @@ function _push_component_buses!(buses::Set{Int64}, device::PSY.StaticInjection)
...
+
+function _warn_reduced_monitored_lines!(removed_arcs, m::DeviceModel{PSY.MonitoredLine})
```
m-bossart:
```suggestion
function _warn_reduced_monitored_lines!(removed_arcs::Set{Tuple{Int, Int}}, m::DeviceModel{PSY.MonitoredLine})
```

m-bossart (approved):

PR #1636, opened 2026-06-19
Fix lossless HVDC double-counted in power-flow-in-the-loop injections

luke-kiernan:
## Problem

When an AC `power_flow_evaluation` runs in the loop on a model containing an `HVDCTwoTerminalLossless` line, the post-solve branch flows diverged sharply from the optimization's own flow variables (observed >1000 MW errors concentrated on inter-area tie lines).

## Cause

`HVDCTwoTerminalLossless` exposes a **single** `FlowActivePowerVariable`, wired to both the `pst_from_to` and `pst_to_from` categories. The pf-data writer subtracted that one variable at **both** terminals (`-F` at the from bus and `-F` at the to bus) instead of `-F / +F`. This injected a large same-sign phantom term (`≈ -2F`) that the AC power flow slack then had to absorb, flooding the lines near the slack. (The `HVDCTwoTerminalDispatch` model was correct, because its two loss-coupled variables `ft + tf = losses` are independently signed.)

## Fix

- Route HVDC terminal injections into `bus_hvdc_net_power` (mirroring how PowerFlows accumulates native HVDC setpoints) instead of `bus_active_power_injections`.
- Thread the variable (entry) type into `_update_pf_data_component!` so the to-bus sign is formulation-aware: `+= value` for the lossless single `FlowActivePowerVariable`, `-= value` for the dispatch `FlowActivePowerToFromVariable`. From-bus and PST handling unchanged.

Routing into `bus_hvdc_net_power` relies on `clear_injection_data!` zeroing that field each solve (Sienna-Platform/PowerFlows.jl#395).

## Tests

- Updated the existing dispatch in-the-loop test to read `bus_hvdc_net_power` and assert no leak into `bus_active_power_injections`.
- Added a lossless regression test (RTS, single flow variable) asserting `-F` / `+F` terminals that net to zero.

## Verification

On a UC(PTDF)→ED(AC OPF) RTS simulation with an HVDC line, max |ACOPF flow − post-solve PF flow| dropped from **1275.6 MW → 3.03 MW** (residual is genuine AC reconciliation), and the net injection handed to the slack dropped from 27.3 pu to 2.57 pu (system losses).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/network_models/power_flow_evaluation.jl
+++ b/src/network_models/power_flow_evaluation.jl
@@ -448,6 +448,18 @@ function add_power_flow_data!(
 end
 
 # How to update the PowerFlowData given a component type. A bit duplicative of code in PowerFlows.jl.
+# Generic forwarder: most categories do not depend on which variable supplied the value, so
+# the entry (variable) type is dropped here. Only the HVDC writers below specialize on it, to
+# distinguish the lossless single-flow variable from the dispatch model's signed From/To pair.
+_update_pf_data_component!(
+    pf_data::PFS.PowerFlowData,
+    category::Val,
+    comp_type::Type,
+    ::Type,
+    index::Int,
+    t::Int,
+    value::Float64,
+) = _update_pf_data_component!(pf_data, category, comp_type, index, t, value)
 _update_pf_data_component!(
     pf_data::PFS.PowerFlowData,
     ::Val{:active_power},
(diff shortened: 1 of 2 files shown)


jd-lara:
closing in favor of https://github.com/Sienna-Platform/PowerSimulations.jl/pull/1637

PR #1637, opened 2026-06-19
pf and print bug fixes

jd-lara:


src/network_models/power_flow_evaluation.jl
```diff
@@ -447,163 +459,164 @@ function add_power_flow_data!(
...
-_update_pf_data_component!(
-    pf_data::PFS.PowerFlowData,
+# Injection-sign resolver: the single source of truth mapping an optimization value to its
+# power-flow contribution. `sign` equals the multiplier `add_to_expression!` applied to the bus
+# balance, so PF reproduces the OPF nodal balance. Two thin writers consume it (below).
+#   quantity : :active | :reactive | :angle | :magnitude
+#   role     : active/reactive array selector :injection | :withdrawal | :hvdc_net (:none for voltage)
+#   sign     : nodal-balance multiplier (+1 / -1)
+#   partial  : System writer only — in/out variables accumulate onto a shared active_power field
+struct PFContribution
```
luke-kiernan: Replace `Symbol` with enums or singleton structs? 1-off use, but would resolve CoPilot's below issue by making `PFContribution` `isbits`.

src/network_models/power_flow_evaluation.jl
```diff
@@ -612,29 +625,35 @@ function _write_value_to_pf_data!(
     key::OptimizationContainerKey,
     component_map)
     result = lookup_value(container, key)
-    presigned = _pf_input_presigned(key)
+    c = pf_contribution(Val(category), get_entry_type(key), get_component_type(key))
```
luke-kiernan: Might be worth having claude do a brief type stability audit on this function. Things that jump out at me:
1. `get_entry_type(key)` and `get_component_type(key)`: replace with type parameters
2. If you make `category` multiple-dispatch-able, `Val(category)` becomes redundant.
3. (1) and (2) above make the `c =` value knowable from the types of the inputs to `_write_value_to_pf_data!`, so then `Val(c.foo)` probably also becomes redundant.

luke-kiernan (changes requested):
Much improved. 2 considerations about Julia-ness and type stability, but otherwise looks good.

src/network_models/power_flow_evaluation.jl
```diff
@@ -447,163 +450,186 @@ function add_power_flow_data!(
...
+#   quantity : :active | :reactive | :angle | :magnitude
+#   role     : :injection | :withdrawal | :hvdc_net | :voltage  (selects the PowerFlowData array)
+#   sign     : the nodal-balance multiplier (+1 / -1) applied to the value
+#   partial  : true for in/out variables, which accumulate onto a shared System active_power field
+struct PFContribution
+    quantity::Symbol
+    role::Symbol
+    sign::Float64
+    partial::Bool
+end
```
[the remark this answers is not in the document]
acostarelli: I tested the following
```julia
struct Foo
    bar::Symbol
end
@assert Foo(:a) == Foo(:a)
@assert Foo(:a) === Foo(:a)
@assert Foo(:a) != Foo(:b)
@assert Foo(:a) !== Foo(:b)
```

src/network_models/power_flow_evaluation.jl
acostarelli: I think this Copilot might be wrong, I replied to its comment.

luke-kiernan (approved):
This code is one of the major pain points in PSI (second only to the MBC tests) imo. The self-documenting struct, more straightforward code flow are a big step in the right direction

PR #1639, opened 2026-06-22
pass system to add expression call

rodrigomha:
Needed for methods in Decomposition. @jd-lara We should discuss this.


jd-lara:
This looks more like a decomposition specific method, not a general one.

rodrigomha:
Not needed as we have added a constructor in Decomposition directly.

PR #1648, opened 2026-07-13
powerflows reactive multiperiod

jd-lara:
Adds support for the multi period solve of HVDC and reactive power controls back into PSI for AC PowerFlow in the loop. requires https://github.com/Sienna-Platform/PowerFlows.jl/pull/413


src/network_models/power_flow_evaluation.jl
```diff
@@ -328,14 +328,126 @@ function bus_aux_vars(data::PFS.ACPowerFlowData)
...
+        return true
+    end
+    return !isempty(PFS.get_generic_hvdc_flows(data))
+end
+
+# The HVDC quantities the AC power flow solves for, one aux variable per (quantity,
+# component type), all read from `PFS.get_hvdc_results`'s per-family tables. Single source
+# of truth for registration and fill: (aux var, component type, table, name column, value
+# column, natural-units power quantity).
+const _HVDC_AUX_SPECS = (
```
luke-kiernan, line 388: Yikes the code repetition here. Add a human-readable table/list somewhere in the docs? Consolidate entries where only 2 things change somehow?

More broadly, we deal with these sorts of correspondences in a lot of places, where we have `(optimization_key, device_type, fieldname/dataframe column)`. Other places (KLU.jl) use an `@eval` loop, but that's typically for ~4 combinations, not dozens. Seems to me like there ought to be a way to build these correspondences into how things work, instead of hard-coding these relationships.

src/network_models/power_flow_evaluation.jl
```diff
@@ -328,14 +328,126 @@ function bus_aux_vars(data::PFS.ACPowerFlowData)
...
+    Dict{DataType, Vector{Tuple{DataType, String}}}()
+function device_aux_vars(pf_data::PFS.ACPowerFlowData)
+    out = Dict{DataType, Vector{Tuple{DataType, String}}}()
+    df = PFS.get_controlled_device_results(pf_data)
+    isempty(df) && return out
+    for (family, comp_type, aux_var) in (
+        ("TapTransformer", PSY.TapTransformer, PowerFlowTapRatio),
+        ("SwitchedAdmittance", PSY.SwitchedAdmittance, PowerFlowSwitchedShuntSusceptance),
+        ("FACTSControlDevice", PSY.FACTSControlDevice, PowerFlowFACTSReactivePower),
+    )
```
luke-kiernan, line 484: Have you checked type stability of this? `comp` changes from `TapTransformer` to `SwitchedAdmittance`.

src/network_models/power_flow_evaluation.jl
```diff
@@ -438,6 +552,17 @@ function add_power_flow_data!(
...
         end
+
+        for aux_var_map in (device_aux_vars(pf_data), hvdc_aux_vars(pf_data, sys))
+            for (device_aux_var, components) in aux_var_map
+                to_add_to = get!(
+                    device_aux_var_components,
+                    device_aux_var,
+                    Set{Tuple{<:DataType, String}}(),
+                )
+                push!.(Ref(to_add_to), components)
```
luke-kiernan: > `push!.(Ref(to_add_to), components)`

Is there a reason why this isn't `union!(to_add_to, components)`? Would be more idiomatic and performant. Same in the other 2 spots above too.

src/core/auxiliary_variables.jl
```diff
@@ -77,16 +77,137 @@ struct PowerFlowBranchActivePowerLoss <: PowerFlowAuxVariableType end
...
-convert_result_to_natural_units(
-    ::Type{
-        <:Union{
-            PowerFlowBranchReactivePowerFromTo, PowerFlowBranchReactivePowerToFrom,
-            PowerFlowBranchActivePowerFromTo, PowerFlowBranchActivePowerToFrom,
-            PowerFlowBranchActivePowerLoss,
-        },
-    },
-) = true
+convert_result_to_natural_units(::Type{PowerFlowBranchReactivePowerFromTo}) = true
```
luke-kiernan: Consider dispatching on super types: `PowerFlowHVDCAuxVariableType` or `BranchFlowAuxVariableType`.

src/core/auxiliary_variables.jl
```diff
@@ -77,16 +77,137 @@ struct PowerFlowBranchActivePowerLoss <: PowerFlowAuxVariableType end
...
+Auxiliary Variable for the reactive power (p.u.) delivered by each `FACTSControlDevice` in
+power-flow discrete control, per power-flow time step.
+"""
+struct PowerFlowFACTSReactivePower <: PowerFlowAuxVariableType end
+
+"""
+Auxiliary Variable for the net HVDC active power (p.u.) at each bus as used by the power flow
+evaluation, per time step. Registered only when the system carries HVDC components.
+"""
+struct PowerFlowHVDCNetPower <: PowerFlowAuxVariableType end
```
luke-kiernan, line 109: Why isn't this a sub type of `PowerFlowHVDCAuxVariableType`?

src/core/auxiliary_variables.jl
```diff
@@ -77,16 +77,137 @@ struct PowerFlowBranchActivePowerLoss <: PowerFlowAuxVariableType end
...
+Parent of the per-component HVDC quantities solved by the AC power flow (VSC/LCC lines, MTDC
+converters and DC branches), read from `PowerFlows.get_hvdc_results`.
+"""
+abstract type PowerFlowHVDCAuxVariableType <: PowerFlowAuxVariableType end
+
+"""
+Auxiliary Variable for the active power (p.u.) entering an HVDC line at its from terminal, from
+the AC power flow solution.
+"""
+struct PowerFlowHVDCActivePowerFromTo <: PowerFlowHVDCAuxVariableType end
```
luke-kiernan, line 121: Could introduce an intermediate super type for HVDC line flows, an HVDC equivalent of `BranchFlowAuxVariableType`. Organizationally, I'm not sure if that helps (groups together like things) or hurts (yet another type), though.

src/network_models/power_flow_evaluation.jl
```diff
@@ -1173,6 +1301,112 @@ _get_pf_result(::Type{PowerFlowBranchActivePowerLoss}, pf_data::PFS.PowerFlowDat
     PFS.get_arc_active_power_flow_from_to(pf_data) .+
     PFS.get_arc_active_power_flow_to_from(pf_data)
 
+# Solved discrete-control device settings, read from the per-time-step results table.
+# Concrete key types: the TapTransformer method must out-dispatch the generic
+# `AuxVarKey{T, U <: PSY.Branch}` arc-flow method below.
```
luke-kiernan: Could `@eval`-loop these 4 nearly identical definitions that are separate purely for dispatch tie breaking.

scripts/verify_pf_reactive_control_psi.jl
```diff
@@ -0,0 +1,468 @@
...
+        control_discrete_devices = control_discrete_devices,
+        check_reactive_power_limits = check_reactive_power_limits,
+    )
+end
+
+"RTS has HydroDispatch/HydroTurbine/HydroReservoir/EnergyReservoirStorage generation that
+`template_economic_dispatch`'s defaults don't cover; leaving them unmodeled starves the
+balance of real capacity and the UC becomes infeasible (verified directly against baseline
+RTS with no reactive-control augmentation). ThermalBasicUnitCommitment (not the no-commitment
+Dispatch variant) mirrors the RTS AC-PF-in-the-loop idiom in test_power_flow_in_the_loop.jl."
```
luke-kiernan: Hmmm looks like AI ran into some unexpected behavior and crafted a workaround. Does its explanation make good sense from a power systems modeling perspective? Or is their a bug/modeling flaw lurking here?

luke-kiernan (changes requested):
One big picture comment: we now have enough children of `PowerFlowAuxVariableType` that the hierarchy is starting to get confusing. Ideas: add a tree (or other graphical depiction) to the docs and/or restructure the hierarchy so that HVDC mirrors non-HVDC more closely.

jd-lara:
### Code review

Found 1 issue:

1. Dead guard in the 3-arg `calculate_aux_variable_value!` fallback: `key` is an `AuxVarKey` *instance*, but `branch_aux_vars(pf_data)`/`bus_aux_vars(pf_data)` return vectors of *types* (`Vector{DataType}`), so `key in branch_aux_vars(pf_data) || key in bus_aux_vars(pf_data)` is always `false` and the "skip the aux vars that the current power flow isn't meant to update" early-return never fires. Since `calculate_aux_variables!` re-runs every PF aux-var key after each evaluator's solve and this shim resolves `latest_solved_power_flow_evaluation_data`, a template with two numeric PF evaluators recomputes every key against whichever evaluator solved last — e.g. an AC-only key like `PowerFlowVoltageMagnitude` gets overwritten from a PTDF evaluator's data. Single-evaluator templates are unaffected, which is why the current tests pass. Note the fix needs more than `get_entry_type(key)` plus inverting the polarity: device-control and HVDC aux vars appear in neither `branch_aux_vars` nor `bus_aux_vars`, so a naive "not in either list → skip" would silently disable the tap/shunt/FACTS/HVDC ingestion this PR adds.

https://github.com/Sienna-Platform/PowerSimulations.jl/blob/e422ed5fc617610ad5f5a69ce69ddb506c12138f/src/network_models/power_flow_evaluation.jl#L1451-L1462

(bug due to `branch_aux_vars` returning types, not keys: https://github.com/Sienna-Platform/PowerSimulations.jl/blob/e422ed5fc617610ad5f5a69ce69ddb506c12138f/src/network_models/power_flow_evaluation.jl#L306-L318)

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/network_models/power_flow_evaluation.jl
```diff
@@ -1452,10 +1479,10 @@ end
 function calculate_aux_variable_value!(container::OptimizationContainer,
     key::AuxVarKey{<:PowerFlowAuxVariableType, <:PSY.Component},
     system::PSY.System)
-    # Skip the aux vars that the current power flow isn't meant to update
     pf_e_data = latest_solved_power_flow_evaluation_data(container)
     pf_data = get_power_flow_data(pf_e_data)
-    (key in branch_aux_vars(pf_data) || key in bus_aux_vars(pf_data)) && return
-    calculate_aux_variable_value!(container, key, system, pf_e_data)
+    if _pf_provides_aux_var(get_entry_type(key), pf_data)
```
luke-kiernan: > get_entry_type(key)

Could replace with type parameter

src/network_models/power_flow_evaluation.jl
```diff
@@ -345,6 +345,33 @@ bus_aux_vars(::PFS.PTDFPowerFlowData) = DataType[]
 bus_aux_vars(::PFS.vPTDFPowerFlowData) = DataType[]
 bus_aux_vars(::PFS.PSSEExporter) = DataType[]
 
+# The discrete-control and HVDC per-component aux vars are registered only from AC
```
luke-kiernan, line 348: The mixing of trait patterns (`_provides_control_aux_vars`) and explicit lists (`{branch,bus}_aux_vars`) here is mildly weird and confusing. Might be clearer to pick one and stick with it.

luke-kiernan (approved):
2 small comments, otherwise looks good

scripts/verify_pf_reactive_control_psi.jl
```diff
@@ -0,0 +1,521 @@
+#! /usr/bin/env julia
```
m-bossart: Is this script intended to be committed?

m-bossart (approved):

PR #1656, opened 2026-08-14
Fail the join of a partitioned simulation if any partition failed

daniel-thom:
_check_jobs read the status of partition 1 for every index, so a partition that failed gracefully was never detected. The join then merged that partition's pre-allocated store and stamped the merged results SUCCESSFULLY_FINALIZED.

Impact:
- If a constituent simulation failed ungracefully because of OOM or walltime timeout, the simulation store file would have been corrupted and the join operation would have always failed. The rest of the impact section is not applicable.
- If a constituent simulation failed with a solver error, the joined simulation store file would have had holes of 0s in the result data for the failed days.
- The reference implementation for running via Distributed.jl threw an exception on any failure, so users of this workflow should have seen the failure.
- Users that ran partitioned simulations on an HPC with Jade would have seen the errors if they followed the standard procedures of checking workflow/job status, where the errors were clearly reported.
- Users that ran partitioned simulations on an HPC with Torc using the reference workflow implementation would have observed the join job not running at all - it would have been canceled because of the other failure(s), which torc would have clearly reported.

Read the status of each partition and, by default, record RunStatus.FAILED in the status file of the joined simulation and raise an error without merging. Add a skip_failures option (--skip-failures on the command line) that logs each failed partition, along with the simulation steps that it covers, and merges the results of the successful partitions. The status of the joined simulation is a failure either way.

Compute the destination offset of each dataset merge from the partition's absolute step range instead of the running write position so that a skipped partition does not shift the data of the partitions that follow it. Handle store files that are missing or corrupted the same way as failed jobs.

Support valueless flags in the command-line arguments.

### DIFF
--- a/src/simulation/simulation_partition_results.jl
+++ b/src/simulation/simulation_partition_results.jl
@@ -29,19 +29,49 @@ function SimulationPartitionResults(path::AbstractString)
 end
 
 """
-Combine all partition simulation files.
+Combine all partition simulation files and return the status of the joined simulation.
+
+Throw an exception if any partition job failed, unless `skip_failures` is `true`.
+
+# Arguments
+
+  - `path::AbstractString`: Directory of the main simulation.
+  - `skip_failures::Bool`: If `true`, log and skip the store files of the partition jobs
+    that failed and merge the results of the successful jobs. The status of the joined
+    simulation is `RunStatus.FAILED` whenever any partition job failed, regardless of this
+    setting.
 """
-function join_simulation(path::AbstractString)
+function join_simulation(path::AbstractString; skip_failures = false)
     results = SimulationPartitionResults(path)
-    join_simulation(results)
-    return
+    return join_simulation(results; skip_failures = skip_failures)
 end
 
-function join_simulation(results::SimulationPartitionResults)
-    status = _check_jobs(results)
-    _merge_store_files!(results)
+function join_simulation(results::SimulationPartitionResults; skip_failures = false)
+    failed_partitions = _check_jobs(results)
+    if !isempty(failed_partitions) && !skip_failures
+        _serialize_failed_status(results)
+        error(
+            "These partition jobs were not successful: $(first.(failed_partitions)). " *
+            "Refer to the log messages above for the affected simulation steps. " *
+            "Pass skip_failures = true (--skip-failures on the command line) to skip the " *
+            "failed jobs and merge the results of the successful jobs.",
+        )
+    end
+
+    not_merged = try
+        _merge_store_files!(results, Set{Int}(first.(failed_partitions)), skip_failures)
+    catch
+        _serialize_failed_status(results)
+        rethrow()
+    end
+
+    status = if isempty(not_merged)
+        RunStatus.SUCCESSFULLY_FINALIZED
+    else
+        RunStatus.FAILED
+    end
     _complete(results, status)
-    return
+    return status
 end
 
 function _partition_path(x::SimulationPartitionResults, i)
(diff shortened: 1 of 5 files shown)


src/simulation/simulation_partition_results.jl
```diff
@@ -20,28 +37,62 @@ function SimulationPartitionResults(path::AbstractString)
...
+    return join_simulation(results; skip_failures = skip_failures)
 end
 
-function join_simulation(results::SimulationPartitionResults)
-    status = _check_jobs(results)
-    _merge_store_files!(results)
+function join_simulation(results::SimulationPartitionResults; skip_failures = false)
+    failed_partitions = _check_jobs(results)
+    if !isempty(failed_partitions) && !skip_failures
+        _try_serialize_failed_status(joinpath(results.path, RESULTS_DIR))
```
jd-lara, line 64: Should this error message be more descriptive of the consequences and results quality?

daniel-thom: What would you suggest?

This is the normal, automated path where the user does not pass `--skip-failures`. If the user reruns with that flag, they will receive one of
```
                @warn "Skip the store file of the failed partition job index = $i. " *
                      "Results for steps = $(_valid_step_range(results, i)) will be " *
                      "invalid in the merged store."
```
```
                @error "Failed to open the store file of partition job index = $i. " *
                       "The file is missing or corrupted. Results for " *
                       "steps = $(_valid_step_range(results, i)) will be invalid in " *
                       "the merged store." exception = (e, catch_backtrace())
```

src/simulation/simulation_partition_results.jl
```diff
@@ -58,39 +109,88 @@ end
...
+)
+    not_merged = Int[]
     HDF5.h5open(_store_path(results), "r+") do dst
         for i in 1:get_num_partitions(results.partitions)
-            HDF5.h5open(joinpath(_partition_path(results, i), _store_subpath()), "r") do src
-                _copy_datasets!(results, i, src, dst)
+            if i in skip_indexes
+                @warn "Skip the store file of the failed partition job index = $i. " *
+                      "Results for steps = $(_valid_step_range(results, i)) will be " *
+                      "invalid in the merged store."
```
jarry7, line 171: Do we just merge with zeros for these failed partitions then? Or `nothing`?   - so that results from `n` partitions are still presented in joined store but with gaps for those failed partitions.

src/simulation/simulation_partition_results.jl
```diff
@@ -101,91 +201,112 @@ function _copy_datasets!(
...
-    return
+    IS.@assert_op src_size % length(step_range) == 0
+    per_step = src_size ÷ length(step_range)
+    # Guarantees that the absolute destination offsets computed below are in bounds and
+    # consistent across all partitions.
+    IS.@assert_op per_step * results.partitions.num_steps == dst_size
+    # Compute both offsets from the same absolute range of valid steps so that they
+    # cannot drift apart, and compute the destination offset from absolute steps rather
+    # than from a running write position so that a skipped (corrupted) partition does
+    # not shift the data of subsequent partitions.
```
jarry7, line 269: I think this is what I was asking about on `join` (and how failed partitions are treated)

src/simulation/simulation_partitions.jl
```diff
@@ -156,23 +196,44 @@ function process_simulation_partition_cli_args(build_function, execute_function,
...
+        status = join_simulation(
+            base_dir;
+            skip_failures = get_bool_option(options, "skip-failures"),
+        )
+        if status != RunStatus.SUCCESSFULLY_FINALIZED
+            error(
+                "The results of the successful partition jobs were merged, but the " *
+                "status of the joined simulation is $status because one or more " *
+                "partition jobs failed.",
+            )
```
jarry7, line 209: Could we not indicate a `RunStatus.SUCCESSFULLY_FINALIZED` if the user opted into skipping failures?

jarry7 (commented):
Left a few comments (mostly for clarity) - but I think this is what I would intend the partitions to be treated as (+ old bug is addressed).

src/simulation/simulation_partition_results.jl
daniel-thom: The merged store will have zeros. We could likely initialize with NaN if that is desired.

jd-lara (approved):
Looks good, I checked the join at the end it seems ok

src/simulation/simulation_partitions.jl
daniel-thom: We could, but then it wouldn't really be accurate. We could consider a third state indicating partial success.

jd-lara: Partial Sucess is confussing, how do you indicate what worked and what didn't. I have concerns about people have too little friction to open partially successful simulations.

PR #1660, opened 2026-08-31
Fix DeviceStatus initial condition binary normalization

### LINKED ISSUE #1659: Snap transferred DeviceStatus initial conditions to binary endpoints
## Describe the bug

`DeviceStatus` initial conditions copy raw `OnVariable` solver values. In the simulation-state path, [`update_initial_conditions!`](https://github.com/Sienna-Platform/PowerSimulations.jl/blob/aa474f9e436bd0f0ee99e3bb1ca2f8ea8cc442bf/src/simulation/initial_condition_update_simulation.jl#L147-L173) transfers the raw `Float64`; the [in-memory path](https://github.com/Sienna-Platform/PowerSimulations.jl/blob/aa474f9e436bd0f0ee99e3bb1ca2f8ea8cc442bf/src/operation/initial_conditions_update_in_memory_store.jl#L103-L109) does the same.

`OnVariable` represents binary commitment status. A normal MIP-solver result such as `0.9999997` can therefore become a fractional status initial condition and be consumed by the next first-period commitment transition equation.

The repository already rounds this value at comparable boundaries:

- `OnVariable` feedforward values use `round` before parameter assignment.
- Binary and integer dual-processing cache values are rounded to avoid MIP-solver numerical-tolerance infeasibilities.

## Expected behavior

Round transferred `OnVariable` values before storing `DeviceStatus` initial conditions in both the simulation-state and in-memory paths. Preserve the existing initial-condition setter behavior for validation. Add regression coverage for a near-one solver value through both paths.

## Additional context

`InitialCondition` values are represented as `Float64`; the desired result is an exact `0.0` or `1.0` value, not a storage-type change.

pesap:
## Summary

- Snap `DeviceStatus` at its central initial-condition write using type-specific dispatch.
- Leave simulation-state and in-memory-store reads and raw solver results unchanged.
- Add regression coverage for both update sources and a JuMP-backed initial condition.

## Validation

- `julia --project=scripts/formatter -e 'include("scripts/formatter/formatter_code.jl")'`
- Syntax parse of the changed Julia files
- `round(0.9999997) === 1.0`
- Targeted test execution is blocked locally by the available Sienna dependency set. CI is running.

Closes #1659

### DIFF
--- a/src/initial_conditions/calculate_initial_condition.jl
+++ b/src/initial_conditions/calculate_initial_condition.jl
@@ -1,3 +1,24 @@
+# MIP solver tolerances can leave binary values near their endpoints.
+function set_ic_quantity!(
+    ic::InitialCondition{DeviceStatus, JuMP.VariableRef},
+    var_value::Float64,
+)
+    @assert isfinite(var_value) ic
+    fix_parameter_value(ic.value, round(var_value))
+    return
+end
+
+function set_ic_quantity!(
+    ic::InitialCondition{DeviceStatus, Float64},
+    var_value::Float64,
+)
+    @assert isfinite(var_value) ic
+    @debug "Initial condition value set with Float64. Won't update the model until rebuild" _group =
+        LOG_GROUP_BUILD_INITIAL_CONDITIONS
+    ic.value = round(var_value)
+    return
+end
+
 """
 Default implementation of set_initial_condition_value
 """
--- a/src/initial_conditions/calculate_initial_condition.jl
+++ b/src/initial_conditions/calculate_initial_condition.jl
@@ -1,4 +1,13 @@
-# MIP solver tolerances can leave binary values near their endpoints.
+"""
+    set_ic_quantity!(
+        ic::InitialCondition{DeviceStatus, JuMP.VariableRef},
+        var_value::Float64,
+    )
+
+Set the `DeviceStatus` initial condition by rounding the finite solver value
+`var_value` before fixing `ic`. This removes small deviations from binary
+endpoints caused by MIP solver tolerances.
+"""
 function set_ic_quantity!(
     ic::InitialCondition{DeviceStatus, JuMP.VariableRef},
     var_value::Float64,
@@ -8,6 +17,16 @@ function set_ic_quantity!(
     return
 end
 
+"""
+    set_ic_quantity!(
+        ic::InitialCondition{DeviceStatus, Float64},
+        var_value::Float64,
+    )
+
+Set the `DeviceStatus` initial condition by rounding the finite solver value
+`var_value` before storing it in `ic`. This removes small deviations from
+binary endpoints caused by MIP solver tolerances.
+"""
 function set_ic_quantity!(
     ic::InitialCondition{DeviceStatus, Float64},
     var_value::Float64,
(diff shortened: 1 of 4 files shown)


src/simulation/initial_condition_update_simulation.jl
```diff
@@ -167,7 +167,7 @@ function update_initial_conditions!(
     for ic in ics
         isnothing(get_value(ic)) && continue
         var_val = get_system_state_value(state, OnVariable(), get_component_type(ic))
-        set_ic_quantity!(ic, var_val[get_component_name(ic)])
+        set_ic_quantity!(ic, round(var_val[get_component_name(ic)]))
```
daniel-thom: Should we perform the round operation whenever this value is set?

jarry7: I think so (as it is uncontrolled as to when it may happen).

In any case - this also only happens when a solver Integer Gap to hand-off the `ic_quantity` is sufficiently large i.e. the solver misses the rounding.

src/initial_conditions/calculate_initial_condition.jl
```diff
@@ -1,3 +1,24 @@
+# MIP solver tolerances can leave binary values near their endpoints.
```
jarry7: Probably need a more structured docstring + formatting

src/initial_conditions/calculate_initial_condition.jl
```diff
@@ -1,3 +1,24 @@
...
+function set_ic_quantity!(
+    ic::InitialCondition{DeviceStatus, JuMP.VariableRef},
+    var_value::Float64,
+)
+    @assert isfinite(var_value) ic
+    fix_parameter_value(ic.value, round(var_value))
+    return
+end
+
+function set_ic_quantity!(
```
jarry7, line 30: Need a docstring

src/simulation/initial_condition_update_simulation.jl
pesap: Agreed. The normalization now happens at the central `DeviceStatus` `set_ic_quantity!` write path for both `Float64` and `JuMP.VariableRef`, so both update sources use it while raw result stores remain unchanged. Resolved in https://github.com/Sienna-Platform/PowerSimulations.jl/commit/76ce6d734953bcf26a715ae78d62aff865caa477.

src/initial_conditions/calculate_initial_condition.jl
pesap: Added structured Julia docstrings for both `DeviceStatus` setter methods, including their signatures, rounding behavior, and solver-tolerance rationale. Resolved in https://github.com/Sienna-Platform/PowerSimulations.jl/commit/b1e7f0eaede5c7459d7187adb2621312546bbb72.

src/initial_conditions/calculate_initial_condition.jl
pesap: Added the requested method docstring in https://github.com/Sienna-Platform/PowerSimulations.jl/commit/b1e7f0eaede5c7459d7187adb2621312546bbb72.

jd-lara (approved):
The change looks good

PR #1663, opened 2026-09-04
Fix/area balance dual dispatch

ymiftah:
## Summary

`NetworkModel(AreaBalancePowerModel; duals = [CopperPlateBalanceConstraint])` and `NetworkModel(AreaPTDFPowerModel; duals = [CopperPlateBalanceConstraint])` both build successfully and then fail during `solve!`.

**Cause.** Both area-aggregated models store `CopperPlateBalanceConstraint` keyed by **`PSY.Area`** — `AreaBalancePowerModel` in `src/network_models/area_balance_model.jl`, `AreaPTDFPowerModel` in `src/network_models/copperplate_model.jl:23`. Neither had a matching `add_constraint_dual!` method, so each fell through to a *different* wrong one:

| Formulation | Supertype | Inherited dual method | Dual keyed by | Constraint keyed by |
|---|---|---|---|---|
| `AreaBalancePowerModel` | `PM.AbstractActivePowerModel` | generic `NetworkModel{<:PM.AbstractPowerModel}` | `PSY.ACBus` | `PSY.Area` |
| `AreaPTDFPowerModel` | `AbstractPTDFModel` | `Union{CopperPlatePowerModel, AbstractPTDFModel}` | `PSY.System` (ref buses) | `PSY.Area` |

Either way the dual container matches no stored constraint, and `_calculate_dual_variable_value!` throws:

```
InvalidValue("constraint CopperPlateBalanceConstraint__ACBus is not stored.
  [..., :CopperPlateBalanceConstraint__Area, ...]")

InvalidValue("constraint CopperPlateBalanceConstraint__System is not stored.
  [..., :CopperPlateBalanceConstraint__Area, ...]")
```

## Fix

One `add_constraint_dual!` method per formulation, plus a shared `assign_dual_variable!` that keys the container by `PSY.Area`. No existing lines are modified.

## Notes

1. I found it rather unintuitive that the `CopperPlateBalanceConstraint` needs to be used to dispatch the duals, even as the `AreaBalancePowerModel` or `AreaPTDFPowerModel` formulations are set. It is possibly a gap in the type hierarchy, which I have not explored further. There seem to exist a `NodalBalanceActiveConstraint` so we could imagine introducing a `AreaBalanceConstraint` for both the `AreaBalancePowerModel` and `AreaPTDFPowerModel`.

2. A possible alternative could dispatch the appropriate aggregation for the balance constraint formulations via a trait (c.f. https://github.com/ymiftah/PowerSimulations.jl/pull/2/changes) but that is a larger change.

## AI usage

The bug was identified with Claude Code Sonnet for the AreaBalancePowerModel, further prompting uncovered the similar issue for AreaPTDFPowerModel and I refined the proposal of the alternative "trait based" formulation.

### DIFF
--- a/src/devices_models/devices/common/add_constraint_dual.jl
+++ b/src/devices_models/devices/common/add_constraint_dual.jl
@@ -179,12 +179,13 @@ end
 
 function assign_dual_variable!(
     container::OptimizationContainer,
-    constraint_type::Type{CopperPlateBalanceConstraint}
+    constraint_type::Type{CopperPlateBalanceConstraint},
     ::U,
     ::NetworkModel{V},
 ) where {U <: PSY.System, V <: Union{AreaBalancePowerModel, AreaPTDFPowerModel}}
     time_steps = get_time_steps(container)
-    existing = get_constraint(container, constraint_type, PSY.Area)
+    existing = get_constraint(container, constraint_type(), PSY.Area)
+    @show existing
     area_names = axes(existing)[1]
     add_dual_container!(container, constraint_type, PSY.Area, area_names, time_steps)
     return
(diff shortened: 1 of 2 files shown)


jd-lara (approved):
Thanks for the fix, this makes sense

jd-lara (changes requested):
The tests needs to pass to merge

jd-lara (approved):

PR #1667, opened 2026-09-07
Fix renewable curtailment cost availability lookup

### LINKED ISSUE #1666: Curtailment-cost expression should use time-varying renewable availability
# Curtailment-cost expression should use time-varying renewable availability

## Summary

`CurtailmentCostExpression` for dispatchable renewable generators must calculate
curtailed energy from the renewable availability at each timestep, rather than
from the generator's static maximum active-power rating.

In affected PowerSimulations versions, the reported expression used the last
piecewise-linear offer breakpoint as `offer_max`. When the availability was
provided through `ActivePowerTimeSeriesParameter`, that value could resolve to
the static rating instead of the time-series availability. The result was an
overstated curtailment quantity and an overstated curtailment cost whenever the
available renewable power was below the rating.

## Affected behavior

For renewable $r$ and time period $t$, the curtailment cost should be:

$$
C_{r,t}^{\mathrm{curt}} = c_r\,\Delta t\,(P_{r,t}^{\mathrm{available}} - P_{r,t}^{\mathrm{dispatch}}).
$$

The affected implementation instead could report:

$$
C_{r,t}^{\mathrm{curt,reported}} = c_r\,\Delta t\,(P_r^{\mathrm{rating}} - P_{r,t}^{\mathrm{dispatch}}).
$$

The error for a timestep is therefore:

$$
c_r\,\Delta t\,(P_r^{\mathrm{rating}} - P_{r,t}^{\mathrm{available}}),
$$

which is positive whenever the availability time series is below the static
rating.

## Steps to reproduce

1. Create a `RenewableDispatch` component with a static maximum active power
   greater than at least one value in its `ActivePowerTimeSeriesParameter`.
2. Assign a nonzero linear `curtailment_cost`.
3. Build a model using a renewable dispatch formulation and solve a period in
   which the renewable is curtailed.
4. Read `CurtailmentCostExpression__RenewableDispatch`.
5. Compare it with the realized availability parameter and active-power
   dispatch.

For example, with a $100$ MW rating, $40$ MW availability, $10$ MW dispatch,
a one-hour resolution, and a $50/MWh curtailment penalty, the expected cost is
$50 \times (40 - 10) = \$1{,}500$. The affected expression reports
$50 \times (100 - 10) = \$4{,}500$.

## Root cause

In PowerSimulations `v0.35.0`, both `_add_curtailment_cost!` overloads obtained
`offer_max` from `_get_pwl_data(false, container, component, t)` and used its
last breakpoint. That path did not reliably represent the time-varying
availability parameter used by the renewable dispatch limit.

The corrected implementation should select the availability source in this
order:

1. When the component has an `ActivePowerTimeSeriesParameter`, use its value
   and parameter multiplier for the current timestep.
2. Fall back to `get_max_active_power(component)` only when no availability
   parameter exists for that component.

## Additional defect in the v0.38 implementation

PowerSimulations `v0.38.2` replaced the piecewise-linear breakpoint with a
time-series-first branch, but the availability test and lookup are not valid
for the parameter container's axes:

```julia
has_ts_param = name in axes(param_array)[1]

offer_max = if has_ts_param
   param_multiplier[name, t] * param_array[name, t]
else
   PSY.get_max_active_power(component)
end
```

For `TimeSeriesParameter` containers, `param_array` is indexed on its first
axis by time-series UUID, whereas `param_multiplier` is indexed by component
name. Since `name` is a component name, it is normally absent from
`axes(param_array)[1]`. Consequently, `has_ts_param` is false and the code
always takes the static-rating fallback, even when the component has the
configured `ActivePowerTimeSeriesParameter`.

The renewable active-power limit does not have this problem. Its
parameterized-constraint helper obtains the parameter through
`get_parameter_column_refs(param_container, name)`, which maps the component
name to the UUID-indexed parameter column. Therefore, a model can correctly
limit renewable dispatch to time-varying availability while
`CurtailmentCostExpression` still reports curtailment relative to the static
rating.

The curtailment-cost code should use the same accessor and test the configured
time series directly, for example:

```julia
param_container = get_parameter(
   container,
   ActivePowerTimeSeriesParameter(),
   PSY.RenewableDispatch,
)
has_ts_param = PSY.has_time_series(component, ts_type, ts_name)

if has_ts_param
   availability = get_parameter_column_refs(param_container, name)
   offer_max = param_multiplier[name, t] * availability[t]
else
   offer_max = PSY.get_max_active_power(component)
end
```

Here `ts_type` and `ts_name` must be obtained from the same model settings
used to build the parameter. The equivalent generic overload for
`PSY.RenewableGen` must be updated as well.

## Impact

- `CurtailmentCostExpression__RenewableDispatch` is incorrect for time-varying
  availability in affected versions.
- Any post-processing that aggregates this expression overstates renewable
  curtailment cost, especially for variable-output resources such as solar and
  wind.
- The expression is recorded for reporting and is not propagated to
  `ProductionCostExpression`; however, it can still mislead reports and any
  downstream analysis that consumes the expression directly.

## Expected regression test

Add a test with a renewable rating strictly above a time-series availability
value and a nonzero curtailment penalty. After solving, assert at every tested
timestep that:

$$
\texttt{CurtailmentCostExpression} = c\,\Delta t\,(P^{\mathrm{available}} - P^{\mathrm{dispatch}}).
$$

The test should include at least two distinct availability values below the
rating and should assert that the dispatch is limited by the same availability
series. Tests that only check nonnegativity or that a higher curtailment penalty
increases dispatch do not detect this reporting error. A direct unit test of
the name-to-UUID parameter lookup would also prevent this axis mismatch from
returning.

## Workaround

For results produced by an affected version, recompute curtailment cost from
realized renewable curtailment (or realized availability minus dispatch), the
model resolution, and the configured penalty. Do not rely on the stored
`CurtailmentCostExpression__RenewableDispatch` for those runs.

SebastianManriqueM:
## Summary

Fixes #1666.

`CurtailmentCostExpression` now calculates curtailed renewable energy from the configured `ActivePowerTimeSeriesParameter` for each time step, instead of falling back to the static maximum active power when a time series exists.

## Why this implementation

`TimeSeriesParameter` arrays are keyed by time-series UUID, while renewable multiplier arrays are keyed by component name. The previous availability check used the component name against the parameter array axis, so it could not find a configured time series and always selected the static-rating fallback.

The fix retrieves the `ParameterContainer`, checks its `component_name_to_ts_uuid` mapping, and obtains the parameter column through `get_parameter_column_refs(param_container, name)`. This is the existing PSI mechanism that resolves a component name to its UUID-backed parameter data and is already used by parameterized renewable limit constraints. The static `get_max_active_power(component)` fallback remains for renewables without an availability parameter.

The change is applied to both the `RenewableDispatch` and generic `RenewableGen` curtailment-cost overloads.

## Tests

- Added `add_renewable_dispatch_with_time_series!`, a documented test fixture that adds a renewable alongside the existing thermal fleet. It copies the highest-rated thermal generator's topology and static capacity, while preserving every original thermal generator.
- Added a `c_sys5` regression with a `RenewableDispatch`, a nonzero `$50/MWh` curtailment cost, and two distinct high availability values below static capacity.
- The regression uses the system's deterministic forecast resolution and interval, asserts dispatch does not exceed availability, verifies actual curtailment occurs, and compares `CurtailmentCostExpression` to `$50/MWh * (available MW - dispatched MW)`.
- The test explicitly converts the per-unit input availability to MW before comparing it with result tables, which report system-base MW.

## Validation

- Ran `julia --project=scripts/formatter -e 'include("scripts/formatter/formatter_code.jl")'`.
- Confirmed no diagnostics in the three touched Julia files.
- Confirmed `git diff --check` passes.

Commits:

1. `da7c51260` Add renewable curtailment cost regression test
2. `03a08f8e8` Use renewable availability in curtailment costs

### DIFF
--- a/src/devices_models/devices/common/objective_function/linear_curve.jl
+++ b/src/devices_models/devices/common/objective_function/linear_curve.jl
@@ -192,8 +192,8 @@ function _add_curtailment_cost!(
 
     name = PSY.get_name(component)
     dispatch_vars = get_variable(container, T(), PSY.RenewableDispatch)
-    param_array =
-        get_parameter_array(
+    param_container =
+        get_parameter(
             container,
             ActivePowerTimeSeriesParameter(),
             PSY.RenewableDispatch,
@@ -203,11 +203,15 @@ function _add_curtailment_cost!(
         ActivePowerTimeSeriesParameter(),
         PSY.RenewableDispatch,
     )
-    has_ts_param = name in axes(param_array)[1]
+    has_ts_param = haskey(get_attributes(param_container).component_name_to_ts_uuid, name)
+    availability = nothing
+    if has_ts_param
+        availability = get_parameter_column_refs(param_container, name)
+    end
 
     for t in get_time_steps(container)
         offer_max = if has_ts_param
-            param_multiplier[name, t] * param_array[name, t]
+            param_multiplier[name, t] * availability[t]
         else
             PSY.get_max_active_power(component)
         end
@@ -263,18 +267,22 @@ function _add_curtailment_cost!(
 
     name = PSY.get_name(component)
     dispatch_vars = get_variable(container, T(), PSY.RenewableGen)
-    param_array =
-        get_parameter_array(container, ActivePowerTimeSeriesParameter(), PSY.RenewableGen)
+    param_container =
+        get_parameter(container, ActivePowerTimeSeriesParameter(), PSY.RenewableGen)
     param_multiplier = get_parameter_multiplier_array(
         container,
         ActivePowerTimeSeriesParameter(),
         PSY.RenewableGen,
     )
-    has_ts_param = name in axes(param_array)[1]
+    has_ts_param = haskey(get_attributes(param_container).component_name_to_ts_uuid, name)
+    availability = nothing
+    if has_ts_param
+        availability = get_parameter_column_refs(param_container, name)
+    end
 
     for t in get_time_steps(container)
         offer_max = if has_ts_param
-            param_multiplier[name, t] * param_array[name, t]
+            param_multiplier[name, t] * availability[t]
         else
             PSY.get_max_active_power(component)
         end
--- a/test/test_device_renewable_generation_constructors.jl
+++ b/test/test_device_renewable_generation_constructors.jl
@@ -177,7 +177,7 @@ end
             rev = true,
         ),
     )
-    availability = [isodd(t) ? 0.9 : 0.8 for t in 1:24]
+    availability = [isodd(t) ? 1.6 : 1.5 for t in 1:24]
     renewable = add_renewable_dispatch_with_time_series!(
         c_sys5,
         thermal_generator,
--- a/test/test_device_renewable_generation_constructors.jl
+++ b/test/test_device_renewable_generation_constructors.jl
@@ -167,3 +167,64 @@ end
     re_vars = PSI.get_variable(container, PSI.ActivePowerVariable(), PSY.RenewableDispatch)
     @test all(JuMP.coefficient(obj, v) < 0 for v in re_vars)
 end
+
+@testset "Renewable CurtailmentCostExpression uses time-varying availability" begin
+    c_sys5 = PSB.build_system(PSITestSystems, "c_sys5")
+    thermal_generator = first(
+        sort(
+            collect(get_components(ThermalStandard, c_sys5));
+            by = generator -> get_rating(generator),
+            rev = true,
+        ),
+    )
+    availability = [isodd(t) ? 0.9 : 0.8 for t in 1:24]
+    renewable = add_renewable_dispatch_with_time_series!(
+        c_sys5,
+        thermal_generator,
+        availability;
+        name = "high_availability_renewable",
+        curtailment_cost = 50.0,
+    )
+
+    template = ProblemTemplate(NetworkModel(CopperPlatePowerModel))
+    set_device_model!(template, RenewableDispatch, RenewableFullDispatch)
+    set_device_model!(template, ThermalStandard, ThermalStandardDispatch)
+    set_device_model!(template, PowerLoad, StaticPowerLoad)
+    model = DecisionModel(
+        template,
+        c_sys5;
+        optimizer = HiGHS_optimizer,
+        resolution = Hour(1),
+        interval = Day(1),
+    )
+
+    @test build!(model; output_dir = mktempdir(; cleanup = true)) ==
+          PSI.ModelBuildStatus.BUILT
+    @test solve!(model) == PSI.RunStatus.SUCCESSFULLY_FINALIZED
+
+    results = OptimizationProblemResults(model)
+    dispatch = read_variable(
+        results,
+        "ActivePowerVariable__RenewableDispatch";
+        table_format = TableFormat.WIDE,
+    )[
+        !,
+        get_name(renewable),
+    ]
+    curtailment_cost = read_expression(
+        results,
+        "CurtailmentCostExpression__RenewableDispatch";
+        table_format = TableFormat.WIDE,
+    )[
+        !,
+        get_name(renewable),
+    ]
+    availability_mw =
+        availability .* get_max_active_power(renewable) .* get_base_power(c_sys5)
+    expected_cost = 50.0 .* (availability_mw .- dispatch)
+
+    @test length(unique(availability_mw)) == 2
+    @test all(dispatch .<= availability_mw .+ 1e-8)
+    @test any(availability_mw .- dispatch .> 1e-8)
+    @test curtailment_cost ≈ expected_cost
+end
--- a/test/test_utils/add_components_to_system.jl
+++ b/test/test_utils/add_components_to_system.jl
@@ -171,3 +171,67 @@ function add_reserve_product_without_requirement_time_series!(
     )
     add_service!(sys, reserve_instance, contributing_devices)
 end
+
+"""
+    add_renewable_dispatch_with_time_series!(
+        sys,
+        reference_generator,
+        availability;
+        name = "renewable_dispatch",
+        curtailment_cost = 0.0,
+    )
+
+Adds a `RenewableDispatch` with the same static maximum active power as
+`reference_generator` and an hourly deterministic `max_active_power` forecast.
+`availability` values are per unit of the renewable's static maximum active
+power.
+
+# Arguments
+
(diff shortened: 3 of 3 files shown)


src/devices_models/devices/common/objective_function/linear_curve.jl
```diff
@@ -263,18 +267,22 @@ function _add_curtailment_cost!(
 
     name = PSY.get_name(component)
     dispatch_vars = get_variable(container, T(), PSY.RenewableGen)
-    param_array =
-        get_parameter_array(container, ActivePowerTimeSeriesParameter(), PSY.RenewableGen)
+    param_container =
+        get_parameter(container, ActivePowerTimeSeriesParameter(), PSY.RenewableGen)
     param_multiplier = get_parameter_multiplier_array(
```
acostarelli, line 272: If we have the `ParameterContainer` we can just call `get_multiplier_array(param_container)`

src/devices_models/devices/common/objective_function/linear_curve.jl
```diff
@@ -263,18 +267,22 @@ function _add_curtailment_cost!(
...
         container,
         ActivePowerTimeSeriesParameter(),
         PSY.RenewableGen,
     )
-    has_ts_param = name in axes(param_array)[1]
+    has_ts_param = haskey(get_attributes(param_container).component_name_to_ts_uuid, name)
+    availability = nothing
+    if has_ts_param
+        availability = get_parameter_column_refs(param_container, name)
```
acostarelli, line 280: There's some weird indexing that goes on with `get_parameter_column_refs` that feels iffy to me. But I'm guessing it's faster to just grab the TS UUID and
`get_parameter_array(param_container)[uuid, t]`

src/devices_models/devices/common/objective_function/linear_curve.jl
```diff
@@ -263,18 +267,22 @@ function _add_curtailment_cost!(
...
+    param_container =
+        get_parameter(container, ActivePowerTimeSeriesParameter(), PSY.RenewableGen)
     param_multiplier = get_parameter_multiplier_array(
         container,
         ActivePowerTimeSeriesParameter(),
         PSY.RenewableGen,
     )
-    has_ts_param = name in axes(param_array)[1]
+    has_ts_param = haskey(get_attributes(param_container).component_name_to_ts_uuid, name)
+    availability = nothing
```
acostarelli, line 278: This is a type-instability. I'm not sure how pressing it is performance-wise, but if there's a clean function guard I would prefer that.

src/devices_models/devices/common/objective_function/linear_curve.jl
SebastianManriqueM: Implemented in d1587c885: this now uses get_multiplier_array(param_container), avoiding the redundant container/key lookup.

src/devices_models/devices/common/objective_function/linear_curve.jl
SebastianManriqueM: Kept get_parameter_column_refs(param_container, name). It is the established PSI component-name-to-UUID resolver and internally performs the UUID lookup before indexing the parameter array. Using it keeps the TimeSeriesAttributes mapping and DenseAxisArray indexing encapsulated; it is already used in parameterized constraints and cost construction.

src/devices_models/devices/common/objective_function/linear_curve.jl
SebastianManriqueM: Implemented in d1587c885: the time-series and static-rating cases are now separate branches. availability is bound only in the time-series branch, removing the Union{Nothing, ...} local and the conditional from the timestep loop.

SebastianManriqueM:
All the tests are passing locally on Julia 1.13:

```
test_utils: 
Test Summary:                |  Pass  Total      Time
Begin PowerSimulations tests | 46982  46982  36m39.2s
2199.706522 seconds (2.27 G allocations: 173.243 GiB, 3.36% gc time, 364 lock conflicts, 52.04% compilation time: <1% of which was recompilation)
```

SebastianManriqueM:
@daniel-thom, @jerrypotts and @acostarelli . I've run on my local all the tests for julia 1.12 and 1.13 and all are passing in both:

```
Test Summary:                |  Pass  Total      Time
Begin PowerSimulations tests | 46982  46982  36m39.2s
2199.706522 seconds (2.27 G allocations: 173.243 GiB, 3.36% gc time, 364 lock conflicts, 52.04% compilation time: <1% of which was recompilation)
```

However, I don't understand why in github actions (https://github.com/Sienna-Platform/PowerSimulations.jl/actions/runs/35183756520/job/106888086139?pr=1667) I'm still getting some failing tests in: 
- test_device_thermal_generation_constructors.jl
- test_model_decision.jl
- test_simulation_partitions.jl
- test_static_injection_security_constrained_models.jl

I asked Anthony to test on his local this branch to see If I get some insights.

acostarelli:
```
julia> include("test/runtests.jl")
...
Test Summary:                |  Pass  Total      Time
Begin PowerSimulations tests | 46982  46982  20m41.1s
1241.420210 seconds (2.23 G allocations: 175.094 GiB, 2.70% gc time, 50.49% compilation time: <1% of which was recompilation)

julia> VERSION
v"1.13.0"
```

SebastianManriqueM:
> @daniel-thom, @jerrypotts and @acostarelli . I've run on my local all the tests for julia 1.12 and 1.13 and all are passing in both:
> 
> ```
> Test Summary:                |  Pass  Total      Time
> Begin PowerSimulations tests | 46982  46982  36m39.2s
> 2199.706522 seconds (2.27 G allocations: 173.243 GiB, 3.36% gc time, 364 lock conflicts, 52.04% compilation time: <1% of which was recompilation)
> ```
> 
> However, I don't understand why in github actions (https://github.com/Sienna-Platform/PowerSimulations.jl/actions/runs/35183756520/job/106888086139?pr=1667) I'm still getting some failing tests in:
> 
> * test_device_thermal_generation_constructors.jl
> * test_model_decision.jl
> * test_simulation_partitions.jl
> * test_static_injection_security_constrained_models.jl
> 
> I asked Anthony to test on his local this branch to see If I get some insights.

Update on this, It seems the test are failing on github actions because HiGHS also released a new version recently (https://github.com/jump-dev/HiGHS.jl) so github actions is resolving to 1.25.4 and I'm testing in 1.25.3. That is causing some other errors  that we were not seeing at the begining. I'm aticipating It'll take me some time to debug those

jd-lara (approved):

PR #1668, opened 2026-09-09
remove parallel_branch_max_rating_method default

acostarelli:


src/devices_models/devices/AC_branches.jl
```diff
@@ -117,15 +117,17 @@ end
...
     name = get_attribute(model, PARALLEL_BRANCH_MAX_RATING_KEY)
     name == "single_element_contingency" &&
         return PNM.get_single_element_contingency_rating(bp)
     name == "sum_of_max" && return PNM.get_sum_of_max_rating(bp)
     name == "impedance_averaged" && return PNM.get_impedance_averaged_rating(bp)
     error(
-        "Unknown $PARALLEL_BRANCH_MAX_RATING_KEY value: $(repr(name)). " *
-        "Valid: \"single_element_contingency\", \"sum_of_max\", \"impedance_averaged\".",
+        "Attribute $PARALLEL_BRANCH_MAX_RATING_KEY on DeviceModel{$D, $M} must be set to one of \"single_element_contingency\", \"sum_of_max\", or \"impedance_averaged\". Note that parallel merges of mixed branch types always use \"sum_of_max\".",
     )
```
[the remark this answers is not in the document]
acostarelli: I don't know if I agree that this is significantly difficult.

src/devices_models/devices/AC_branches.jl
```diff
@@ -59,25 +59,27 @@ end
...
+    return Dict{String, Any}(PARALLEL_BRANCH_MAX_RATING_KEY => "")
 end
 
 function get_default_attributes(
     ::Type{U},
     ::Type{V},
 ) where {U <: PSY.ACTransmission, V <: AbstractSecurityConstrainedStaticBranch}
     return Dict{String, Any}(
-        PARALLEL_BRANCH_MAX_RATING_KEY => "single_element_contingency",
+        PARALLEL_BRANCH_MAX_RATING_KEY => "",
```
SebastianManriqueM: @acostarelli maybe for this case where the user is explicitly asking to consider n-1 contingencies we can compute the equivalent rating using "single_element_contingency", that can be a more computationally efficient than add flow gates... but also can complicate the selection of the flow gates for users.
Any thoughts @jarry7 ?

SebastianManriqueM: I thought about this and I think it is better to left as it is

PR #1674, opened 2026-09-15
Update tests for Julia 1.13 and current PowerSystems psy6 API

jd-lara:
Compare result column names as a set: the wide table follows the container axis, which follows Dict iteration order, and Julia 1.13 changed it.

Replace symbols PowerSystems psy6 removed: PSY.MW -> u"MW", PSY.DU ->
PSY.CU, IS.DeviceBaseUnit -> IS.ComponentBaseUnit, Bool status -> OperationalStates, set_must_run! -> set_commitment_mode!, and tag ramp limits with a per-minute rate.


docs/Project.toml
```diff
@@ -39,7 +39,7 @@ InfrastructureSystems = {url = "https://github.com/Sienna-Platform/Infrastructur
 InfrastructureOptimizationModels = {url = "https://github.com/Sienna-Platform/InfrastructureOptimizationModels.jl", rev = "main"}
 PowerSystems = {url = "https://github.com/Sienna-Platform/PowerSystems.jl", rev = "psy6"}
 PowerNetworkMatrices = {url = "https://github.com/Sienna-Platform/PowerNetworkMatrices.jl", rev = "psy6"}
-PowerOperationsModels = {url = "https://github.com/Sienna-Platform/PowerOperationsModels.jl", rev = "main"}
+PowerOperationsModels = {url = "https://github.com/Sienna-Platform/PowerOperationsModels.jl", rev = "jd/fix_tests"}
```
jd-lara, line 42: Temporary

luke-kiernan (approved):
Looks like a straightforward find-and-replace.

Oh...one part that isn't: `update_cost_parameters.jl‎`. Before `set_expression!` was called with unsigned gen cost; now, signed gen cost. Is that a bugfix or a known POM vs PSI difference?

PR #1677, opened 2026-09-18
Store the results system as an OpenAPI document

jd-lara:
`to_json`/`from_json` are gone from psy6 PowerSystems. The snapshot in the results HDF5 is now an OpenAPI document written with `include_time_series = false`: the time series values are already in the store's own datasets, so association rows would duplicate them. The store's group key restores the System UUID on read, which `results.system_uuid` is checked against.

`get_system!` and `set_system!` were still reaching for the psy5 `System(file)` constructor and a `system-<uuid>.json` nothing writes; both now read the `system-<uuid>` bundle PowerOperationsModels writes, the only form that carries the time series.


jd-lara:
this PR is also catching up with the EnumX stuff

